PR #167, opened 2013-08-25
Update french phone numbers

vchabot:
Add these french phone numbers:
07 XX XX XX XX (new mobile phone numbers)
09 XX XX XX XX (phone numbers)


armetiz:
Faker generate 0729101484 which is not a correct french number.

Found from this issue : https://github.com/googlei18n/libphonenumber/issues/500
See how 07 should be managed : 
- http://www.arcep.fr/index.php?id=8146

fzaninotto:
@armetiz fixed in #1046

armetiz:
Thanks ;)

Issue #304, opened 2014-04-17
Add biased/weighted “random” numbers


max-m:
What about adding an generator for “weighted random numbers”?
For example a generator like this gamma “corrected” random number generator:
> [Author: bobince](http://stackoverflow.com/a/445363/1112384)
> ```PHP
/*
 * Generates a weighted random number
 *
 * Choose a continuous random number between 0..1
 * Raise to a power γ, to bias it. 1 is unweighted, lower gives more of the higher numbers and vice versa
 * Scale to desired range and round to integer
 *
 * @author bobince <http://stackoverflow.com/a/445363/1112384>
 */
public static function weightedRand($min, $max, $gamma) {
	$offset = $max - $min + 1;
	return floor($min + pow(lcg_value(), $gamma) * $offset);
}
```

Here is the probability distribution of calling `weightedRand(1, 4, $gamma);`.
May need some tweeking with values lower or equal to 0 though.

```
[kernel@~]php development/test.php 2.2
Gamma: 2.2
1: 53233 (53.233%)
2: 19662 (19.662%)
3: 14678 (14.678%)
4: 12427 (12.427%)
[kernel@~]php development/test.php 1.9
Gamma: 1.9
1: 48102 (48.102%)
2: 21317 (21.317%)
3: 16449 (16.449%)
4: 14132 (14.132%)
[kernel@~]php development/test.php 0
Gamma: 0
5: 100000 (100%)
[kernel@~]php development/test.php -.09
Gamma: -0.09
5: 91459 (91.459%)
6: 7433 (7.433%)
7: 901 (0.901%)
8: 155 (0.155%)
9: 35 (0.035%)
10: 8 (0.008%)
11: 5 (0.005%)
12: 3 (0.003%)
14: 1 (0.001%)
```

This might come in handy some times ^^

fzaninotto:
Fixed by #332

PR #332, opened 2014-05-14
Biased integer

TimWolla:
biasedInteger allows generating an integer that is biased by a given function.

See fzaninotto/Faker#304

See this example data:
graph

Number	|unbiased|	linearLow|	linearHigh|	sqrt	|square
-------------|------------|------------------|-------------------|------------|---------
1	|9857	|19016|	1007	|3222|	84
2	|10032	|17000|	2931	|5715|	717
3|	9966	|14977|	4928	|7538|	1846
4|	9914	|13015|	7033	|8872|	3645
5|	10099	|11080|	9145	|9915|	6095
6|	9800	|9012	|11001|	11234|	9138
7|	9987	|6860	|12936|	11906|	12800
8|	10096	|5053	|15047|	13160|	17019
9|	10026	|2945	|16919|	13775|	21512
10|	10223	|1042	|19053|	14663|	27144					
Loop iterations:	|100000|	200217|	199822	|150115|	299345
Loop iterations (percent)	|100.00%|	200.22%	|199.82%|	150.12%	|299.34%

Loop iterations are the number of times the contents of the loop were executed, `100000` is perfect, no values were discarded.

For even more fascination see this `sqrt` distribution of a higher number range :smile:
graph2

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -233,6 +233,10 @@ Each of the generator properties (like `name`, `address`, and `lorem`) are calle
     countryCode             // UK
     languageCode            // en
 
+### `Faker\Provider\Biased`
+
+    biasedInteger($min, $max, $function) // 42
+
 ## Unique and Optional modifiers
 
 Faker provides two special providers, `unique()` and `optional()`, to be called before any provider. `optional()` can be useful for seeding non-required fields, like a mobile telephone number; `unique()` is required to populate fields that cannot accept twice the same value, like primary identifiers.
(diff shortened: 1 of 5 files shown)


fzaninotto:
Thanks!

jayenne:
Hi, would this function be able to have return a number biased towards a 3rd number using a function?

min = 0, max = 100, focus = 33, function= ? returns a "falloff curve" around the focal point rather than one end or the other?

christhofer:
Hi, what other values that we can use in 3rd parameter? Please update the documentation with more example, e.g how to make it biased for lower values, for middle values, or to use our own function maybe.

PR #333, opened 2014-05-15
Added transliterator to email and username

hugofonseca:
see #311
I've added transliteration to email and username.
I've added some tests that will make sure that every email and username from any provider are valid.

I've updated readme.md file to inform of the new intl and php dependency.
Should we be worried about those people that need transliteration (of any kind) but still use a lower PHP version?
For these people Faker will not work after the update as I've removed all "toAscii" methods.


fzaninotto:
Superseded by #490

PR #372, opened 2014-07-18
add random GET argument for imageUrl faker

Nicals:
When using multiple imageUrl with same configuration, only one image is loaded from lorempixel since the generated url are the same.

This pull requests adds an argument to tell Faker to add a random GET parameter to the generated url in order to have different images from lorempixel.


fzaninotto:
Superseded by #494

fzaninotto:
Merged in the other PR, thanks for your contribution!

PR #373, opened 2014-07-22
Add MaxGenerator to Faker.

pinktrink:
MaxGenerator will allow the use of `$faker->max($num)->something` to
only return a maximum of $num different results.

I'm not sure if this functionality exists already, as I've only minimally messed with Faker for doing a bit of fuzz testing, and haven't really had time to fully explore the codebase. The problem I ran into was that while generating random data, I also wanted to ensure that collisions existed within the dataset I was generating. The solution was to add `MaxGenerator` and expose a `max()` method on `\Faker\Provider\Base` which would allow me to limit the data I would receive.

### DIFF
--- a/src/Faker/MaxGenerator.php
+++ b/src/Faker/MaxGenerator.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace Faker;
+
+/**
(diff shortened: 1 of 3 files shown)


fzaninotto:
Closing because no news for 6 months.

PR #401, opened 2014-08-25
some hu_HU fixes

peet86:
- fixed the city formats (replaced the first and lastname generation)
- company fix (the second letters replaced to lowercase, "association" and "found" words added)


fzaninotto:
No news for a while, closing.

PR #421, opened 2014-09-07
en_GB people provider

AlexCutts:
I've recently found myself using Faker for some test data to provide a demonstration system and found that the names which were being generated in most cases, weren't names which are commonly used in the UK.

I've created a simple Person class for the UK.

Data for forenames came from ONS [female names here](http://www.ons.gov.uk/ons/rel/vsob1/baby-names--england-and-wales/2013/info-girls-names-2013.html) and [male names here](http://www.ons.gov.uk/ons/rel/vsob1/baby-names--england-and-wales/2013/info-boys-names-2013.html) - I used the data from 2003 and 2013.

I also used some of the names already in the en_US locale which were relevant to a UK audience.

For surnames, I found a list [here](http://surname.sofeminine.co.uk/w/surnames/most-common-surnames-in-great-britain.html) which seemed to reflect usage well.

I haven't added any tests as I've only changed the contents of the arrays, but if there's anything else I should be doing let me know. I've not made an open source contribution before, and I wasn't entirely sure if I should be submitting this as none of the other EN locales have specific Person providers (other than the default en_GB).

### DIFF
--- a/src/Faker/Provider/en_GB/Person.php
+++ b/src/Faker/Provider/en_GB/Person.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace Faker\Provider\en_GB;
+
+class Person extends \Faker\Provider\Person
+{
+    protected static $maleNameFormats = array(
+        '{{firstNameMale}} {{lastName}}',
+    );
+
+    protected static $femaleNameFormats = array(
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks!

PR #427, opened 2014-09-23
Add zh_TW Provider

tzhuan:


vlakoff:
Why the added [`Internet.php`](https://github.com/fzaninotto/Faker/pull/427/files#diff-ec78d44163a7674f58bb5bccc0beda1f) and [`Payment.php`](https://github.com/fzaninotto/Faker/pull/427/files#diff-16d843a6ae434a16f179461161d05adc)?

tzhuan:
The romanization of the name is usually required and used for credit-card in Taiwan.
Chinese domain names do exist but they are rarely used. Most domain names are still in English in Taiwan. Therefore, I added the Payment.php and Internet.php to keep the English version of the `Payment::creditCardInfo()`, `Internet::userName()` and `Internet::domainWord()` methods.

PR #438, opened 2014-10-09
Addition of en_GB mobileNumber methods

daveblake:
Additions to the PhoneNumber Provider for en_GB so that a UK format phone number can be returned.


fzaninotto:
Thanks!

PR #456, opened 2014-11-10
Sanitizing username to prevent invalid email

EmileBons:
In certain locales (e.g. nl_NL), special characters are used in names such as ç; when using these characters for generating email addresses, this will result in an invalid email address (e.g. valbinusgenaamdweissvonweissenlöw@example.net or sarah.ünal@example.com), which causes tests in my case to fail.

### DIFF
--- a/src/Faker/Provider/Internet.php
+++ b/src/Faker/Provider/Internet.php
@@ -94,7 +94,7 @@ class Internet extends \Faker\Provider\Base
 
         $userName = static::toLower(static::bothify($this->generator->parse($format)));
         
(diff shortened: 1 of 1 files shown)


nicodmf:
@EmileBons I'm not sure that should be necessary as utf8 email address become actually an accepted option : 

- http://en.wikipedia.org/wiki/Email_address#Internationalization
- http://tools.ietf.org/html/rfc5335

fzaninotto:
We're now using transliteration ; this is not necessary anymore. Thanks anyway for your PR.

Issue #68, opened 2012-11-06
Idea: image faker


drodil:
Would it be bad idea to also provide API for faking images? This could easily be done with gd library and would also help when styling webpage with random data in it. What do you think?

PR #474, opened 2014-12-10
generating image filled with random color locally

JarekW:
This adds image generation filled with one random color. Idea was started in #68.
Lorempixel lastly have problems with service unavailability. This also generates images faster than downloading images.


fzaninotto:
Thanks for your contribution. 

I don't think this fits a need that is common. I suggest that you publish this as a standalone extension to Faker on your own GitHub account.

PR #475, opened 2014-12-11
Doctrine orm polish

lexicalbits:
I made 2 changes to facilitate my use of Faker against Doctrine entities.

First, to track down formatter errors, I re-throw the InvalidArgumentException with the specific class name and field name that failed.  This helped me track down several issues I had tracking down decimal fields that were bigger than mt_getrandmax.

Second, Doctrine entities usually have setters for their properties, and those setters can have their own behaviors (regenerating slugs on setting a title, passing relationship assignment between owner and inversion, etc.).  I therefore thought trying to use the setter to assign a value before falling back on reflection might improve the resulting object mockup.  The reflection strategy is useful for hydration, where the data has already been established and is merely being re-loaded into an object, not for creating new data, as the EntityPopulator is trying to do.

### DIFF
--- a/src/Faker/ORM/Doctrine/EntityPopulator.php
+++ b/src/Faker/ORM/Doctrine/EntityPopulator.php
@@ -170,7 +170,10 @@ class EntityPopulator
                         $ex->getMessage()
                     ));
                 }
-                $this->class->reflFields[$field]->setValue($obj, $value);
+                //Try a standard setter if it's available, otherwise fall back on reflection
+                $setter = sprintf("set%s", ucfirst($field));
+                if(method_exists($obj, $setter)) $obj->$setter($value);
+                else $this->class->reflFields[$field]->setValue($obj, $value);
             }
         }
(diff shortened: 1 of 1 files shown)


fzaninotto:
Tests still fail.

fzaninotto:
@cbourgois any feedback on this one?

lexicalbits:
@fzaninotto I'm happy to take a more detailed pass at this if you think my position makes any sense.  If my premise doesn't fit what you're trying to accomplish, feel free to ditch the PR.

fzaninotto:
@lexicalbits My problem is that I'm unable to review your code, and I'd like another Doctrine user to double check.

cbourgois:
It seems ok.

Tests still fails. 
@lexicalbits , read https://travis-ci.org/fzaninotto/Faker/jobs/43672548 for help

fzaninotto:
Merged via the command line. Thanks!

PR #478, opened 2014-12-14
add new locale for en_SG

simkimsia:
unit test for mobile phone numbers is included


src/Faker/Provider/en_SG/PhoneNumber.php
```diff
@@ -0,0 +1,130 @@
...
+        $format = static::randomElement(static::$sgPhoneNumberFormats);
+
+        return $this->generator->parse($format);
+    }
+
+    /**
+     * this overrides the static function in Faker\Provider\PhoneNumber
+     * @return string PhoneNumber in string format
+     */
+    public static function phoneNumber()
```
fzaninotto: Why is that useful? I think you have to remove it.

fzaninotto:
Merged via the command line, thanks.

PR #480, opened 2014-12-18
added Twitter Handler to Internet Provider

j0an:


fzaninotto:
All that for an `@`... I think this doesn't bring much value. Faker users will prefer concatenating the `'@'` themselves instead of learning another formatter.

Thanks for your suggestion anyway.

GrahamCampbell:
Same. This would have been really useful.

fzaninotto:
@GrahamCampbell Not sure I get your point. You think a twitter handler formatter would be useful or not?

GrahamCampbell:
I think it would have been useful to have, yeh.

j0an:
I thought that it might be useful. I needed and I think that hardcode an @ was just fine, but a specific handler will be better.



j.

fzaninotto:
But then we need a formatter for Google + which adds a "+", and a formatter for Facebook which just returns `username`, and I don't talk about snapchat or Pinterest... It doesn't make sense in my opinion.

j0an:
aha, you have a point.

perhaps separating the *Internet Handlers* and creating a *Social Media
Handlers*?




j.

PR #483, opened 2014-12-22
Fix cases where no entities prev. inserted

jadb:
Happens when a TestFixture uses Faker but associated models (fixtures) do not (or have no
records seeded yet).


jadb:
Bump...

fzaninotto:
Thanks!

PR #490, opened 2015-01-02
Add transliterator to email and username

fzaninotto:
Based on the great work by @hugofonseca

Supersedes #333


GrahamCampbell:
Might be an idea to squash commits per author.

fzaninotto:
Merged! Thanks a lot @hugofonseca!

PR #491, opened 2015-01-04
Improve unique generator performance

ikwattro:
Hi,

Currently the unique generator keeps the already used elements as array values and performs the lookup with `in_array` to assert that the generated value has not already been used. This operation is a O(n) operation which becomes slower in time for large datasets generation.

I refactored this operation to uses array indexes instead, which will leads to hash table lookups that are O(1) operations.

I did a little benchmark generating 10k unixTimes, the generation time is decreased from 1.77 seconds to 0.37 seconds.


GrahamCampbell:
Cool. :)

fzaninotto:
Thanks!

PR #497, opened 2015-01-10
Created faker for Currency Codes.

stelgenhof:
Created faker for currency codes (similar to country codes and language codes). The list of currencies are taken from the ISO 4217 standard list.

### DIFF
--- a/src/Faker/Provider/Miscellaneous.php
+++ b/src/Faker/Provider/Miscellaneous.php
@@ -154,6 +154,30 @@ class Miscellaneous extends \Faker\Provider\Base
     );
 
     /**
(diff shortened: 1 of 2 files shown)


fzaninotto:
Thanks!

stelgenhof:
Thanks a lot! One thing I forgot and that was to update the realm file :) Hopefully you can update that so I don't need to create another PR.

fzaninotto:
Ah, right. Can you create a small additional PR just for that? It's better if the commit has your name on it :)

stelgenhof:
Sure! Besides the readme file, are there any other documentation files I need to update? Changelog for example?

fzaninotto:
No, just the readme ; I'll handle the changelog.

PR #508, opened 2015-01-21
add more options for imageUrl

ctf0:
u still can use as usual and u will get the lormpixel or pass the first argument as ('Unsplash' or 'placeimg') "the same names from the original links to avoid any confusion" ,

the original function still have the same order for options "again to avoid any confusion" so if u want to pass options to this services here is the fill function with arguments

```
imageUrl($service = 'lorem', $width = 640, $height = 480, $category = null, $randomize = true, $filter = null, $blur = null, $gravity = null)
```

### DIFF
--- a/src/Faker/Provider/Image.php
+++ b/src/Faker/Provider/Image.php
@@ -12,6 +12,11 @@ class Image extends Base
         'fashion', 'people', 'nature', 'sports', 'technics', 'transport'
     );
 
+    protected static $placeimg_cats = array(
+        'animals', 'arch', 'nature', 'people', 'tech', 'any'
+    );
+
+
     /**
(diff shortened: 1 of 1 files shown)


fzaninotto:
So "category" is ignored or has a different meaning, and you add a useless "filter" parameter which only works once...

I'n not fan of having a one-size-fits-all provider with inconsistent signature. If you're not satisfied with the current randomImage, feel free to create a third-party provider and host it on your own GitHub, and I'll be happy to link to it in the Faker Readme.

ctf0:
- `category` only works for `lorem` & `placeimg` but not for `Unsplash`
- 'filter' works for both `placeimg` & `Unsplash` but not for `lorem`

PR #511, opened 2015-01-25
Update UserAgent.php

ajbdev:
Updating random user agent strings to appear more modern

I based my changes off a recent log file on a relatively active website I help maintain

### DIFF
--- a/src/Faker/Provider/UserAgent.php
+++ b/src/Faker/Provider/UserAgent.php
@@ -65,9 +65,9 @@ class UserAgent extends \Faker\Provider\Base
         $saf = mt_rand(531, 536) . mt_rand(0, 2);
 
         $platforms = array(
-            '(' . static::linuxPlatformToken() . ") AppleWebKit/$saf (KHTML, like Gecko) Chrome/" . mt_rand(13, 15) . '.0.' . mt_rand(800, 899) . ".0 Safari/$saf",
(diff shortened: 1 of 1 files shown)


ajbdev:
I am aware that the contributers file reads: "Avoid changing existing sets of data. Some developers use Faker with seeding for unit tests ; changing the data makes their tests fail."

However, the UserAgent provider is a bit unique, as realistic fake data evolves much quicker than the other providers. I'm running into a use case where I am using faker to generate a fake user agent for a scraper. Given the current state of Faker it will create a pattern of outdated browsers, which is undesirable for a scraping tool.

fzaninotto:
Does that mean we'll have to update this provider each time a browser vendor bumps their version? I don't think it makes sense.

I agree with you that the current useragents are outdated, but Faker doesn't mean to be exhaustive, it means to be random. 

I wonder what other users feel about this?

ajbdev:
Yeah, I can certainly understand that point of view - there's no way I would want to maintain this so frequently. Really I just did the PR to cover my specific use case, if it's a no-go I understand I'll just use my fork. 

If the data was just a year or two old it would have been fine for my use case, but it was pretty far outdated. If it's ever been considered to update it, here's an opportunity with most of the work already done. 

I'll be in the same spot again in 2-3 years, but I'm comfortable saying that I'd maintain a frequency at that rate :)

fzaninotto:
Thanks!

bomjastik:
Hi @ajbdev. 

Just decided to use this functionality for a random user-agent generation too. 

I can't find a more specific solution/api.

> I'll be in the same spot again in 2-3 years.

The time is definitely have come :)

Will, you able to update this again?

Regards!

PR #514, opened 2015-02-04
+1 Internet::html() provider, based on Lorem::text()

ronanguilloux:
Attempting to re-propose an HTML+Lorem Ipsum provider, as in #250


ronanguilloux:
rebasing...

test/Faker/Provider/InternetTest.php
```diff
@@ -117,4 +117,50 @@ public function testMacAddress()
...
+    }
+
+    public function test_htmlTooSmall()
+    {
+        $this->setExpectedException('InvalidArgumentException');
+        Internet::html(Internet::SMALLEST_HTML_LENGTH + 4);
+        Internet::html(Internet::SMALLEST_HTML_LENGTH - 1);
+    }
+
+    public function test_htmlMinimal()
```
ronanguilloux: #noteToSelf : useless, test_htmlTooSmall() already contains that test

src/Faker/Provider/Internet.php
```diff
@@ -331,4 +331,94 @@ public static function macAddress()
...
+        }
+        $textLength = $maxLength - $htmlTagsLength;
+
+        switch($maxLength) {
+            case ($maxLength >= (self::SMALLEST_HTML_LENGTH+150)): // actual $htmlTagsLength is 92
+                /*
+                Since Lorem::text() output length is 99% always below than the request $maxNbChars
+                while looping over the tag list, generating the html output,
+                we recalculate the real remaining, "weighted", text length to be outputted.
+                Makes The result as closer as possible from the max expected value
```
ronanguilloux: #noteToSelf: s/The/the

src/Faker/Provider/Internet.php
```diff
@@ -331,4 +331,94 @@ public static function macAddress()
...
+     * Principe 2: Available html tags list is fixed, for now
+     * Principe 3: Rendered HTML output is poorly semantic, for now.
+     *
+     * @param int $maxLength max string size expected
+     *
+     * @return string html output - html tags + Lorem text rendering
+     */
+    public static function html($maxLength = 100)
+    {
+        if (0 > ($maxLength - self::SMALLEST_HTML_LENGTH - 5)) { // 5 == Lorem::text() argument limit
```
ronanguilloux: #noteToSelf: this repeated "5" value may become a const in the `Lorem` class, or at least a local variable

test/Faker/Provider/InternetTest.php
```diff
@@ -117,4 +117,50 @@ public function testMacAddress()
...
+
+    public function test_htmlFibonacci()
+    {
+        $fiboLength = 22; // Fibonacci's sequence 22th number is 17711
+        $fibos= array(0, 1);
+        for ($i=2; $i<=$fiboLength; ++$i) {
+            $fibos[$i] = $fibos[$i-1] + $fibos[$i-2];
+        }
+        foreach($fibos as $fibo) {
+            if(9 < $fibo) {
```
ronanguilloux: #noteToSelf: this "9" var as to be explained a bit (cf. minimal int argument required by ::html())

test/Faker/Provider/InternetTest.php
```diff
@@ -117,4 +117,50 @@ public function testMacAddress()
...
+        Internet::html(Internet::SMALLEST_HTML_LENGTH - 1);
+    }
+
+    public function test_htmlMinimal()
+    {
+        $this->setExpectedException('InvalidArgumentException');
+        Internet::html(Internet::SMALLEST_HTML_LENGTH - 1);
+    }
+
+    public function test_htmlFibonacci()
```
fzaninotto: I don't understand the purpose of this test.

test/Faker/Provider/InternetTest.php
```diff
@@ -117,4 +117,50 @@ public function testMacAddress()
...
+
+    public function test_htmlLoremSmallLengths()
+    {
+        $plus5 = Internet::html(Internet::SMALLEST_HTML_LENGTH+5);
+        $this->assertTrue($this->containsHtml($plus5));
+    }
+
+    public function test_htmlTooSmall()
+    {
+        $this->setExpectedException('InvalidArgumentException');
```
fzaninotto: Can you use annotations for this, as everywhere else in Faker?

test/Faker/Provider/InternetTest.php
```diff
@@ -117,4 +117,50 @@ public function testMacAddress()
...
+        Internet::html(Internet::SMALLEST_HTML_LENGTH + 4);
+        Internet::html(Internet::SMALLEST_HTML_LENGTH - 1);
```
fzaninotto: Which one is supposed to throw an exception, the first or the second? Perhaps you need two tests?

test/Faker/Provider/InternetTest.php
```diff
@@ -117,4 +117,50 @@ public function testMacAddress()
...
+                $html = Internet::html($fibo);
+                $this->assertTrue($this->containsHtml($html));
+            }
+        }
+
+    }
+
+    protected function containsHtml($string)
+    {
+        return preg_match("/<[^<]+>/",$string, $matches) !== 0;
```
fzaninotto: Not very strict. `<div>foo<span>` matches...

fzaninotto:
I don't think you have to focus on `<article>` or `<section>`. Usually, rich text is output inside `<div>`s anyway.

It would be great to add a little more sophistication:

- distinguish block-level and inline tags, and allow inline tags inside blocks
- add `<h1>`, `<h2>` semantics (usually one after the other)

src/Faker/Provider/Internet.php
```diff
@@ -331,4 +331,94 @@ public static function macAddress()
...
+    const SMALLEST_HTML_LENGTH = 4; // <br>
+
+    /**
+     * @param string $tag
+     * @param string $content
+     * @param string $attributes as a string
+     *
+     * @return string
+     */
+    protected static function buildHtmlTag($tag, $content = '', $attributes = '')
```
fzaninotto: attributes should be an array

src/Faker/Provider/Internet.php
```diff
@@ -331,4 +331,94 @@ public static function macAddress()
 
         return $mac;
     }
+
```
fzaninotto: Not sure if `Internet` is the right provider for it.

test/Faker/Provider/InternetTest.php
ronanguilloux: ok!

test/Faker/Provider/InternetTest.php
ronanguilloux: You're right, both must, so we may need individual tests for each of them two

src/Faker/Provider/Internet.php
ronanguilloux: ok

src/Faker/Provider/Internet.php
ronanguilloux: I'm very OK to create a distinct `Html::` provider instead: is it OK to you ?

test/Faker/Provider/InternetTest.php
ronanguilloux: That's right. Well, the purpose here was only checking that `Internet::html` actually return some html tags, rather  _validate_ any HTML output. I may remove this func, in order not to measled people.

test/Faker/Provider/InternetTest.php
ronanguilloux: It tests the provider over large numbers of text argument (using a Fibonacci suite), but it does not check that these various text sizes actually match the various cases of the `switch($maxLength)` operator of the Internet::html() implementation. So all this remains quite useless for now.
=> Removed!
Thanks for having spotted this.

src/Faker/Provider/Internet.php
fzaninotto: Yes, would be great.

src/Faker/Provider/Internet.php
```diff
@@ -331,4 +331,128 @@ public static function macAddress()
...
+    /**
+     * @link http://w3c.github.io/html-reference/global-attributes.html
+     * pro-memoria for now
+     *
+     * @return array
+     */
+    protected static function getRandomHtmlTagsGlobalAttribute()
+    {
+        return self::randomElement(array(
```
fzaninotto: The right way to do that in Faker is to use a list of possible formats, pick one random format, then call `$generator->parse($format)`. Something like:

```php
protected static $htmlAttributeFormats = array(
  'accesskey="{{randomletter}}"',
  'class="{{word}}"',
  'contenteditable="true"',
  'contenteditable="false"',
  // etc
);
```

And if the `parse()` method seems too limited because it doesn't allow formatters with arguments... improve it!

src/Faker/Provider/Internet.php
```diff
@@ -331,4 +331,128 @@ public static function macAddress()
...
+    {
+        return sprintf('<%s>%s</%s>', $tag, $content, $tag);
+    }
+
+    /**
+     * @return array of distinct block-level and inline-level elements names
+     */
+    protected static function getHtmlTags()
+    {
+        return array(
```
fzaninotto: You should move that to a static property to let devs override it in subclasses.

src/Faker/Provider/Internet.php
```diff
@@ -331,4 +331,128 @@ public static function macAddress()
...
+                $maxLength,
+                self::SMALLEST_HTML_LENGTH + Lorem::TOO_SHORT_TEXT_ARGUMENT,
+                $maxLength - self::SMALLEST_HTML_LENGTH - Lorem::TOO_SHORT_TEXT_ARGUMENT
+            );
+            throw new \InvalidArgumentException($message);
+        }
+
+        $output         = "";
+        $htmlTags       = self::getHtmlTags();
+        //$htmlBlockTags  = $htmlTags['block'];
```
fzaninotto: Commented out code, please remove

src/Faker/Provider/Internet.php
```diff
@@ -331,4 +331,128 @@ public static function macAddress()
...
+        $textLength = $maxLength - $htmlTagsLength;
+        switch ($maxLength) {
+            case ($maxLength >= (self::SMALLEST_HTML_LENGTH + 200)):
+                /*
+                Since Lorem::text() output length is 99% always below than the request $maxNbChars
+                while looping over the tag list, generating the html output,
+                we recalculate the real remaining  text length to be outputted.
+                Makes the result as closer as possible from the max expected value
+                */
+                foreach ($htmlTags['block'] as $i => $tag) {
```
fzaninotto: very limited structure imo.

fzaninotto:
Although you respect the size, I see several drawbacks with this approach:

- you call `Lorem::test()` several times, it's not necessarily great for performance.
- The generated html has poor randomness as far as bloc elements are concerned.

Perhaps another approach would be to manipulate a DOM element instead of a string, make it grow incrementally until the serialized version has the requested size? That way you could add paragraphs, titles, semantically. What do you think?

src/Faker/Provider/Internet.php
ronanguilloux: ok

src/Faker/Provider/Internet.php
ronanguilloux: ok

src/Faker/Provider/Internet.php
ronanguilloux: ok, this will be done in the next commit

ronanguilloux:
I'm very OK to turn the approach around, manipulating a DOM element instead of a string. Thanks for this suggestion!

I propose a two-step approach : 
- first, resolving this single DOM element generation use case. A `Html` provider would then allow to generate single DOM elements with a given `$maxLength` parameter, and maybe an optional random global HTML5 attribute. 
- second, allowing then to generate a collection of DOM elements, or even an embedded collection of DOM elements, such as a whole HTML document.

Is that okay? 

A question: In a DOM generation context, I'm not sure that having a fixed length output string make sense. Must the `$maxLength` parameter apply to the whole final outputted string, tags & attributes included, rather than just the `innerText` of the DOM element ? AFAIC, I would choose the latter, because giving an order of magnitude (grandeur) of the output I need is 99% sufficient, while waiting for a exact fixed HTML string length hardly ever occurs to me.
Any feedback, opinion or thought is very welcome.

fzaninotto:
I think your approach is OK. As for the length, I think users will expect that to concern the entire serialized DOM, not the innerText. The most common use case is probably: I have this "body" field in my database table which is 3000 chars at most, I want to fill it with valid HTML. 

A way to achieve that is to serialize the DomDocument at every iteration, and grow the tree as long as it doesn't reach the limit (or a random percentage of the limit). It's slow, but I don't see any other way.

ronanguilloux:
OK for that.
WIP!

vinkla:
Any updates on this? Would be a great feature to add!

ronanguilloux:
Hi @vinkla, thanks for reviving this PR

- The [work in progress is here](https://github.com/ronanguilloux/Faker/tree/htmlLorem) (htmlLorem branch). I didn't had time to push it forward recently.
- Generating simple content inside a random HTML tag with random attributes is OK (see the public `htmlTag` function and related protected func it uses)
- The deal is now to generated a piece of DOM with a given - precise - length, that would concern this entire serialized DOM: The use case here was described by François in this [comment](#issuecomment-76206084). He also gently proposed a way to do it, in an iterative (slow but effective) way.

Please feel free to PR inside my working branch, your commits will be credit the same.

vinkla:
@ronanguilloux Okay, great. So this PR is accepted and just not finished yet?

ronanguilloux:
Well the proposition is accepted and the choosen way to achieve it was confimed, but PR is incomplete (not accepted yet) and, yes, to be finished.

fzaninotto:
Abandoned PR, closing.

PR #517, opened 2015-02-11
Adding it_IT Text Provider

endelwar:
Added Dante Alighieri's Inferno to the italian text provider

### DIFF
--- a/src/Faker/Provider/it_IT/Text.php
+++ b/src/Faker/Provider/it_IT/Text.php
@@ -1,0 +1,6958 @@
+<?php
+
+namespace Faker\Provider\it_IT;
+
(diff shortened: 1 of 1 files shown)


fzaninotto:
I think a more modern text in prose would produce better results...

endelwar:
What about Pinocchio by Collodi?

fzaninotto:
Better idea !

davelopware:
By total coincidence, I've just done the Dante's Inferno Text.php file myself too, and came here to offer it as a pull :)

endelwar:
I forgot to report here that the text was updated @fzaninotto

fzaninotto:
Thanks!

PR #519, opened 2015-02-15
Add properties for autocomplete

Tom5om:
Missing faker address properties added, so for example phpstorms autocomplete will find them

### DIFF
--- a/src/Faker/Generator.php
+++ b/src/Faker/Generator.php
@@ -7,6 +7,10 @@ namespace Faker;
  * @property string $firstName
  * @property string $lastName
  *
+ * @property string $cityPrefix
+ * @property string $secondaryAddress
+ * @property string $state
+ * @property string $stateAbbr
  * @property string $citySuffix
  * @property string $streetSuffix
  * @property string $buildingNumber


fzaninotto:
These ones are not in the main `Address` provider, so they're not implemented in every locale. That's why they aren't in the autocomplete comments.

Adding them would trick users into thinking that they can use these methods, while they are probably not available in their locale.

So thanks, but I will close this one without merging.

Tom5om:
Ok, but could we not somehow say that in the comments? Because now it looks
like no one can use them?
Something like :
+ * @property string $cityPrefix *Note this is not for all locales

Thanks for your explanation anyways!


On Thu, Feb 19, 2015 at 7:27 PM, Francois Zaninotto <
notifications@github.com> wrote:

> These ones are not in the main Address provider, so they're not
> implemented in every locale. That's why they aren't in the autocomplete
> comments.
>
> Adding them would trick users into thinking that they can use these
> methods, while they are probably not available in their locale.
>
> So thanks, but I will close this one without merging.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/fzaninotto/Faker/pull/519#issuecomment-75014056>.
>



-- 
Tom Somerville

fzaninotto:
It would be hard to maintain and confusing IMO. Locale-specific formatters are in the README, not in autocomplete.

PR #521, opened 2015-02-18
Improve phonenumber formats for nl_NL

SpaceK33z:
A lot of the previous generated phonenumbers were invalid, and therefore some formatters like [libphonenumber](https://github.com/googlei18n/libphonenumber) throw an error. This is fixed with this PR.


fzaninotto:
Thanks!

PR #526, opened 2015-02-27
Feature : Add dateTimeInInterval Method

nicodmf:
Permit to take a date and add or substract a php recognized interval.

I'm sure this can be usefull, because i write that in one of my alice loader, but not if this can be look as good writed (long in comparison with other pretty date functions) or in your waited tasks list.

So let me know if you want i finish with csfixer, tests, rebase and all or if this is not usefull.

@see http://php.net/manual/fr/dateinterval.createfromdatestring.php
@see http://php.net/manual/fr/datetime.formats.relative.php

@note : edited directly on github for moment

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -146,6 +146,7 @@ Each of the generator properties (like `name`, `address`, and `lorem`) are calle
     date($format = 'Y-m-d', $max = 'now') // '1979-06-09'
     time($format = 'H:i:s', $max = 'now') // '20:49:42'
     dateTimeBetween($startDate = '-30 years', $endDate = 'now') // DateTime('2003-03-15 02:00:49')
+    dateTimeInInterval($startDate = '-30 years', $interval = '+ 5 days') // DateTime('2003-03-15 02:00:49')
     dateTimeThisCentury($max = 'now')     // DateTime('1915-05-30 19:28:21')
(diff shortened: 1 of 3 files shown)


fzaninotto:
What's the difference with `dateTimeBetween()`? You just pass a date and an interval instead of passing two dates?

nicodmf:
Yes that the idea. 

And i can pass a date from another field with alice and not an exact date.

Exemple :

```yaml

EntityBundle\Entity\HourSheet:
    Hours{1..1000}:
        start: <dateTimeBetweenInterval('-5 days', '+10 days')>
        end: <dateTimeBetweenInterval($start, '+5 hours')>
        validated: 50%? <dateTimeBetweenInterval($end, '+15 days')>
        sended: 50%? <dateTimeBetweenInterval($validated, '+15 days')>
        activity: @OnlyActivity*
        user: @User*
        yard: @Yard*

```

fzaninotto:
Ok, I can understand the requirement. 

To avoid code duplication, please use `dateTimeBetween()` in your code.

Also, please add a reference of the new formatter in the README.

nicodmf:
@fzaninotto Code rebased, tested and the readme is updated. I can make another changes if needed?

fzaninotto:
Thanks!

PR #533, opened 2015-03-11
add generator for nif

jjbier:


fonsecas72:
Just wanted to say that it seems that the Spanish NIF is generated in the same way as the Portuguese NIF. Maybe we should give it a thought.
The tests are failing because of CS btw.

fzaninotto:
Yep, please fix CS and add a line to the locale-specific section in README

fzaninotto:
Bump. Can you please update your PR according to my comments?

fzaninotto:

PR #540, opened 2015-03-18
Updated es_AR Address.php provider

ivanmirson:
CitySuffix should be separated by a space from the name.
The building number should be after the street name.


fzaninotto:
Thanks!

PR #335, opened 2014-05-15
Improving pt_BR Providers

igorsantos07:
- In overall, code extensibility was improved;
- Added some Brazilian documents to the Person provider;
- Phone generation was greatly improved, being able to differentiate between landline and cellphones + additional local rules for valid numbers.

Also, I've added samples to the repository README :)

PR #545, opened 2015-03-22
Improving pt_BR Providers

igorsantos07:
- In overall, code extensibility was improved;
- Added some Brazilian documents to the Person provider;
- Phone generation was greatly improved, being able to differentiate between landline and cellphones + additional local rules for valid numbers.
- added samples to the repository README

New version of #335, with CS and other fixes


src/Faker/Provider/pt_BR/Person.php
```diff
@@ -90,7 +90,7 @@ class Person extends \Faker\Provider\Person
```
fzaninotto: The file permissions have changed, please revert to 644.

src/Faker/Provider/pt_BR/Person.php
```diff
@@ -99,4 +99,56 @@ public static function suffix()
     {
         return static::randomElement(static::$suffix);
     }
+
+
+    /**
+     * A random CPF number.
```
fzaninotto: Could you link to a definition of the format, e.g. in Wikipedia? Same for cnpj.

src/Faker/Provider/pt_BR/Person.php
```diff
@@ -99,4 +99,56 @@ public static function suffix()
...
+     * @param bool $formatted If the number should have dots/dashes or not.
+     * @return string
+     */
+    public function cpf($formatted = true)
+    {
+        $n = $this->generator->numerify('#########');
+        $n .= $this->verifierDigit($n);
```
fzaninotto: If you always double it, why don't the `verifierDigit()` return it twice instead?

igorsantos07: I do it that way because the first digit verifies the document number, and the second one verifies the document number and the first digit itself. Thus, for modularity, I've created a method that only calculates a digit, instead of generating all the needed digits. It could be used later for other types of documents if they use the same modulus 11 algorithm, such as driver's license or RG (yes, we brazilians do have two different document numbers for identity).

igorsantos07:
It seems the Travis build failed because of composer issues, on PHP 7.

src/Faker/Provider/pt_BR/Person.php
```diff
@@ -99,4 +99,59 @@ public static function suffix()
...
+    }
+
+    /**
+     * A random CNPJ number.
+     * @link http://en.wikipedia.org/wiki/CNPJ
+     * @link http://pt.wikipedia.org/wiki/CNPJ#Algoritmo_de_Valida.C3.A7.C3.A3o
+     * @param bool $formatted If the number should have dots/slashes/dashes or not.
+     * @return string
+     */
+    public function cnpj($formatted = true)
```
hussani: CNPJ is a Company property, not Person.
Can you move it?

igorsantos07: I'm not really sure if we need to create a whole new provider just to generate company documents. Even more if you consider that's a juridic person document, same as the CPF for a physical person.

hussani: Here are the company provider.
https://github.com/fzaninotto/Faker/blob/master/src/Faker/Provider/pt_BR/Company.php

A juridic person is a company, how said on Wikipedia, CNPJ is an identification number to companies and legal entities. I think is more logical move this property to company provider.

> Eg.
> I, as a person, have a CPF
> I, as a company, have a CNPJ

I don't know if it make sense for you. Please, tell me what you think about it.

oliveiraev:
I, particularly, disagree with an RG generator. RG doesn't follows any pattern across the states. While RG's in São Paulo matches ```\d{1,3}\.\d{3}\.\d{3}-[0-9x]```, RG's in Minas Gerais follow something near ```\d{4,}[A-Z]``` and no state provides any validator or algorithm checker to avoid generate real persons documents.

igorsantos07:
I guess there's no such a thing as an algorithm to avoid real documents
whatsoever.

Usually RG is accepted as plain text across systems. On top of that, it's
semantically better to have something like an RG method instead of a random
number generator.

If someone needs a specific state's version, one can simply PR an algorithm
flag :)
I should also add that the algorithm used covers SP and RJ needs, to say
the least.

igorsantos07:
Any news on this, @fzaninotto?

test/Faker/Provider/pt_BR/PersonTest.php
```diff
@@ -0,0 +1,35 @@
...
+        $cpf = $this->faker->cpf(false);
+        $this->assertRegExp('/\d{9}\d{2}/', $cpf);
+        $cpf = $this->faker->cpf(true);
+        $this->assertRegExp('/\d{3}\.\d{3}\.\d{3}-\d{2}/', $cpf);
+    }
+
+    public function testRgFormatIsValid()
+    {
+        $rg = $this->faker->rg(false);
+        echo $rg."\n";
```
fzaninotto: please remove

test/Faker/Provider/pt_BR/PersonTest.php
```diff
@@ -0,0 +1,35 @@
...
+        $this->assertRegExp('/\d{8}\d/', $rg);
+        $rg = $this->faker->rg(true);
+        echo $rg."\n";
```
fzaninotto: please remove

fzaninotto:
Good PR, I'll merge as soon as you fix the two minor debug statements.

igorsantos07:
Whoops! Fixed now :)

igorsantos07:
@fzaninotto, stuff is fixed ;)

fzaninotto:
Thanks!

PR #546, opened 2015-03-23
Update IPv6-method in Internet.php

hnesland:
Keep example IPv6-address within scope designated for documentation etc.


fzaninotto:
Not sure I understand. The documentation doesn't mention unicast. This should be offered as an option, otherwise the behavior differs from the `ipv4()` provider.

hnesland:
Allright, well I just figured it would be nice to not use adresses which could be mistaken for being real :-)

fzaninotto:
The very purpose of Faker it to provide real-looking data ;)

Issue #539, opened 2015-03-18
copy(): The first argument to copy() function cannot be a directory


montogeek:
I get that error doing 
`$faker->file()`

The `/tmp` folder exists.
PHP 5.4.38 (cli) (built: Feb 20 2015 02:42:34)
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.4.0, Copyright (c) 1998-2014 Zend Technologies

CentOS 6.2

vuchl:
What are your parameters for ```$faker->file();``` like ```$faker->file($sourceDir, $targetDir, false)```

montogeek:
I didnt use any parameters, just `file()`

fzaninotto:
`file()` returns a random file from the source directory (here `/tmp). I suspect it fails because this directory is empty, am I right?

fzaninotto:
Sorry, I just understood the source of the problem (#547). Continuing discussion there.

fzaninotto:
Fixed by #547

PR #547, opened 2015-03-25
Fix for 'The first argument to copy() function cannot be a directory'

svrnm:
This is a fix for the issue #539. Directories are excluded using the provided filter.


fzaninotto:
Thanks for the patch, can you fix coding standards (see build report)?

svrnm:
I fixed it and double checked it with phpspec:-)

fzaninotto:
Thanks!

montogeek:
Thanks!

PR #550, opened 2015-03-26
Update the phpdoc for magic methods

stof:
This adds missing magic methods in the Generator for new providers and adds arguments for methods to avoid warnings about unused arguments when using them.

It also fixes a few return types to be consistent with the implementation.


stof:
Note that I did this update manually. But it could be possible to write a script automating this update to make it easier to keep things uptodate.

fzaninotto:
Thanks a lot!

PR #488, opened 2014-12-30
Added brazilian RG and CPF support

ianrodrigues:
* RG is a brazilian ID card number
* CPF is like a registry of contributors' identification

PR #551, opened 2015-03-31
[pt_BR][Person] Adding brazillian RG and CPF

hussani:
- RG is a brazilian ID card number
- CPF is like a registry of contributors' identification

This PR is a quick fix of #488, initialized by @ianrodrigues.
All credits belong to him.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -864,6 +864,23 @@ echo $faker->taxpayerIdentificationNumber; // '165249277'
 
 ```
 
+### `Faker\Provider\pt_BR\Person`
+
+```php
+<?php
+
+// Generates a random identification number (in portuguese - Registro Geral RG do Estado de São Paulo) - based on SSP-SP algorithm
+echo $faker->rg; // '84.405.736-3'
+// or as integer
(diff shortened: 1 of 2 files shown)


igorsantos07:
Is the code working? The samples doesn't seems to reflect the code behaviour - furthermore, that behaviour of passing a string into a provider's method is uncommon across the library.

Finally, considering there's an year-old PR (#545) about brazilian documents and other improvements, maybe we should add the RG to it instead of having two PR on the same matter?

hussani:
@igorsantos07 I just needed a RG and CPF provider and I saw the open PRs, found one and fix.
Your solution is larger, have some details inexistent on this PR (like CNPJ and others) and the code is more organized.
Personally, prefer your solution than this. I would be glad if your code was merged instead that. :+1:

igorsantos07:
You can make a PR to igorsantos07:improved-pt-cherry-picked and I'll merge it. That will include your commit into #545 :)

fzaninotto:
Superseded by #545. Please collaborate with @igorsantos07

PR #553, opened 2015-04-04
Russian Bank names provider

wizardjedi:
Hi, folks.

I've just implemented generator for Russian Bank names. Real bank names you can get from banki.ru  (see complete list: http://www.banki.ru/banks/)


wizardjedi:
Added description to readme.md

src/Faker/Provider/ru_RU/Payment.php
```diff
@@ -0,0 +1,811 @@
...
+
+class Payment extends \Faker\Provider\Payment
+{
+    /**
+     * @see list of Russian banks (2015-04-04), source: http://www.banki.ru/banks/
+     * @example "cat *.html | grep 'b-cb-list__name' | iconv --f windows-1251 --t utf-8 | grep -o '>.*<' | \
+     * sed -r 's/&mdash;//' | sed -r 's/[\<\>]//g' | sed -r "s/(^|$)/'/g" | sed -r 's/$/,/' | sed -r 's/\&(laquo|raquo);/"/g' | \
+     * sed -r 's/\s+/ /g'"
+     */
+    protected static $banks = array(
```
nineinchnick: If that list would be indexed by bank code it could be used to generate an account number for a specific bank. This could be useful to verify that a bank name detection mechanism in an application works correctly.

wizardjedi: You absolutly right! But i've no enough information about bank odes in russia. It seems that there are multiple bank registers for banks. So it's hard to get bank id.

That's why i didn't set any ids for banks. Maybe i'll add bank codes in future.

wizardjedi:
Hi.

Any comments?

PR #554, opened 2015-04-07
Update Person.php

tassoman:
A better italianische personification ;)

### DIFF
--- a/src/Faker/Provider/it_IT/Person.php
+++ b/src/Faker/Provider/it_IT/Person.php
@@ -23,41 +23,42 @@ class Person extends \Faker\Provider\Person
     );
 
     protected static $firstNameMale = array(
-        'Aaron', 'Abramo', 'Adriano', 'Akira', 'Alan', 'Alberto', 'Albino', 'Alessandro', 'Alessio', 'Amedeo', 'Amos', 'Anastasio', 'Anselmo',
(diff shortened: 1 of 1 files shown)


fzaninotto:
Would you mind adding a reference for the data source? It makes future updates easier for other maintainers

tassoman:
Well, I've tried tried to guess. But if you need a referenced source, there's some Open Data about it.
http://www.istat.it/en/products/interactive-contents/baby-names 
If you feel interesting I could try to merge

fzaninotto:
I would love to have a list referencing a single source. That would ease maintenance so much.

tassoman:
I've found this link, 
http://demo.istat.it/altridati/IscrittiNascita/2013/T2.9.xls

It's a yearly report of 30 most used names for babies born in Italy

fzaninotto:
We need to reference a web page mentioning the content license, and it should be compatible with MIT.

tassoman:
Open Data licence from Istat website is described at http://www.istat.it/en/about-istat/activities/open-data they chose CC Attribute 3.0 but I can't say if it's compatible with MIT.

Feel free to merge my PR there's no link between my own activity and Istat open-data distributors.

fzaninotto:
If we can't reference an external source, I'm not sure your PR has a high value. Another contributor could argue that the names you provided aren't better than the ones in place, I have no way to tell which one is better. 

I really want to have a reference for e

PR #556, opened 2015-04-09
Add my name in pt_BR

tassianoalencar:


fzaninotto:
Not much value in this PR. Every time someone changes the fixtures, it may break tests based on a given seed. This kind of PR breaks compatibility for not enough benefit. Closing.

PR #560, opened 2015-04-15
It's supposed to be a title after a last name

killerog:
Because the name created in #559 seems really weird, it being a title is more logical.


fzaninotto:
Thanks!

PR #563, opened 2015-04-25
Add `pt_PT` address provider readme information

nunorafaelrocha:


fzaninotto:
You are documenting generic providers in a locale-specific doc section. Faker doc doesn't repeat the doc for every locale; the locale doc only lists locale-specific providers.

PR #565, opened 2015-04-28
Fixed file provider to ignore unreadable and special files

svrnm:
After fixing #547 I found out that there are some other cases which will cause ```copy``` to fail: unreadable files and special files (e.g. pipes).

The proposed changes should fix this issue and additionally I added an exception if the chosen directory is empty.


fzaninotto:
Thanks!

PR #570, opened 2015-05-04
min / max for lat long values

actuallymab:
Sometimes you may want to set lat/lng parameters for a spesific area. 

Ex: 
For Istanbul:
'lat' => $faker->latitude(40.958836, 41.136456),
'lng' => $faker->longitude(28.64, 29.98),

### DIFF
--- a/src/Faker/Provider/Address.php
+++ b/src/Faker/Provider/Address.php
@@ -104,20 +104,24 @@ class Address extends \Faker\Provider\Base
     }
 
     /**
-     * @example 77.147489
-     * @return float Uses signed degrees format (returns a float number between -90 and 90)
+     * @example '77.147489'
+     * @param int $min
+     * @param int $max
+     * @return string
      */
-    public static function latitude()
+    public static function latitude($min = -90, $max = 90)
     {
-        return static::randomFloat(6, 0, 180) - 90;
+        return number_format(mt_rand($min * 1000000, $max * 1000000)/1000000, 6);
     }
 
     /**
-     * @example 86.211205
-     * @return float Uses signed degrees format (returns a float number between -180 and 180)
+     * @example '86.211205'
+     * @param int $min
+     * @param int $max
+     * @return string
      */
-    public static function longitude()
+    public static function longitude($min = -180, $max = 180)
     {
-        return static::randomFloat(6, 0, 360) - 180;
+        return number_format(mt_rand($min * 1000000, $max * 1000000)/1000000, 6);
     }
 }
--- a/src/Faker/Provider/Address.php
+++ b/src/Faker/Provider/Address.php
@@ -105,23 +105,23 @@ class Address extends \Faker\Provider\Base
 
     /**
      * @example '77.147489'
-     * @param int $min
-     * @param int $max
-     * @return string
+     * @param float|int $min
+     * @param float|int $max
+     * @return float Uses signed degrees format (returns a float number between -90 and 90)
      */
     public static function latitude($min = -90, $max = 90)
     {
-        return number_format(mt_rand($min * 1000000, $max * 1000000)/1000000, 6);
+        return floatval(number_format(mt_rand($min * 1000000, $max * 1000000)/1000000, 6));
     }
 
     /**
      * @example '86.211205'
-     * @param int $min
-     * @param int $max
-     * @return string
+     * @param float|int $min
+     * @param float|int $max
+     * @return float Uses signed degrees format (returns a float number between -180 and 180)
      */
     public static function longitude($min = -180, $max = 180)
     {
-        return number_format(mt_rand($min * 1000000, $max * 1000000)/1000000, 6);
+        return floatval(number_format(mt_rand($min * 1000000, $max * 1000000)/1000000, 6));
     }
 }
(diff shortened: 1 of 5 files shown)


src/Faker/Provider/Address.php
```diff
@@ -104,20 +104,24 @@ public static function country()
     }
 
     /**
-     * @example 77.147489
-     * @return float Uses signed degrees format (returns a float number between -90 and 90)
+     * @example '77.147489'
+     * @param int $min
```
fzaninotto: it's a `float`, not an `int`, right?

src/Faker/Provider/Address.php
```diff
@@ -104,20 +104,24 @@ public static function country()
...
+     * @param int $max
+     * @return string
      */
-    public static function latitude()
+    public static function latitude($min = -90, $max = 90)
     {
-        return static::randomFloat(6, 0, 180) - 90;
+        return number_format(mt_rand($min * 1000000, $max * 1000000)/1000000, 6);
```
fzaninotto: why don't you use `random_float` anymore?

fzaninotto:
Please also update the README to add an example with the new parameter

src/Faker/Provider/Address.php
aydnbhdr: this is a coordinate, not a "just" float number. So I decided to update lat lng functions

src/Faker/Provider/Address.php
aydnbhdr: Sorry for missing that, Phpstorm decided it, not me :)

src/Faker/Provider/Address.php
```diff
@@ -104,20 +104,24 @@ public static function country()
...
+     * @example '77.147489'
+     * @param float|int $min
+     * @param float|int $max
      * @return float Uses signed degrees format (returns a float number between -90 and 90)
      */
-    public static function latitude()
+    public static function latitude($min = -90, $max = 90)
     {
-        return static::randomFloat(6, 0, 180) - 90;
+        return floatval(number_format(mt_rand($min * 1000000, $max * 1000000)/1000000, 6));
```
fzaninotto: I don't understand the interest of converting a float to string (with `number_format`) and then back to float (with `floatval`)

fzaninotto:
Tests fail

src/Faker/Provider/Address.php
fzaninotto: Still, `static::randomFloat()` seems to be a better option:

```php
// not that clear
return mt_rand($min * 1000000, $max * 1000000)/1000000;
// better
return static::randomFloat(6, $min, $max);
```

fzaninotto:
Also, please see #416 for why you should use `randomFloat()`

ikwattro:
@fzaninotto @actuallymab Any update on this PR, we are really interested in this feature

actuallymab:
don't know.. @fzaninotto ok with this?

src/Faker/Provider/Address.php
```diff
@@ -104,20 +104,24 @@ public static function country()
...
 
     /**
-     * @example 77.147489
+     * @example '77.147489'
+     * @param float|int $min
+     * @param float|int $max
      * @return float Uses signed degrees format (returns a float number between -90 and 90)
      */
-    public static function latitude()
+    public static function latitude($min = -90, $max = 90)
```
fzaninotto: You're changing the results of existing calls. Please use 0 and 180 as min and max values.

src/Faker/Provider/Address.php
```diff
@@ -104,20 +104,24 @@ public static function country()
...
 
     /**
-     * @example 86.211205
+     * @example '86.211205'
+     * @param float|int $min
+     * @param float|int $max
      * @return float Uses signed degrees format (returns a float number between -180 and 180)
      */
-    public static function longitude()
+    public static function longitude($min = -180, $max = 180)
```
fzaninotto: You're changing the results of existing calls. Please use 0 and 360 as min and max values.

src/Faker/Provider/me_ME/Address.php
```diff
@@ -111,17 +111,23 @@ public function cityName()
 
     /**
      * @example '77.147489'
+     * @param float|int $min
+     * @param float|int $max
+     * @return float Uses signed degrees format (returns a float number between -90 and 90)
      */
-    public static function latitude()
+    public static function latitude($min = 42.43, $max = 42.45)
```
fzaninotto: `latitude()` and `longitude()` shouldn't be localized in the first place. This is a mistake in the existing `me_ME` formatters. Please create a new formatter, like `localGeolocation()`, for that purpose. That way, you can generate latite and longitude within a polygon.

src/Faker/Provider/Address.php
TimWolla: He isn't. Look again.

fzaninotto: My bad, you're right.

src/Faker/Provider/Address.php
fzaninotto: Disregard.

src/Faker/Provider/me_ME/Address.php
fzaninotto: This is still an issue.

fzaninotto:
Thanks

PR #572, opened 2015-05-13
Update readme.md

totophe:
Adding instruction for installing with composer


fzaninotto:
Thanks!

totophe:
Ur welcome :-) seems obvious but not for everyone apparently.

PR #574, opened 2015-05-13
Added CPF and CNPJ in pt_BR provider

acfreitas:


src/Faker/Provider/Imei.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace Faker\Provider;
+
+class Imei
+{
+
+    /**
+     * International Mobile Equipment Identity (IMEI)
```
fzaninotto: Should not be part of this PR, please remove

fzaninotto:
#545 already does that, duplicate.

acfreitas:
Sorry, I created a new PR to IMEI.

PR #582, opened 2015-05-19
Add International Mobile Equipment Identity (IMEI)

acfreitas:
Removed from other PR #574 and create a new PR for Provider IMEI


ronanguilloux:
I @acfreitas, thanks a lot for your contribution!

IMEI generation/validation uses Luhn check algorithm: I suggest to use the  `src/Faker/Calculator/Luhn` calcultor instead.

Unit tests should be appreciated too!

acfreitas:
@ronanguilloux, I created the tests and edited the code to use ```src/Faker/Calculator/Luhn```. Thanks for your suggestion.

src/Faker/Provider/Imei.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Faker\Provider;
+
+use Faker\Calculator\Luhn;
+
+class Imei extends Base
```
fzaninotto: I suggest you move that to the PhoneNumber provider

acfreitas: @fzaninotto. the IMEI is used in some mobile devices, not just Phone. Do you think that move to PhoneNumber Provider is necessary?

fzaninotto: Let's consider that `PhoneNumber` is named `Phone` (which should have been the case from the beginning), so yes, this is necessary.

fzaninotto:
Merged using the command line in 54e105fedd5715fb8070afb88334f212b6f2a84a. Thanks!

PR #590, opened 2015-05-24
Fix IBAN generator

okj579:
This fixes:
* IBAN formats for Kuwait and Turkey (source: http://www.swift.com/dsp/resources/documents/IBAN_Registry.pdf)
* Checksum calculation for IBANs with letters
* Off by one error in format boundaries

### DIFF
--- a/src/Faker/Provider/Payment.php
+++ b/src/Faker/Provider/Payment.php
@@ -231,7 +231,7 @@ class Payment extends Base
         }
 
         $expandedFormat = '';
-        foreach($format as list($class, $length)) {
+        foreach ($format as list($class, $length)) {
             $expandedFormat .=  str_repeat($class, $length);
         }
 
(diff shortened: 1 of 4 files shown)


src/Faker/Provider/Payment.php
```diff
@@ -225,22 +225,20 @@ protected static function iban($countryCode, $prefix = '', $length = null)
                 }
             }
         }
+        if ($format === null) {
+            return false;
```
fzaninotto: please remove

fzaninotto:
Please add unit tests to prove the bugs, and to show that your PR fixes them.

okj579:
Ok, I'm working on this. The format is not to hard to test, but to test the checksum, the generator code would essentially have to be copied into the unit test, which sort of seems to defeat the purpose. An IBAN Checker (e.g. [okj579/swiftery](https://github.com/okj579/swiftery) or [globalcitizen/php-iban](https://github.com/globalcitizen/php-iban)) could be added to the project to do the testing. This seems like a lot for a unit test, but I think probably the best option. Should I do that?

fzaninotto:
You should move the checksum calculation to a standalone Calculator class (see [the Luhn class](https://github.com/fzaninotto/Faker/blob/a3aad09dc8b0f4767e113fc2a7e8fd2ccb11aa65/src/Faker/Calculator/Luhn.php)), and reuse that class in your test.

okj579:
That's a much better idea. I've refactored it, so the checksum logic is done in Faker\Calculator\Iban (it is its own algorithm), which can also be tested with unit tests.

fzaninotto:
Awesome, thanks!

PR #596, opened 2015-06-04
added Miscellaneous class for en_US'

lamaiac:
added Miscellaneous class for en_US

### DIFF
--- a/src/Faker/Provider/en_US/Miscellaneous.php
+++ b/src/Faker/Provider/en_US/Miscellaneous.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace Faker\Provider\en_US;
+
+class Miscellaneous extends \Faker\Provider\Base
+{
+    /**
+     * Return a boolean, true or false
+     *
+     * @param integer $chanceOfGettingTrue Between 0 (always get false) and 100 (always get true).
+     * @return bool
+     * @example true
+     */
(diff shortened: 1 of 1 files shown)


src/Faker/Provider/en_US/Miscellaneous.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+namespace Faker\Provider\en_US;
+
+class Miscellaneous extends \Faker\Provider\Base
```
fzaninotto: You must extend `\Faker\Provider\Miscellaneous` instead

fzaninotto:
In fact, I don't understand this PR. Miscellaneous provides random data for language, etc. Your PR implies that these formatters won't be random anymore for the en_US locale. It doesn't make sense.

src/Faker/Provider/en_US/Miscellaneous.php
lamaiac: ok

lamaiac:
@fzaninotto -- if the code specifies the locale, I want to use the locale that was specified.  As Faker sits, now, when I run this command:
*$faker = \Faker\Factory::create('en_US');*
I expect to get data for the US, but I get random country choices.

fzaninotto:
Your `country` formatter returns "USA". You don't need Faker for that:

```php
echo 'USA';
```

Faker is a library generating random data.

PR #597, opened 2015-06-07
Add Romanian text

nevvermind:


fzaninotto:
Rhymes don't produce the best random text. I'd prefer a prose text, if possible a contemporary one.

Issue #599, opened 2015-06-09
Dates before 1901


ethanclevenger91:
I noticed in `dateTimeBetween()`, the random date is being generated using timestamps and `mt_rand()`. The issue here is that using either `strtotime()` or DateTime's `getTimestamp()` bust on dates before 1901 (you run into a 32-bit integer issue, in [the notes for `strtotime()`](http://php.net/manual/en/function.strtotime.php). 

Looks like to get around this you could force a DateTime object and then [use DateTime's `format()`](http://php.net/manual/en/datetime.gettimestamp.php#98374). 

I can work on a PR a little later on, but I want to be thorough, as I'm sure this is an issue elsewhere as well.

ethanclevenger91:
This also introduced an issue because `mt_rand()` suffers the same drawbacks. One thought was to get the difference between the two timestamps, do `mt_rand(0, $difference)` and add that to the start timestamp. However, if you have a very _wide_ range of dates, this actually _exacerbates_ the problem because your int is even larger.

Next thought is to generate a random date piece-by-piece (randomize year, then month, than day...)

fzaninotto:
Wait, I don't understand. `DateTime::getTimestamp()` doesn't even work for dates before the Unix epoch (1970). So `dateTimeBetween()` should simply send an error if the `$startDate` is below that date.

ethanclevenger91:
IIRC I wasn't getting errors, I was just getting dates only after the epoch. So the title of this would actually more appropriately be "Dates before the epoch". The PR adds support for a much wider range of days

ethanclevenger91:
Further thoughts on this?

thealonekhan:
hi, i have an issue while converting a giving date or time for a specific timezone, using faker DateTime.
The issue is: it always convert the date for a given timezone, exact one hour prior.
This is my code:
            $date = new \DateTime($start_time, new \DateTimeZone('GMT'));
            $date->setTimezone(new\ DateTimeZone('EDT'));
            $start_time = $date->format('H:i');

Now the $start_time will be one hour earlier than given timezone Eastern Daylight Time(EDT).
Any suggestion?

mstnorris:
Can't generate; for example: _date of births_ before the epoch. I thought about manually creating the different elements of the date/timestamp but that seems to be overkill for something like this.

jmauerhan:
"Next thought is to generate a random date piece-by-piece (randomize year, then month, than day...)"

[...]

PR #600, opened 2015-06-09
Altered date_between() to avoid 32-bit limits and support a wider and…

ethanclevenger91:
… older range of random dates.

This is gross, but it now avoids the limits of `mt_rand` by first randomizing the date clamped to the min and max and then randomizing the time (clamped to the given min and max times if the date happened to fall on one of them). I also updated `dateTimeAD()` to utilize this, since it appeared that previously it was not, in fact, getting a random date since the beginning of time but rather operating within `mt_rand()`'s limits.

I tested it a bit as part of the FakerPress plugin and it appears to be working quite nicely. If someone has a better solution or can explain how there wasn't an issue previously, please let me know. See #599

### DIFF
--- a/src/Faker/Provider/DateTime.php
+++ b/src/Faker/Provider/DateTime.php
@@ -12,17 +12,18 @@ class DateTime extends \Faker\Provider\Base
             return (int) $max;
         }
 
-        if (! $max instanceof \DateTime) {
+        if (!$max instanceof \DateTime) {
             $max = new \DateTime($max);
         }
 
-        return $max->format("U");
+        return $max->format('U');
     }
 
     /**
-     * Get a timestamp between January 1, 1970 and now
+     * Get a timestamp between January 1, 1970 and now.
      *
      * @param \DateTime|int|string $max maximum timestamp used as random end limit, default to "now"
+     *
      * @return int
      *
      * @example 1061306726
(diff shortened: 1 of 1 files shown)


src/Faker/Provider/DateTime.php
```diff
@@ -33,34 +34,42 @@ public static function unixTime($max = 'now')
...
      * @param \DateTime|int|string $max maximum timestamp used as random end limit, default to "now"
+     *
      * @example DateTime('2005-08-16 20:39:21')
+     *
      * @return \DateTime
      */
     public static function dateTime($max = 'now')
     {
-        return new \DateTime('@' . static::unixTime($max));
+        return new \DateTime('@'.static::unixTime($max));
```
fzaninotto: no, this is not the CS of Faker, please revert changes like this one.

src/Faker/Provider/DateTime.php
```diff
@@ -12,17 +12,18 @@ protected static function getMaxTimestamp($max = 'now')
...
+            $max = new \DateTime($max);
         }
 
-        return strtotime(empty($max) ? 'now' : $max);
+        return $max->format('U');
     }
 
     /**
-     * Get a timestamp between January 1, 1970 and now
+     * Get a timestamp between January 1, 1970 and now.
```
fzaninotto: No, this is not the CS of Faker, please revert changes like this one.

src/Faker/Provider/DateTime.php
```diff
@@ -33,34 +34,42 @@ public static function unixTime($max = 'now')
...
      *
      * @param \DateTime|int|string $max maximum timestamp used as random end limit, default to "now"
+     *
      * @example DateTime('1265-03-22 21:15:52')
+     *
      * @return \DateTime
      */
     public static function dateTimeAD($max = 'now')
     {
-        return new \DateTime('@' . mt_rand(-62135597361, static::getMaxTimestamp($max)));
```
fzaninotto: This method already works and is much faster than your `dateTimeBetween()` implementation. So your change just makes it slower. Please revert.

src/Faker/Provider/DateTime.php
```diff
@@ -100,29 +113,67 @@ public static function time($format = 'H:i:s', $max = 'now')
...
+        $startTimestamp = $startDate->format('U');
+        $endTimestamp = $endDate->format('U');
 
         if ($startTimestamp > $endTimestamp) {
             throw new \InvalidArgumentException('Start date must be anterior to end date.');
         }
 
-        $timestamp = mt_rand($startTimestamp, $endTimestamp);
+        //get difference between timestamps
+        $diff = $endTimestamp - $startTimestamp;
```
fzaninotto: This will already break for big numbers, right? You still exacerbate the problems.

the only way, I think, is to avoid using microseconds (the 'U' format) altogether.

src/Faker/Provider/DateTime.php
```diff
@@ -100,29 +113,67 @@ public static function time($format = 'H:i:s', $max = 'now')
...
     {
-        $startTimestamp = $startDate instanceof \DateTime ? $startDate->getTimestamp() : strtotime($startDate);
-        $endTimestamp = static::getMaxTimestamp($endDate);
+        if (!$startDate instanceof \DateTime) {
+            $startDate = new \DateTime($startDate);
+        }
+        if (!$endDate instanceof \DateTime) {
+            $endDate = new \DateTime($endDate);
+        }
+        $returnDate = $startDate;
```
fzaninotto: You need to `clone()`, otherwise you'll be altering the `$startDate` parameter, which is unexpected.

fzaninotto:
Interesting approach, but doesn't fix the problem completely I think.

I'm afraid this is very slow, could you benchmark your method?

Also, you should add unit tests to show that it works.

fzaninotto:
Any feedback on this one?

fzaninotto:
No news for six months, closing.

PR #387, opened 2014-08-12
Latitud/Longitud Perú

Oxicode:

PR #601, opened 2015-06-09
Latitude and Longitude added

glorand:
Latitude and Longitude added


fzaninotto:
See #387: latitude should not be localized, use a localLatitude instead.

PR #594, opened 2015-06-02
+1 VAT Provider w/ 40+ countries rules, fix #591

ronanguilloux:
I had to handle backward compatibility on 

- src/Faker/Provider/at_AT/Payment.php, `vat`method
- src/Faker/Provider/bg_BG/Payment.php, `vat`method

I propose to tag them as `@deprecated`. Existing unit tests are maintained.

PR #602, opened 2015-06-11
Add Dutch VAT number function

hiddeco:


hiddeco:
Not needed as of #594

PR #606, opened 2015-06-14
Improve name guesser logic using field length when available

watermanio:
I kept finding dummy users with Lorem Ipsum as their salutation. This change uses the field length to be a bit more intelligent when that data is known. I've also tried to guess a few more column name and types.

### DIFF
--- a/src/Faker/Guesser/Name.php
+++ b/src/Faker/Guesser/Name.php
@@ -15,7 +15,7 @@ class Name
 
     /**
      * @param string $name
-     * @param int $size
+     * @param int|null $size Length of field, if known
      * @return callable
      */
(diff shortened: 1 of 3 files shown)


src/Faker/Guesser/Name.php
```diff
@@ -13,7 +13,12 @@ public function __construct(\Faker\Generator $generator)
         $this->generator = $generator;
     }
 
-    public function guessFormat($name)
+    /**
+     * @param string $name
+     * @param int $size
```
GrahamCampbell: int|null

adwww: Done

fzaninotto:
Great patch, thanks!

PR #608, opened 2015-06-15
Brazilian credit card providers

igorsantos07:
Adding Brazilian credit card BINs and vendors.
The original BIN list produces some cards not used in Brazil.


fzaninotto:
Thanks!

PR #609, opened 2015-06-16
Add en_US implementation for bank information

okj579:
This adds formatters for US account and routing numbers.

https://en.wikipedia.org/wiki/Routing_transit_number#MICR_Routing_number_format


okj579:
Waiting on #607, then rebase, formatting fixes, etc.

PR #610, opened 2015-06-17
Removed non NL names

markmooibroek:


fzaninotto:
Could you find a reference for common first and last name, and grab data from that reference? Otherwise, nothing prevents any developer from adding or removing a name. I have no way of verifying if the change is good, and this breaks user tests based on seedable random data. I only want to modify the random data if I am sure it won't be modified again in a few weeks, so we need an outer reference.

fzaninotto:
Bump, will close shortly if no feedback.

fzaninotto:
Abandoned PR, closing.

PR #615, opened 2015-06-20
US phone numbers

### LINKED ISSUE #588: Phone numbers not formatting correctly
According to the documentation in the README,

echo $faker->phoneNumber;

Is supposed to output a U.S. phone number.  However, here is an example of a phone number I got using this command:

+30(9)4312207752

I also have had phone numbers with an area code or a prefix that starts with '1', which is not allowed in U.S. phone numbers.

okj579:
This implements the rules for the North American Numbering Plan (NANP). It does not generate numbers including so-called "easily recognizable codes" which are used for various special services and not assigned to geographic regions, for example 911 (emergency), 411 (directory), 511 (traffic information) or 800 (toll-free).

Toll-free numbers can be generated with `tollFreePhoneNumber()`.

Fixes #588

### DIFF
--- a/src/Faker/Provider/PhoneNumber.php
+++ b/src/Faker/Provider/PhoneNumber.php
@@ -9,8 +9,8 @@ class PhoneNumber extends \Faker\Provider\Base
     /**
      * @example '555-123-546'
      */
-    public static function phoneNumber()
+    public function phoneNumber()
     {
-        return static::numerify(static::randomElement(static::$formats));
+        return static::numerify($this->generator->parse(static::randomElement(static::$formats)));
     }
 }
--- a/src/Faker/Provider/en_US/PhoneNumber.php
+++ b/src/Faker/Provider/en_US/PhoneNumber.php
@@ -51,18 +51,20 @@ class PhoneNumber extends \Faker\Provider\PhoneNumber
     );
     protected static $tollFreeFormats = array(
         // Standard formats
-        '{{areaCode}}-{{exchangeCode}}-####',
-        '({{areaCode}}) {{exchangeCode}}-####',
-        '1-{{areaCode}}-{{exchangeCode}}-####',
-        '{{areaCode}}.{{exchangeCode}}.####',
+        '{{tollFreeAreaCode}}-{{exchangeCode}}-####',
+        '({{tollFreeAreaCode}}) {{exchangeCode}}-####',
+        '1-{{tollFreeAreaCode}}-{{exchangeCode}}-####',
+        '{{tollFreeAreaCode}}.{{exchangeCode}}.####',
     );
+
+    public function tollFreeAreaCode()
+    {
+        return self::randomElement(static::$tollFreeAreaCodes);
+    }
 
     public function tollFreePhoneNumber()
     {
-        $areaCode = self::randomElement(static::$tollFreeAreaCodes);
         $format = self::randomElement(static::$tollFreeFormats);
-
-        $format = str_replace('{{areaCode}}', $areaCode, $format);
 
         return self::numerify($this->generator->parse($format));
     }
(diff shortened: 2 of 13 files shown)


src/Faker/Provider/PhoneNumber.php
```diff
@@ -9,8 +9,8 @@ class PhoneNumber extends \Faker\Provider\Base
     /**
      * @example '555-123-546'
      */
-    public static function phoneNumber()
+    public function phoneNumber()
```
fzaninotto: This will break many locales overriding this method with a `static` one. If you change it here, you must change it in locales, too.

fzaninotto: Sorry, I just saw that you actually did that further down in the PR ;)

src/Faker/Provider/en_US/PhoneNumber.php
```diff
@@ -4,30 +4,118 @@
...
+        '1-{{areaCode}}-{{exchangeCode}}-####',
+        '{{areaCode}}.{{exchangeCode}}.####',
+    );
+
+    public function tollFreePhoneNumber()
+    {
+        $areaCode = self::randomElement(static::$tollFreeAreaCodes);
+        $format = self::randomElement(static::$tollFreeFormats);
+
+        $format = str_replace('{{areaCode}}', $areaCode, $format);
```
fzaninotto: I don't understand why you need that. Since there is an `areaCode`formatter, `$generator->parse()` will do the replacement.

src/Faker/Provider/en_US/PhoneNumber.php
```diff
@@ -4,30 +4,118 @@
...
+        return join('', $digits);
+    }
+
+    /**
+     * Generates a random digit, which cannot be $except
+     *
+     * @param int $except
+     * @return int
+     */
+    private static function randomDigitNot($except)
```
fzaninotto: Should not be in PhoneNumber but in the Base provider

src/Faker/Provider/en_US/PhoneNumber.php
```diff
@@ -4,30 +4,118 @@
...
+    {
+        $result = self::numberBetween(0, 8);
+        if ($result >= $except) {
```
fzaninotto: do you mean `==`?

Add unit tests to make sure you don't forget a case

fzaninotto:
Good PR, needs a few tweaks but definitely worth merging!

src/Faker/Provider/en_US/PhoneNumber.php
okj579: Only certain area codes are used for toll free numbers, so the areaCode formatter won't work here. It actually won't generate any toll free area codes at all.

src/Faker/Provider/en_US/PhoneNumber.php
okj579: No, to make it random among the 9 possibilities, it shifts the whole range over the selected nimber up. For example, if 5 is passed as the exception, it generates a digit from 0-8 then shifts anything over 5 up one to make the range 0-4,6-9.

src/Faker/Provider/en_US/PhoneNumber.php
fzaninotto: Then why don't you create a `tollFreeAreaCode` formatter?

src/Faker/Provider/en_US/PhoneNumber.php
fzaninotto: Gotcha. A unit test is necessary nonetheless.

src/Faker/Provider/en_US/PhoneNumber.php
okj579: I guess I thought that formatter would be _too_ simple, but I guess that's not actually a problem :)

fzaninotto:
Thanks!

PR #617, opened 2015-06-22
Add setter and getter to config default locale

vluzrmos:


vluzrmos:
Keeping the constant DEFAULT_LOCALE to compatibility

vluzrmos:
The methods could be called setLocale and getLocale, whatever. I think, setDefaultLocale and getDefaultLocale is better.

fzaninotto:
What's the use case for that? I fail to see any interest.

vluzrmos:
I've tried to do that on laravel core, but was rejected, I thinking in cases of you can define the default faker locale.

fzaninotto:
You can't define the default Faker locale. You can define the locale you want to use for a given set of fake data.

PR #618, opened 2015-06-23
Add ar_SA locale

ibrasho:


fzaninotto:
Tests fail, can you check?

fzaninotto:
Bump, I will close your PR shortly of you don't fix the tests

ibrasho:
I just updated the entries. This should resolve the issue.

fzaninotto:
Why did you repeat the `realText()` method in the `Text` provider?

ibrasho:
If I don't, won't it default to faking text from the en_US locale?

fzaninotto:
You just need to override the data in the `Text` provider, not the `realText()` method (which is inherited).

ibrasho:
I've updated the PR and also updated the ar_JO Text provider.

fzaninotto:
Thanks!

PR #621, opened 2015-06-24
Add lebanon cities in Arabic

Mahmoudz:
Add Lebanon cities names in the Arabic Language for Address.
Add my Arabian name in the names list.


fzaninotto:
The locale is ar_JO, it doesn't make sense to add lebanon city names. Please create a lebanon locale.

Also, adding each person's name isn't the type of patch I agree to merge. It's too much noise and not enough added value.

Mahmoudz:
Oh I didn't knew about this `ar_JO` thingy I will consider making an `ar_LB` in the future, as I will be heavily using this package in my next project.

Regarding the name I do agree, but I thought adding one name will always make the list richer.

Thanks for your quick response.

PR #622, opened 2015-06-25
Improving pt_BR radio/phone number generation

igorsantos07:
Actually any radio number used as a cellphone should have the ninth digit as well.

### DIFF
--- a/src/Faker/Provider/pt_BR/PhoneNumber.php
+++ b/src/Faker/Provider/pt_BR/PhoneNumber.php
@@ -33,14 +33,13 @@ class PhoneNumber extends \Faker\Provider\PhoneNumber
      * Generates a 8/9-digit cellphone number without formatting characters.
      * @param bool $formatted [def: true] If it should return a formatted number or not.
      * @param bool $ninth     [def: false] If the number should have a nine in the beginning or not.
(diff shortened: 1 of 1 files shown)


src/Faker/Provider/pt_BR/PhoneNumber.php
```diff
@@ -33,14 +33,13 @@ public static function areaCode()
      * Generates a 8/9-digit cellphone number without formatting characters.
      * @param bool $formatted [def: true] If it should return a formatted number or not.
      * @param bool $ninth     [def: false] If the number should have a nine in the beginning or not.
-     *                        If the generated number begins with 7 this is ignored.
```
fzaninotto: Can you link to a public explanation of the rule behind that logic?

igorsantos07: It's explained in the docs for `$ninthDigitAreaCodes`. In short, Brazilian cellphones are receiving a ninth digit in a phased way, across our area codes.

fzaninotto:
needs rebase

igorsantos07:
Actually #623 seems like an improvement over this. I should have closed this PR, sorry!

PR #623, opened 2015-06-25
pt_BR: more flexibility to cellphone generation

igorsantos07:
Currently a boolean is used to define if the cellphone should
contain the ninth digit or not. This was extended so if an area
code needs to be enforced, it can be generated externally and given
to the cellphone generator, and it will create a correct number.

> This wasn't made in the same branch as #622 since it's changing a method's argument, and it could create a little controversy. But in the end the behaviour is still backwards compatible :)


fzaninotto:
I don't understand. A user passing an area code will get a leading 9. What's the difference with a boolean parameter?

igorsantos07:
If you ask for the phone number only the boolean is used to allow for the additional digit or not. If you pass in the area code, it will be used to verify if the ninth digit is needed.

fzaninotto:
Thanks!

PR #624, opened 2015-06-27
Fix some data for ukrainian generator

DesertCat:
Many thanks to the author for this wonderful library!

Using the library, I found some obsolete or incorrect data. Please find my contribution useful.


src/Faker/Provider/uk_UA/Address.php
```diff
@@ -130,7 +130,7 @@ class Address extends \Faker\Provider\Address
         'Чорногорія',
         'Марокко',
         'Мозамбік',
-        'М\'янма',
+        'М’янма',
```
fzaninotto: Do you mean that ' (the ascii quote) isn't at all used in Ukrainian, or is it a matter of typography?

src/Faker/Provider/uk_UA/PhoneNumber.php
```diff
@@ -8,22 +8,28 @@ class PhoneNumber extends \Faker\Provider\PhoneNumber
 
         // International format (mobile)
         '+38050#######',
-        '+38096#######',
-        '+38067#######',
-        '+38093#######',
         '+38063#######',
-        '+38099#######',
+        '+38066#######',
```
fzaninotto: I have no way to tell which is right, your version or the previous one. Please provide a link to a public page explaining the phone number formats for future reference.

DesertCat:
The ascii single quote used as apostrophe in ukrainian, but only if text has non-unicode encoding. Text editors, such as MS Word, automatically replace quote in appropriate words to U+2019.

In the matter of phone number generator, you can look [here](https://uk.wikipedia.org/wiki/Коди_мобільних_операторів_України) and [here](https://en.wikipedia.org/wiki/Telephone_numbers_in_Ukraine): last 15 lines are about mobile phone numbers in our country. And all numbers with national prefix started with 9 (+3809########) are mobile. Properly, would be worth to add generator for landline phones, but so many variants exists for landline numbers.

And previous version was little incorrect, because zero after +38 belongs to international phone prefix, not to area or operator code. List of area codes is [here](https://en.wikipedia.org/wiki/List_of_dialling_codes_in_Ukraine).

fzaninotto:
You need to add the references you mentioned in your comment in phpDoc blocks above the lists. Think about future developers who will wonder how that list was built, and if they can add another format of their own.

fzaninotto:
Abandoned PR, closing.

PR #627, opened 2015-07-02
Add missing address formatter properties.

tonynelson19:


fzaninotto:
These are locale-specific providers, they are not present in the default provider. The policy for IDE hints is to add only the names of the formatters that are present everywhere.

PR #632, opened 2015-07-12
Update Generator.php

theofidry:
Cast `$seed` instead of checking if `null`. This way it does not throw an error if we passed a stringified int value either.

### DIFF
--- a/src/Faker/Generator.php
+++ b/src/Faker/Generator.php
@@ -184,7 +184,7 @@ class Generator
         if ($seed === null) {
             mt_srand();
         } else {
-            mt_srand($seed);
+            mt_srand((int) $seed);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/Faker/Generator.php
```diff
@@ -184,7 +184,7 @@ public function seed($seed = null)
         if ($seed === null) {
             mt_srand();
         } else {
-            mt_srand($seed);
+            mt_srand(intval($seed));
```
GrahamCampbell: no trailing whitespace please

GrahamCampbell: Why not just use `(int) $seed`. It's exactly the same, but has better performance...

theofidry:
Thanks @GrahamCampbell, sorry for the trailing spaces :)

src/Faker/Generator.php
```diff
@@ -184,7 +184,7 @@ public function seed($seed = null)
         if ($seed === null) {
             mt_srand();
         } else {
-            mt_srand($seed);
+            mt_srand((int)$seed);
```
GrahamCampbell: Should be `(int) $seed`.

fzaninotto:
An `assertTrue()` after your statement will do.

theofidry:
An empty one?

fzaninotto:
Include a message explaining why you do this, something like:

```php
$this->assertTrue(true, 'seeding with a string doesn\'t throw an exception');
```

Unfortunately , that's all phpUnit offers us to test that (cf https://github.com/sebastianbergmann/phpunit-documentation/issues/171).

theofidry:
ping @fzaninotto fast merge :)

fzaninotto:
Thanks!

PR #634, opened 2015-07-13
Add hautelook/AliceBundle

theofidry:


fzaninotto:
What's the difference with the other bundle?

theofidry:
Historically [hautelook/AliceBundle][2] was developed first but h4cc was a little bit frustrated as there was some functionalities lacking and the bundle not very active. But now both are no longer very active which is a bit annoying.

As [nelmio/alice](https://github.com/nelmio/alice) released a new version 2.0, I'm currently discussing with [hautelook/AliceBundle][2] to upgrade the version and discussing with its maintainer to keep this bundle living.

But I'm not sure it's really the point: this part is a list of third-party libraries relying on this library. Shouldn't we let to the choice of which library should be used to the user?

[1]: https://github.com/h4cc/AliceFixturesBundle
[2]: https://github.com/hautelook/AliceBundle

fzaninotto:
You're right, but repeating the same description doesn't help. I'd prefer if you listed the two names, and give a single description, something like:

> [Foo](), [Bar](): Description

theofidry:
Changed. I regrouped them under a `Symfony2`. In the end even BazingaFakerBundle does pretty much the same as the two other, although does not rely on fixtures loader.

fzaninotto:
Thanks!

theofidry:
You're welcome :)

PR #635, opened 2015-07-13
Update PHPDoc

theofidry:
Update PHPDoc `Faker\Provider\DateTime::dateTimeBetween()` which accept `\DateTime` parameters as well as `string` ones.


fzaninotto:
Thanks!

PR #636, opened 2015-07-14
added syntax highlighting

royopa:


fzaninotto:
Looking at [your fork](https://github.com/royopa/Faker/tree/patch-1), I din't think it eases reading.

PR #642, opened 2015-07-17
Fix: Dev dependencies are installed by default

localheinz:
This PR

* [x] removes the `--dev` option when installing dependencies with Composer, as this is the default anyway


fzaninotto:
Thanks!

PR #645, opened 2015-07-20
Added format to dateTimeBetween

hittten:
A feature to create a date time between with a custom format. that can be very useful for set an exact hour with a date time random.

### DIFF
--- a/src/Faker/Provider/DateTime.php
+++ b/src/Faker/Provider/DateTime.php
@@ -99,11 +99,13 @@ class DateTime extends \Faker\Provider\Base
      * Accepts date strings that can be recognized by strtotime().
      *
      * @param string $startDate Defaults to 30 years ago
-     * @param string $endDate   Defaults to "now"
-     * @example DateTime('1999-02-02 11:42:52')
+     * @param string $endDate Defaults to "now"
+     * @param string $format Defaults to "ISO8601"
+     *
      * @return \DateTime
+     * @example DateTime('1999-02-02 11:42:52')
      */
-    public static function dateTimeBetween($startDate = '-30 years', $endDate = 'now')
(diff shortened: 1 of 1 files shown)


src/Faker/Provider/DateTime.php
```diff
@@ -113,11 +115,10 @@ public static function dateTimeBetween($startDate = '-30 years', $endDate = 'now
         }
 
         $timestamp = mt_rand($startTimestamp, $endTimestamp);
-
         $ts = new \DateTime('@' . $timestamp);
         $ts->setTimezone(new \DateTimeZone(date_default_timezone_get()));
 
-        return $ts;
+        return new \DateTime($ts->format($format));
```
fzaninotto: I don't understand. You return a DateTime object, not a formatted string. What's the purpose of a format in this case?

hittten: The purpose of format is the posibility to create a DateTime object based on a format. For example:

<dateTimeBetween('-4 days', 'now', 'Y-m-d')>
This will give me a date between 4 days ago and now but, always at 00:00:00 becase i dont especify the hour.

<dateTimeBetween('-4 days', 'now', 'Y-m-d 8:00')>
This will give me a date between 4 days ago and now but, always at 08:00:00


<dateTimeBetween('-4 days', 'now', 'Y-m-d H:00')>
This will give me a date between 4 days ago and now but, always at actual hour in 0 minutes and 0 seconds

<dateTimeBetween('-4 days', 'now')>
This will give me a date between 4 days ago and now but, always at random time

fzaninotto: I don't like it. You're using a format not to format, but to restrict the possible random dates.

PR #648, opened 2015-07-23
Fix random integer generation, fixes #646

### LINKED ISSUE #646: RandomNumber for integer/biginteger for CakePHP fails
When guessing the column type for CakePHP, `randomNumber` gets called with either `10` or `20` number of digits. This causes an error, because that's above the `mt_getrandmax()` on basically all systems.

The calls causing this issue:
https://github.com/fzaninotto/Faker/blob/master/src/Faker/ORM/CakePHP/ColumnTypeGuesser.php#L24-L31

Maybe just generate a lower random value for those fields? Usually real data won't go to the max of the range anyway.

jadb:


jadb:
ping @fzaninotto

fzaninotto:
Thanks!

PR #649, opened 2015-07-23
Adding SSN to en_US Person Provider

oh4real:
We are using a 3rd party service that validates the format of SSN, so randomNumber(9) failed their validation about 10% of the time - no SSN in the 900s at the moment.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0",
-        "squizlabs/php_codesniffer": "~1.5"
+        "squizlabs/php_codesniffer": "~2.0"
     },
     "suggest": {
          "ext-intl": "*"
--- a/src/Faker/Provider/en_US/Person.php
+++ b/src/Faker/Provider/en_US/Person.php
@@ -119,6 +119,7 @@ class Person extends \Faker\Provider\Person
 
     /**
      * @example '444114444' or '444-11-4444'
+     * @param boolean|null $dashed true, false or null
      */
     public function ssn($dashed = false)
     {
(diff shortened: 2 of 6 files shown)


composer.json
```diff
@@ -14,7 +14,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0",
-        "squizlabs/php_codesniffer": "~1.5"
+        "squizlabs/php_codesniffer": "~2.0"
```
fzaninotto: Why this change in this PR?

src/Faker/Provider/en_US/Person.php
```diff
@@ -116,4 +116,21 @@ public static function suffix()
...
     }
+
+    /**
+     * @example '444114444' or '444-11-4444'
+     * @param boolean|null $dashed true, false or null
+     */
+    public function ssn($dashed = false)
+    {
+        $ssn = static::randomNumber(9);
+        while (!preg_match('/(?!000)(?!666)(?!9)[0-9]{3}(?!00)[0-9]{2}(?!0000)[0-9]{4}/', $ssn)) {
```
fzaninotto: No, that's not the way to do it, it may take too many tries to get a correct ssn. Use a list of preformatted SSN patterns (like in the `phoneNumber` generator) instead.

oh4real: Howdy,

1/10 (the 9XX-XX-XXXXs) will fail and the extremely rare other cases. We run this routinely and it 90% runs once and 99% runs in two.

I'll look at pre-formatting.

Thanks for feedback!
oh4real

fzaninotto:
No news for 6 months, closing.

PR #653, opened 2015-07-28
Add RGBA functionality to Color Provider

apsylone:
I had to use this Faker data for a project. So I share this simple new functionality with the community :-)

### DIFF
--- a/src/Faker/Provider/Color.php
+++ b/src/Faker/Provider/Color.php
@@ -91,6 +91,14 @@ class Color extends Base
     }
 
     /**
+     * @example 'rgba(0,255,122,0.8)'
+     */
+    public static function rgbaCssColor()
+    {
+        return 'rgba(' . static::rgbColor() . ',' . static::randomFloat(1, 0, 1) . ')';
+    }
+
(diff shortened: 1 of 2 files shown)


src/Faker/Provider/Color.php
```diff
@@ -91,6 +91,14 @@ public static function rgbCssColor()
     }
 
     /**
+     * @example 'rgba(0,255,122,0.8)'
+     */
+    public static function rgbaCssColor()
```
fzaninotto: I'd name it `rgbaColor()`

fzaninotto:
Great! Would you mind adding a unit test, checking that the output satisfies a regular expression?

src/Faker/Provider/Color.php
apsylone: I tried to respect the naming convention of the function "rgbCssColor". That's why I think this is the most appropriate. Don't you think so ?

fzaninotto: right!

fzaninotto:
Thanks!

PR #663, opened 2015-08-08
Fix: Namespace is Faker\Test, not Faker\PHPUnit

localheinz:
This PR

* [x] fixes the namespace in `autoload-dev`

:information_desk_person: The namespace is `Faker\Test`, not `Faker\PHPUnit`.


fzaninotto:
Thanks

PR #666, opened 2015-08-10
updated regexify to ignore escaped parenthesis

kkiernan:
I wanted to do something like the following to generate a specific id format, but the parenthesis were missing in the result. This pull request fixes this issue.

`$faker->regexify('(Q|O)-\([0-9]{4}\)-[0-9A-Z]{6}')`

The above results in something like `Q-(2297)-ZNS2IJ` instead of `Q-2297-ZNS2IJ`

### DIFF
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -420,7 +420,7 @@ class Base
             return str_repeat($matches[1], Base::randomElement(range($matches[2], $matches[3])));
         }, $regex);
         // (this|that) becomes 'this' or 'that'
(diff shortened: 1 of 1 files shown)


fzaninotto:
`regexify` is already slow, I bet your change makes it even slower. I prefer not to support this edge case and avoid backreferences. 

Besides, your change misses a unit test.

kkiernan:
I don't think it's my change that misses the unit test. Running the tests a few times in a row results in anywhere from 3 - 5 missed tests even without my change.

I'd also argue that regexify is sort of meant to serve edge cases, otherwise you'd just use one of the many other provided methods.

PR #667, opened 2015-08-11
Modification to randomKey Method to support array_key searches

billmurrin:
While working on a Provider addition to Faker I came across a use case where I wanted to use the Faker method for randomKey to return only a subset of the keys which contain a particular value (e.g. category or type as a value). The php function array_keys function supports search value as an optional second argument and I saw that the randomKey method uses the array_keys function to return the array keys for randomization. 

This modification edits Base::randomKey method to accept an optional input for $search. In certain use cases, this will ensure that only a subset of the array keys are returned for randomization.

This is my first pull request. I hope I didn't screw anything up. If so, your constructive criticism is appreciated.

### DIFF
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -195,12 +195,12 @@ class Base
      * @param  array $array
      * @return int|string|null
      */
-    public static function randomKey($array = array())
+    public static function randomKey($array = array(), $search = null)
     {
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks, but it's not the purpose of Faker to provide all the PHP features. Faker is just here to give random data, not to help searching for data.

PR #670, opened 2015-08-13
Added missing @property docs to Generator

jessedc:
No codechanges, just docs.


pddevins:
@jessedc : Should `@property \DateTime $dateTimeInInterval` also be included near line 90-ish (makes most sense) to comply with the docs?

fzaninotto:
Duplicate of #627

PR #675, opened 2015-08-19
Adds time zone support for provider methods returning DateTime instance

bishopb:
There are four `DateTime` returning methods: `dateTime`, `dateTimeAD`, `dateTimeBetween`, and `dateTimeInterval`,  Two of these (`dateTimeBetween` and `dateTimeInterval`) set the timezone on the returned object, while the other two do not.  This can lead to user land juggling timezones when comparing dates returned from the different generators.

This PR updates these four `DateTime` returning methods to set the time zone to result of `date_default_timezone_get` or the  user-specified timezone given in the final argument to all methods.  Test coverage updated to check that the default and custom time zones are handled.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -180,13 +180,13 @@ Each of the generator properties (like `name`, `address`, and `lorem`) are calle
 ### `Faker\Provider\DateTime`
 
     unixTime($max = 'now')                // 58781813
-    dateTime($max = 'now')                // DateTime('2008-04-25 08:37:17')
-    dateTimeAD($max = 'now')              // DateTime('1800-04-29 20:38:49')
+    dateTime($max = 'now', $timezone = date_default_timezone_get()) // DateTime('2008-04-25 08:37:17', 'UTC')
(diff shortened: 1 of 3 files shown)


fzaninotto:
This is a BC break, because dates that were using `date_default_timezone_get()` now use `UTC` by default. Please consider using `date_default_timezone_get()` as the default, or I'll have to close the PR.

bishopb:
@fzaninotto Ah, yes. Fixed and tests passing.

fzaninotto:
Great! One thing is missing though: could you update the part of the Readme that documents the formatters you've modified to add the new parameter?

bishopb:
@fzaninotto Done, but looking over the code again, I think the other `DateTime` methods should also take `$timezone` to be consistent. Specifically `dateTimeThisCentury`, `dateTimeThisDecade`, `dateTimeThisYear`, and `dateTimeThisMonth`.

What do you think?

fzaninotto:
Thanks a lot!

PR #677, opened 2015-08-20
CakePHP ORM: Fix primaryKey warning

davidyell:
The return from the `primaryKey()` method can be both array and string, so this should be accounted for.

/cc @jadb


jadb:
:+1:

fzaninotto:
Thanks!

PR #680, opened 2015-08-24
move intl extension to require-dev in composer.json file

### LINKED ISSUE #584: Test not passing
Hi,
the file test/Faker/Provider/InternetTest.php has one test (testEmailIsValid, line 46) that is not passing to me, it try to assert that '.@yahoo.com' is a valid Email Address, I'm not a sysadmin expert but one email called '.@yahoo.com' exists or is the regex pattern in the test invalid ?

jaschweder:
fix #584

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,11 +14,9 @@
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0",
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks!

PR #681, opened 2015-08-25
Do not generate special purpose IPv4s

ravage84:
Do not genate 255.255.255.255 (Broadcast) and IPs between 0.0.0.0 and 0.255.255.255 (0.0.0.0/8, current network).

https://en.wikipedia.org/wiki/IPv4#Special-use_addresses

> 0.0.0.0/8 	Current network (only valid as source address) 	RFC 6890

https://en.wikipedia.org/wiki/List_of_assigned_/8_IPv4_address_blocks

And 255.255.255.255 (Broadcast) must not be used, also.

````
16777216 = 1.0.0.0 (new lowest generatable IPv4)
-2 = 255.255.255.254 (new highest generatable IPv4)
````

### DIFF
--- a/test/Faker/Provider/InternetTest.php
+++ b/test/Faker/Provider/InternetTest.php
@@ -110,7 +110,7 @@ class InternetTest extends \PHPUnit_Framework_TestCase
 
     public function testIpv4NotLocalNetwork()
     {
-        $this->assertNotEquals('1.0.0.0', long2ip(ip2long($this->faker->ipv4()) & (-1 << (32 - 8))));
+        $this->assertNotRegExp('/\A1\./', $this->faker->ipv4());
     }
 
     public function testIpv4NotBroadcast()
--- a/test/Faker/Provider/InternetTest.php
+++ b/test/Faker/Provider/InternetTest.php
@@ -108,6 +108,16 @@ class InternetTest extends \PHPUnit_Framework_TestCase
         $this->assertNotFalse(filter_var($this->faker->ipv4(), FILTER_VALIDATE_IP, FILTER_FLAG_IPV4));
     }
 
+    public function testIpv4NotLocalNetwork()
+    {
+        $this->assertNotEquals('1.0.0.0', long2ip(ip2long($this->faker->ipv4()) & (-1 << (32 - 8))));
+    }
+
+    public function testIpv4NotBroadcast()
+    {
+        $this->assertNotEquals('255.255.255.255', $this->faker->ipv4());
+    }
+
     public function testIpv6()
     {
         $this->assertNotFalse(filter_var($this->faker->ipv6(), FILTER_VALIDATE_IP, FILTER_FLAG_IPV6));
--- a/src/Faker/Provider/Internet.php
+++ b/src/Faker/Provider/Internet.php
@@ -288,7 +288,7 @@ class Internet extends \Faker\Provider\Base
      */
     public function ipv4()
     {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by steampilot** [comment]
How about the Loopback, shouldn't we exclude that too?
> 127.0.0.0/8 	Loopback 	RFC 6890

**Comment by fzaninotto** [comment]
Could you please add a unit test?

**Comment by ravage84** [comment]
@fzaninotto are those tests OK for you? Writing test cases for random data feels strange :smirk_cat: 

What do you think about @steampilot 's proposal about also exluding 127.0.0.0/8?

File: test/Faker/Provider/InternetTest.php [hunk 1]
```diff
@@ -108,6 +108,16 @@ public function testIpv4()
         $this->assertNotFalse(filter_var($this->faker->ipv4(), FILTER_VALIDATE_IP, FILTER_FLAG_IPV4));
     }
 
+    public function testIpv4NotLocalNetwork()
+    {
+        $this->assertNotEquals('1.0.0.0', long2ip(ip2long($this->faker->ipv4()) & (-1 << (32 - 8))));
```
fzaninotto: I don't understand your binary voodoo here.

fzaninotto:
Testing random data is OK - the CI isn't seeded so its runs with new random data every time.

I think your two cases are sufficient for exclusion.

test/Faker/Provider/InternetTest.php
ravage84: It shifts away the last three bits (B, C,& D from A.B.C.D), so only the first bit (A) remains.
This way we can compare it.
Probably there is a more efficient way, but I'm no binary calculation hero :smirk_cat:

TimWolla: `ip2long($this->faker->ipv4()) >= 0x01000000` will disallow IP addresses starting with `0x00` (i.e. starting with `0.`)

ravage84: @TimWolla you mean like this?
``$this->assertTrue(ip2long($this->faker->ipv4()) >= 0x01000000);``

fzaninotto: How about using a regex instead?

TimWolla: @ravage84 [`assertLessThanOrEqual(0x01000000, ip2long($this->faker->ipv4())`](https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertLessThanOrEqual).

ravage84: @TimWolla why less? :confused:
Shouldn't it be greater or equal?

ravage84: @fzaninotto I think @TimWolla 's proposal makes and is certainly the fastest of all.

TimWolla: > @TimWolla why less? :confused:

I inversed the order, as it is expected and then actual in PHPUnit.

fzaninotto: It is not maintainable. Even you guys don't agree on it. Please switch to a regex assertion.

ravage84: > It is not maintainable.

That's a bit exaggerated, but fair enough.
How about ``$this->assertNotRegExp('/\A1\./', $this->faker->ipv4());``
This checks if the IPv4 __starts__ with ``1.``, so ``10.`` does not fail.

fzaninotto: Yep, good for me

ravage84:
@fzaninotto done.

fzaninotto:
Awesome, thanks.

PR #683, opened 2015-08-26
Modify numberBetween() to be order agnostic

xfxf:
Hi, modified numberBetween() to be order agnostic (allow min/max to be supplied in either order).

I've done this as I found without this here, I had to put a *lot* of supporting conditional ordering code in my seeder script for a few specific usecases.  This makes using it much cleaner and won't break anything that currently uses it.

(And yes, am aware of randomFloat())

Thanks!

### DIFF
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -121,16 +121,18 @@ class Base
     }
 
     /**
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks

PR #689, opened 2015-08-27
Remove an invalid Dutch lastname that breaks email generator

SpaceK33z:
I changed the lastname `Jasper dr.` to `Jasper dr`. This lastname (with dot) does not occur anywhere in the Netherlands, plus it breaks the email generator. The email generator does not strip the last dot, so it can generate an email like this:  `Hassan.Jolijn@Jasperdr..net`.
This is, of course, an invalid email address according to PHP.


fzaninotto:
Thanks!

PR #694, opened 2015-09-01
Fix result example of text()

vlakoff:
Follow-up to #692.

I have chosen an example pointing out there may be several paragraphs. Though this occurs at a rate of about 2.5 %, I guess the newline might cause bad surprises if unexpected ;)


fzaninotto:
Thanks!

PR #696, opened 2015-09-02
Fix README’s improper definition

chanmix51:
Pomm is not a relational mapper.


fzaninotto:
That was a really bad mistake from me, I beg your pardon!

PR #699, opened 2015-09-08
Added (tk_TM) Turkmen lang support

isa424:


src/Faker/Provider/tk_TM/Person.php
```diff
@@ -0,0 +1,81 @@
...
+        '{{firstNameFemale}} {{lastName}}a',
+        '{{firstNameFemale}} {{lastName}}a',
+        '{{firstNameFemale}} {{lastName}}a',
+        '{{titleFemale}} {{firstNameFemale}} {{lastName}}a',
+    );
+
+    /**
+     * @var array Turkmen first names.
+     */
+    protected static $firstNameMale = array(
```
fzaninotto: This list, like the two following ones, is a little bit too small for Faker. Please find a larger dataset to insure a good variety in generated names.

Also, please mention the datasource in a phpDoc comment to ease future updates.

isa424: I could add a lot more names of course, but there is no source for names in my language to my knowlegde because I have added these from my head. Finding a source might be a problem, but I will keep adding more data.

fzaninotto: Please look for a source, because is the data only comes from your head, other devs might add what's in their own heads and we can't keep a stable list.

fzaninotto:
Please remove the Person provider if you can't link to a source.

fzaninotto:
Abandoned PR, closing.

PR #701, opened 2015-09-13
Support an asterisk in bothify and optimize

nineinchnick:
`Base::bothify()` now replaces an asterisk ('*') with either a random number or a random letter.

I also replaced `preg_replace_callback` in `numerify` and `lexify` as it is slower than iterating over a range of chars found by using `strpos` and `strrpos`.

Benchmarks are available here: https://gist.github.com/nineinchnick/84895a560910e3ead9cc

The biggest question is should we worry about UTF-8? I guess in a worst case it could replace the wrong byte. There are a few SO questions how to iterate over UTF-8 strings efficiently.

Should the new helper method `Base::replaceWildcard` be public?

### DIFF
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -317,6 +317,19 @@ class Base
         return join('', static::shuffleArray($array));
     }
 
+    private static function replaceWildcard($string, $wildcard = '#', $callback = 'static::randomDigit')
+    {
+        if (($pos = strpos($string, $wildcard)) === false) {
+            return $string;
+        }
+        for ($i = $pos, $last = strrpos($string, $wildcard, $pos) + 1; $i < $last; $i++) {
+            if ($string[$i] === $wildcard) {
+                $string[$i] = call_user_func($callback);
+            }
+        }
+        return $string;
+    }
+
     /**
      * Replaces all hash sign ('#') occurrences with a random number
      * Replaces all percentage sign ('%') occurrences with a not null number
@@ -329,9 +342,11 @@ class Base
         // instead of using randomDigit() several times, which is slow,
         // count the number of hashes and generate once a large number
         $toReplace = array();
-        for ($i = 0, $count = strlen($string); $i < $count; $i++) {
-            if ($string[$i] === '#') {
-                $toReplace []= $i;
+        if (($pos = strpos($string, '#')) !== false) {
+            for ($i = $pos, $last = strrpos($string, '#', $pos) + 1; $i < $last; $i++) {
+                if ($string[$i] === '#') {
+                    $toReplace[] = $i;
+                }
             }
         }
         if ($nbReplacements = count($toReplace)) {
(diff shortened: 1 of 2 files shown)


src/Faker/Provider/Base.php
```diff
@@ -360,17 +375,21 @@ public static function numerify($string = '###')
...
+     * Replaces hash signs ('#') and question marks ('?') with random numbers and letters
+     * An asterisk ('*') is replaced with either a random number or a random letter
      *
      * @param  string $string String that needs to bet parsed
      * @return string
      */
     public static function bothify($string = '## ??')
     {
+        $string = self::replaceWildcard($string, '*', function () {
+            return mt_rand(0, 1) ? '#' : '?';
```
fzaninotto: Why don't you directly pass `randomDigit()` or `randomAscii()` here? That's save one function call for each placeholder.

fzaninotto:
Hi, thanks for the patch! 

Being utf8-safe also explains why the current version may be slower. Although faster, your method drops an important ability (UTF-8), which is a stopper. I think `preg_replcae_callback()` with a `/u` is a must.

src/Faker/Provider/Base.php
nineinchnick: How does it save a function call? For `randomLetter()`, it will be called once anyway and `randomDigit()` is not called at all, there's an optimization already that gets many digits from a big number.

nineinchnick:
I've been thinking about it. We can still use `strpos` to extract a shorter string to pass to `preg_replace_callback()`. It also allows to skip the call completely in `bothify` if there are no '#' or '?'.

I'll benchmark that version and the one with custom utf-8 string iteration.

nineinchnick:
Btw there should be an unit test to see it fail for special utf-8 strings. Maybe I'll try it with [minimaxir/big-list-of-naughty-strings](https://github.com/minimaxir/big-list-of-naughty-strings).

nineinchnick:
There's an [answer on SO](http://stackoverflow.com/questions/15050710/php-strpos-substr-with-utf-8) that says it should be fine to search for an ascii char in an utf-8 string.

fzaninotto:
Then you'll need a test to prove it :)

nineinchnick:
Done and it seems to work. Done with https://gist.github.com/nineinchnick/917e644df42ccd62db5c
on every possible utf-8 char and those files:

* [minimaxir/big-list-of-naughty-strings](https://github.com/minimaxir/big-list-of-naughty-strings)
* [bits/UTF-8-Unicode-Test-Documents](https://github.com/bits/UTF-8-Unicode-Test-Documents)

fzaninotto:
I mean a unit test un Faker.

nineinchnick:
Done.

fzaninotto:
I meant testing Faker, not PHP. You have to make sure `numerify`, `lexify`, and `bothify` work correctly with UTF-8 strings.

nineinchnick:
I'm not sure how such a test would look like. Should I run it on that big utf-8 string, replace the single char back to the placeholder and compare?

fzaninotto:
No, just use a sample UTF-8 string with a couple special chars AND placeholders, and check that the result matches the expected with a regular expression (using `/u`)

nineinchnick:
Done.

fzaninotto:
Great, thanks!

PR #703, opened 2015-09-14
Improve en_AU/Address $postcode

xfxf:
Improves the $postcode options in the en_AU/Address.php Provider by restricting only to postcode that are valid.  Added because Faker was inserting invalid postcodes causing the application to not work correctly.

see https://en.wikipedia.org/wiki/Postcodes_in_Australia


fzaninotto:
Please add the reference to the wiki above the format list as a PHP comment, to help future maintainers understand this.

xfxf:
hi @fzaninotto , done

fzaninotto:
Thanks!

PR #708, opened 2015-09-21
added Japanese text for realText.

w1zardhead:
Included the novel called [Run,Melos! by Osamu Dazai](http://www.aozora.gr.jp/index.html) from Project Aozora Bunko to provide japanese real text.


fzaninotto:
Is it large enough to generate real randomness? Markov Chain generators usually require a larger text than what you included.

shunsuke-takahashi:
OK. So , let's provide a little more long text .

fzaninotto:
I squashed your two commits into one to avoid increasing the Faker Git checkout even more (see 3ad26cb1c4c9f66541d20ca4a64478a143cb992e). This has been merged to master.

PR #710, opened 2015-09-25
Enhancement: Add take() modifier, yielding a given number of results

localheinz:
This PR

* [x] implements a `TakeGenerator` and provides it via `Faker\Provider\Base`
* [x] updates `README.md`

:information_desk_person: Instead of doing this all the time

```php
$count = 5;

$randomNumbers = [];
for ($i = 0; $i < $count; $i++) {
    $randomNumbers[] = $faker->randomNumber;
}
```

you will like this:

```php
$count = 5;

$randomNumbers = $faker->take($count)->randomNumber;
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -31,7 +31,7 @@ Faker requires PHP >= 5.3.3.
 	- [Barcode](#fakerproviderbarcode)
 	- [Miscellaneous](#fakerprovidermiscellaneous)
 	- [Biased](#fakerproviderbiased)
-- [Unique and Optional modifiers](#unique-and-optional-modifiers)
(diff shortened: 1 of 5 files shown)


vlakoff:
I like the idea.

fzaninotto:
Sorry, Faker isn't here to complement PHP control structures. You'd better contribute to PHP core the necessary code to write:

```php
$randomNumbers = array_map([1...5], function() use ($faker) { return $faker->randomNumber })
```

Or even better:

```js
randomNumbers = [1...5].map(() => $faker->randomNumber)
```

PR #712, opened 2015-09-29
Image generation using GD

djlemmings:
Hello,

It's my first participation on Faker so excuse me in advance if I did some mistake.

I added a new formatter that creates images directly from PHP using the GD methods (usually by default on most setup).

I felt it could be a good addition as the image formatter currently uses the great LoremPixel, but has a drawback that it is quite slow on generating multiples pictures and need an internet connection.
This formatter doesn't require network and is way much faster, but... not as pretty indeed.

I added it on the Image Provider as it seemed obvious to me.

It takes about the same arguments as the image formatter (dir, width, height...) and I added possibility to add a text and change colors.

For the text I included the open source font Roboto (https://www.google.com/fonts/specimen/Roboto under Apache licence), I hope it's ok for you but I can understand it could be the main problem of my new feature.

I hope you'll like it and I'm ready to make any good changes suggested.

Best regards.


fzaninotto:
Hi, interesting generator!

However, I think generating random colors and text in a bitmap is far from generating a random image. the needs for random images can be to have random products, random profile pictures, etc. 

Including a new generator in Faker Core would only be possible if it covered such uses cases, but IMO it's an entirely different job (it means building up a library of tagged-open-source images and a resizer).

Anyway, I suggest you make it a standalone provider (such as https://github.com/fzaninotto/CompanyNameGenerator) and send me a PR to include a link to it in Faker's README.

Thanks anyway for the PR.

djlemmings:
Hello,

I'm a bit disapointed but I understand your point that makes sense.

I will convert this into a standalone provider, but if in any case I can change your mind, you say that the goal of faker is to generate "as much as possible" real content, that's perfectly true, but there are some cases where there isn't this need:

1) on a développement point of view, having real images might not be needed and it would greatly help when Faker is used on fixtures generation (as I told before, much faster than online fetching of data)
2) for automated tests, the same apply. An integration server doesn't care about the picture content.

Anyway, the new PR should be here soon.

Best regards.

nineinchnick:
How about passing a template image and only slightly modifying it?

djlemmings:
@nineinchnick I thought about that but it's a bit more complicated, and I'm not sure it would fit the "realistic" vision of Faker @fzaninotto told about.

But I'll think about something if I have time.

nineinchnick:
To keep it simple I'd suggest overlaying an [Identicon](https://en.wikipedia.org/wiki/Identicon) with partial opacity over the template. There could be an optional string provided to seed the Identicon.

This way, you could generate product images using the category image as template and product name/id as hash.

Actually, the Identicon generator is a good idea itself.

djlemmings:
@nineinchnick could be an idea!

PR #718, opened 2015-10-02
Fix call to missing method typo

sdustinh:
Fixes a call to a missing method in the CakePHP ORM provider.

```php
$this->populator->addEntity('Articles', 50, $formatter, $modifier);
$this->populator->execute();  // Fatal error
```


fzaninotto:
Thanks!

PR #719, opened 2015-10-03
minor update

pomaxa:
* shorten namespaces
* fix calling static methods as non static
* joins - implode change

### DIFF
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -263,7 +263,7 @@ class Base
      * @param array $array The set to shuffle
      * @return array The shuffled set
      */
-    public static function shuffleArray($array = array())
+    public static function shuffleArray(array $array = array())
     {
         $shuffledArray = array();
         $i = 0;
@@ -314,7 +314,7 @@ class Base
         } else {
             $array = str_split($string, 1);
         }
-        return join('', static::shuffleArray($array));
+        return implode('', static::shuffleArray($array));
     }
 
     private static function replaceWildcard($string, $wildcard = '#', $callback = 'static::randomDigit')
@@ -462,7 +462,7 @@ class Base
         // All A-F inside of [] become ABCDEF
         $regex = preg_replace_callback('/\[([^\]]+)\]/', function ($matches) {
             return '[' . preg_replace_callback('/(\w|\d)\-(\w|\d)/', function ($range) {
-                return join(range($range[1], $range[2]), '');
+                return implode(range($range[1], $range[2]), '');
             }, $matches[1]) . ']';
         }, $regex);
         // All [ABC] become B (or A or C)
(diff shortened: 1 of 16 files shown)


src/Faker/Provider/Base.php
```diff
@@ -263,7 +263,7 @@ public static function shuffle($arg = '')
      * @param array $array The set to shuffle
      * @return array The shuffled set
      */
-    public static function shuffleArray($array = array())
+    public static function shuffleArray(array $array = array())
```
GrahamCampbell: breaking change

pomaxa: not at all.

1) in php doc we got: "@param array $array The set to shuffle"
2) in unit tests there is no check for behavior of shuffleArray in case of passing non array there.
3) all usage of shuffleArray -> always pass array into
4) in case of someone tries to pass non array, 

PHP Warning:  reset() expects parameter 1 to be array, string given in php shell code on line 1

will be thrown, - cuz `reset()` is calling on $array var without checking it.

@GrahamCampbell  - please point me on where is BC, - cuz I do not see any.

GrahamCampbell: Changing the signature of a public or protected function on a non-final class is breaking for anyone extending the class.

pomaxa: @GrahamCampbell , but this method willcall en php error in case non array passed, + param type is written in phpdoc ?

pomaxa: I'll remove this change, if someone can explain why it is an BC; I understand that it's looks like it, but I don't see any situation, where it will be a problem. Cuz this function can't be used without an array as param.

vlakoff: You would be breaking this:

    class CustomBase extends Base {
        public static function shuffleArray($array = array()) {
            // ...
        }
    }

pomaxa:
```
class Base2 extends Faker\Provider\Base
{

    public static function shuffleArray($array = array()) {
        return $array;
    }
}

$generator = new \Faker\Generator();
$b = new Base2($generator);

var_dump($b->shuffleArray(array('a','b')));
```

and there is no error; what the problem ?

```
php -v                                                                                                            1:37  pmx@pmx-K56CB
PHP 5.6.4-4ubuntu6.2 (cli) (built: Jul  2 2015 15:29:28) 
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2014 Zend Technologies
    with Zend OPcache v7.0.4-dev, Copyright (c) 1999-2014, by Zend Technologies

```

GrahamCampbell: Very simply, there will be a fatal error...

GrahamCampbell: Ish. This is what actually happens: https://3v4l.org/m3RLg.

pomaxa: @GrahamCampbell thanks. I've forgot to enable errors; It really throw warning for this case. I have removed this changes.

fzaninotto:
Thanks!

PR #722, opened 2015-10-04
Fix pt br check digit

pomaxa:
remove include; change it for autooading simple staic class. + minor fixes.


src/Faker/Provider/pt_BR/LocalUtils.php
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Calculates one MOD 11 check digit based on customary Brazilian algorithms.
+     * @link http://en.wikipedia.org/wiki/Check_digit
+     * @link http://pt.wikipedia.org/wiki/CNPJ#Algoritmo_de_Valida.C3.A7.C3.A3o
+     * @link http://en.wikipedia.org/wiki/Cadastro_de_Pessoas_F%C3%ADsicas#Validation
+     *
+     * @param string $numbers Numbers on which generate the check digit
+     * @return integer
+     */
+    public static function checkDigit($numbers)
```
fzaninotto: Why don't you call it MOD11?

fzaninotto: Also, why create a class with a single static function instead of using a function like before?

fzaninotto:
Please rebase to avoid the coding style changes. Also, your method needs a unit test.

src/Faker/Provider/pt_BR/LocalUtils.php
pomaxa: @fzaninotto - I've kept name of function as it was before. If it's ok to change name of it, I'll do it too. Class was created just to remove require inside Providers and let it be done via autoloader

fzaninotto: Well, I don't mind using a ` require`  in that case.

fzaninotto:
Bump, please update your PR or I'll close it without merge.

fzaninotto:
Abandoned PR, closing.

Issue #711, opened 2015-09-29
Provider\Internet::domainWord() should produce lowercase results


vlakoff:
Using `$faker->domainName()`, I noticed the results were capitalized. `companyEmail()` is also affected, as it uses `domainName()`.

It is because the underlying `domainWord()` uses `Provider\Person::lastName()` (with localization and transliteration), but keeping the initial capital.

I think the expected behaviour is undoubtedly to have lowercase results.

vlakoff:
This happens only when intl extension missing. See code of [Provider\Internet::transliterate()](https://github.com/fzaninotto/Faker/blob/f23b5d6f4f728a4f00be859b4e3b02e12e896d7b/src/Faker/Provider/Internet.php#L147), and its [latest changes](https://github.com/fzaninotto/Faker/commits/master/src/Faker/Provider/Internet.php).

fzaninotto:
Capital letters are valid in domain names and emails - even if ignored by browsers and mail clients. I wonder what's best for Faker: generate expected output, or unexpected output? I think the latter...

vlakoff:
I think the contrary, as it is very uncommon to encounter capital letters.

If you want to cover more cases, you might consider:
* hyphens, which are allowed unless leading/trailing
* IDN

fzaninotto:
Fixed by #725

ravage84:
I just tripped over this. Good to see that it is already fixed.

@fzaninotto is there any chance we see a new release (bug fix/patch or minor, I don't care) which contains this fix any time soon?

The last release, v1.5.0, was released quite some time ago and people could make use the those 192 commits integrated since. I'd be happy to help, if I can.

fzaninotto:
@ravage84 not in 2015. If you want to help, well there is a lot to do:

- review pending PRs
- generate changelog
- write a blog post about the new stuff
- ????
- profit!

PR #723, opened 2015-10-04
Add missing lowercasing to the "no intl-ext" fallback

vlakoff:
Follow-up to #711.

* Looking at #333, it clearly seems to be an overlook.
* Makes results consistent with/without intl-ext.


fzaninotto:
Misleading. A method called `transillterate` should only translitterate. I guess it's the caller's responsibility to compose with a call to `lowercase`.

vlakoff:
Agree, I thought the same. Improved PR to come.

PR #724, opened 2015-10-04
Optimize Internet::toAscii() by using a static cache

vlakoff:
It makes quite a difference, even the `$transliterationTable` array creation is costly.
No possible drawback as it is contained inside the method.


vlakoff:
BTW, looking at code before #333, it seems this method shouldn't be public.

src/Faker/Provider/Internet.php
```diff
@@ -32,116 +32,122 @@ class Internet extends \Faker\Provider\Base
...
-            'Č' => 'C','ș' => 's','ț' => 't', 'ộ' => 'o','ắ' => 'a','ş' => 's',
-            "'" => '', 'ու' => 'u','ա' => 'a','բ' => 'b','գ' => 'g','դ' => 'd',
-            'ե' => 'e','զ' => 'z','է' => 'e','ը' => 'y','թ' => 't','ժ' => 'zh',
-            'ի' => 'i','լ' => 'l','խ' => 'kh','ծ' => 'ts','կ' => 'k','հ' => 'h',
-            'ձ' => 'dz','ղ' => 'gh','ճ' => 'ch','մ' => 'm','յ' => 'y','ն' => 'n',
-            'շ' => 'sh','ո' => 'o','չ' => 'ch','պ' => 'p','ջ' => 'j','ռ' => 'r',
-            'ս' => 's','վ' => 'v','տ' => 't','ր' => 'r','ց' => 'ts','փ' => 'p',
-            'ք' => 'q','և' => 'ev','օ' => 'o','ֆ' => 'f',
-        );
+        static $arrayFrom, $arrayTo;
```
GrahamCampbell: please don't define multiple variables on one line

GrahamCampbell: oh, wait, ignore me - i thought this was on a different repo...

fzaninotto:
Great, thanks!

vlakoff:
So, `toAscii()` was previously private methods in localized providers, and isn't meant to be used externally (or it shouldn't belong to `Internet` provider). Shall its visibility be changed to `protected`?

Also, `toAscii()` and `transliterate()` should be moved to bottom of the file to ease reading of the rest.

fzaninotto:
Agreed on both propositions.

PR #725, opened 2015-10-04
Move lowercasing out of transliterate()

vlakoff:
Also fixes the missing lowercasing with the "no intl-ext" fallback.

Follow-up to #333 and #723.


vlakoff:
Just to point out, localized providers don't need any update.

fzaninotto:
Thanks!

PR #728, opened 2015-10-05
StyleCI / PHP-CS-Fixer

Soullivaneuh:
I recommend you to use [StyleCI](https://styleci.io/) for CS sniffing instead of a rules on Travis.

This will be more reliable (see CS fix commit) and quicker (quasi-instant result on PR).

To enabled it, just go to the related website, login though GitHub, and activate your repository. I can update PR to trigger StyleCI event if you want to check that before.

I also provided a bridge to parse `.styleci.yml` config file and automatically configure PHP-CS-Fixer.

With that, contributors will just have to run `make cs` command to get CS fixed.

For the moment, I set `psr2` rules as you did on your Makefile, but I would suggest to use a higher CS preset like symfony one, and add some extra rules like `aligne_double_arrow` or `phpunit_strict`.

Please tell me if you want to add it, i'll update my PR.

Kind regards.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 vendor
 composer.lock
+.php_cs.cache
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,10 @@
+<?php
+
+// Needed to get styleci-bridge loaded
+require_once './vendor/autoload.php';
+
+use SLLH\StyleCIBridge\ConfigBridge;
+
+return ConfigBridge::create()
+    ->setUsingCache(true)
+;
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,0 +1,1 @@
+preset: psr2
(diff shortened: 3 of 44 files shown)


fzaninotto:
Hi Soullivaneuh, thanks for your PR but I'm fine with the current setup (`make sniff` integrated to the Travis build), and I don't want to enforce higher CS.

Soullivaneuh:
If I can insist, it seems that your `sniff` command does not works well.

With php-cs-fixer configured to psr-2, I found some CS issue as you can see on 048d9dd32a3a2393ec9c6a9e81f689b688ad900a.

And have a separate service for CS with lighten Travis process IMHO.

Don't you want to give a single try? :-)

fzaninotto:
Again, I don't mind one missing or extra space here or there. I'm trying to build a library that people use, not a library that people fear.

GrahamCampbell:
People shouldn't fear it because of cs rules. It's surely fine to patch errors after a PR is merged?

fzaninotto:
People, I get it, you like CS. I like CS too, up to a certain point. The current setup works, the code quality is satisfactory IMO. I won't add a tool to enforce stricter standards.

PR #729, opened 2015-10-05
Optimize userName and domainWord

vlakoff:
My apologies for the amount of PRs! (and there is one more to come after)

Something I forgot when moving from #723 to #725.
`transliterate()` result is surely ascii compatible[1]. And in any hypothetical worst-case scenario (like user messing low level), `mb_strtolower()` wouldn't fix anything anyways, but make upfront bug less noticeable.

The contract is clear: `transliterate()` result is ascii compatible. And with this commit we save 2 function calls each time, which is great.

[1] `return preg_replace('/[^A-Za-z0-9_.]/u', '', $transString);`


fzaninotto:
Thanks!

PR #730, opened 2015-10-06
Final changes on transliterate() and toAscii()

vlakoff:
Follow-up to discussion at bottom of #724.

* `transliterate()` from private to protected
* `toAscii()` from public to protected

Then move these to bottom of file, as they are protected, and `toAscii()` is huge.


fzaninotto:
Thanks!

PR #737, opened 2015-10-10
Force first digit in building number to be non-zero. Fixes #706

### LINKED ISSUE #706: $buildingNumber can be 0, 00, 000
Hi there, $buildingNumber can actually produce numbers like 0, 00, 000.  Is this actually valid anywhere in the world?

Happy to provide a patch, just verifying if this is intentional behaviour first.

jmauerhan:
I have updated the default address provider and the US address provider to start building number with a non-zero number, as per the bug #706


fzaninotto:
Thanks!

jmauerhan:
Thank you!

PR #738, opened 2015-10-16
added PHPDocs, removed unused variable

daniel-mueller:
Hi,

I added some PHPDocs and removed a unused variable.

greetings,
Daniel

### DIFF
--- a/src/Faker/DefaultGenerator.php
+++ b/src/Faker/DefaultGenerator.php
@@ -10,16 +10,26 @@ class DefaultGenerator
 {
     protected $default;
 
+    /**
+     * @param $default
+     */
     public function __construct($default = null)
     {
         $this->default = $default;
     }
 
+    /**
+     * @param $attribute
+     */
     public function __get($attribute)
     {
         return $this->default;
     }
 
+    /**
+     * @param $method
+     * @param $attributes
+     */
     public function __call($method, $attributes)
     {
         return $this->default;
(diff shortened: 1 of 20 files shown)


fzaninotto:
`@param` annotations without type and/or comments are useless IMO. Please amend your PR to add types for the phpDoc you added, or remove them.

daniel-mueller:
Thanks for feedback. I removed every @param where I couldn't safely determine the type. On the others I added a type.

src/Faker/DefaultGenerator.php
```diff
@@ -15,11 +15,18 @@ public function __construct($default = null)
         $this->default = $default;
     }
 
+    /**
+     * @param string $attribute
```
fzaninotto: nah, not necessarily `string`. Same for `__call`

src/Faker/Factory.php
```diff
@@ -25,6 +25,10 @@ public static function create($locale = self::DEFAULT_LOCALE)
         return $generator;
     }
 
+    /**
+     * @param string $locale
```
fzaninotto: why don't you document `$provider`, too? (string). Same for `findProviderClassname`

src/Faker/DefaultGenerator.php
fzaninotto: Sorry, forget that.

daniel-mueller:
I added the two annotations.

fzaninotto:
Thanks!

PR #740, opened 2015-10-16
[BUGFIX] Fix path to autoload.php in phpunit.xml

r3h6:


GrahamCampbell:
This actually looks wrong to me.

GrahamCampbell:
We want to be using composer's autoloader.

GrahamCampbell:
The one in src is only there if you want to use this library without composer.

r3h6:
Didn't thought about composer -> forget then this request.

PR #742, opened 2015-10-16
Update Address.php

nickwebcouk:
Minor change, added Scottish localisms (Loch as prefix, grove as suffixe)


fzaninotto:
Low value, so I'll refuse this PR. Remember that each change in a provider might break tests of developers using seeding. If we break those tests, it must be for a good reason.

PR #743, opened 2015-10-18
Add Lorem to Persian (fa_IR) language.

nimahejazi:
I've added Lorem.php to Persian (fa_IR) language.

### DIFF
--- a/src/Faker/Provider/fa_IR/Lorem.php
+++ b/src/Faker/Provider/fa_IR/Lorem.php
@@ -1,0 +1,181 @@
+<?php
+
+namespace Faker\Provider\fa_IR;
+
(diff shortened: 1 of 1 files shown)


fzaninotto:
Lorem is Locale invariable. Look for the Text provider for localized text.

nimahejazi:
I understand Lorem is invariable, but for Persian language which has different alphabet than Latin, it needs to be random words of Persian language, otherwise it would be useless.

mojtabaahn:
It's a long time since this was posted but I think it needs to re-open.
As @nimahejazi said English Lorem is completely different from Persian (characters, direction, typography..) and using it is pointless!

PR #744, opened 2015-10-18
Latest craze in TLD's in South Africa

sarelvdwalt:
Added org.za which is common, but more importantly the 3 primary city ones that are used lately.


fzaninotto:
Low value PR IMO.

Issue #748, opened 2015-10-21
Unique doesn't work with words.


berpcor:
```
\Faker\Factory::create()->unique()->words(14);
```

returns 

```ErrorException in UniqueGenerator.php 
line 50: array_key_exists(): The first argument should be either a string or an integer```

(Laravel)

Is it normal?

berpcor:
Damn...
I just need 250 unique words.

for ($i = 1; $i <= 250; $i++) {
      $iadefaults[] = $faker_iattributes_label->unique()->word();
}

gives me 

Maximum retries of 10000 reached without finding a unique value

How to generate unique words???

fzaninotto:
For your need of unique words, don't use the `Lorem` provider, but create a custom provider based on a dictionary (as in #754). This won't be merged to Faker core, though.

PR #749, opened 2015-10-21
Serialized `$res` key

EmanueleMinotto:
Closes https://github.com/fzaninotto/Faker/issues/748

Since `$res` could be anything, it could also be an array or an object, so it must be serialized to be used for key checking.


dlondero:
Agree with @EmanueleMinotto. Was about to open an issue for a problem with `dateTimeBetween` in conjunction with `UniqueGenerator`: in this case `$res` is a `DateTime` object which was ok for the previous implementation (see https://github.com/fzaninotto/Faker/pull/491) using `in_array` but definitely not ok for `array_key_exists`.

fzaninotto:
Thanks!

vlakoff:
I suppose `serialize($res)` should be moved out of the loop.

PR #752, opened 2015-10-24
Update Person.php

tytarenko:
Added middle name for women and corrected name format for woman


fzaninotto:
There are problems with your code, please see the CI report.

PR #754, opened 2015-10-28
Implemented dictionary for issue #748

daleattree:
User can enable a full dictionary load when needing more than 182 unique words as specified in issue #748


fzaninotto:
This won't be merged in Faker, for two reasons:

* `Lorem` returns fake Latin test by design. If you need more words, use another provider.
* For localized text, the provider is `Text`. It'd make more sense to grab a word from these already existing providers, rather than require a new dictionary for each locale.

Nothing prevents you from bundling this as a standalone formatter, though. I'd be happy to reference it in the README.

daleattree:
Fair enough. Thanks.

PR #755, opened 2015-10-28
Added Swedish surname Sjösten

tobiassjosten:
What would a name library be without it? :wink:


fzaninotto:
Low value. Each change in the fake data breaks tests of people relying on seeding. We should only modify data for good reasons, this one isn't one.

PR #757, opened 2015-10-29
Change middleName method, added female middle names

tytarenko:


fzaninotto:
Same remarks as for #752.

fzaninotto:
CS check fails, no datasource, no feedback for some months: closing.

PR #759, opened 2015-10-30
* Added an en_ZA company provider, that can generate company registration numbers

smithandre:
* Added an en_ZA company provider, that can generate company registration numbers


fzaninotto:
Thanks!

PR #761, opened 2015-10-31
Updated TLD list for South Africa locale

smithandre:
Updated TLD list for South Africa locale and added Unit Test for Internet Provider


fzaninotto:
Thanks!

PR #763, opened 2015-11-03
add DNI generator to spanis person

mikk150:
I needed to generate DNI on unit tests and I use faker as data provider. I couldn't find DNI anywhere, so I implemented it

### DIFF
--- a/src/Faker/Provider/es_ES/Person.php
+++ b/src/Faker/Provider/es_ES/Person.php
@@ -4,6 +4,8 @@ namespace Faker\Provider\es_ES;
 
 class Person extends \Faker\Provider\Person
 {
+    private static $letters=array('T', 'R', 'W', 'A', 'G', 'M', 'Y', 'F', 'P', 'D', 'X', 'B', 'N', 'J', 'Z', 'S', 'Q', 'V', 'H', 'L', 'C', 'K', 'E', 'T');
+
     protected static $maleNameFormats = array(
         '{{firstNameMale}} {{lastName}}',
         '{{firstNameMale}} {{lastName}}',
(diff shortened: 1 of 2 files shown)


src/Faker/Provider/es_ES/Person.php
```diff
@@ -68,4 +70,12 @@ public static function suffix()
     {
         return static::randomElement(static::$suffix);
     }
+
+    public static function dni()
```
fzaninotto: please add a link to the dni specification

src/Faker/Provider/es_ES/Person.php
```diff
@@ -4,6 +4,8 @@
 
 class Person extends \Faker\Provider\Person
 {
+    private static $letters=array('T', 'R', 'W', 'A', 'G', 'M', 'Y', 'F', 'P', 'D', 'X', 'B', 'N', 'J', 'Z', 'S', 'Q', 'V', 'H', 'L', 'C', 'K', 'E', 'T');
```
fzaninotto: `letters` is a too specific name. Please rename it to make it obvious that this relates to DNI validation

fzaninotto:
You should add a unit test for this new formatter.

src/Faker/Provider/es_ES/Person.php
mikk150: https://es.wikibooks.org/wiki/Algoritmo_para_obtener_la_letra_del_NIF#Algoritmo

src/Faker/Provider/es_ES/Person.php
mikk150: maybe crcMap?

mikk150: @fzaninotto is it better now?

src/Faker/Provider/es_ES/Person.php
fzaninotto: Please add it as a `@link` annotation in the method phpDoc

PR #767, opened 2015-11-06
Added en_AU unit test folder

Newman101:
Added en_AU unit test folder. 

Thank you.

PR #769, opened 2015-11-09
en_CA unit test folder added

Newman101:
Unit test support added to en_CA Provider, with PhoneNumberTest.php test kit. 

Thank you.

### DIFF
--- a/test/Faker/Provider/en_CA/PhoneNumberTest.php
+++ b/test/Faker/Provider/en_CA/PhoneNumberTest.php
@@ -29,7 +29,7 @@ class PhoneNumberTest extends \PHPUnit_Framework_TestCase
     public function testPhoneNumberFormat()
     {
       $number = $this->faker->phoneNumber;
-      $this->assertRegExp('/^(\+?1)?([ -.]*\d{3}[ -.]*| *\(\d{3}\) *)\d{3}[-.]?\d{4}$/', $number);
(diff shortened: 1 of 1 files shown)


Newman101:
Looks like the RegEx does not pass test cases without international calling code. I'll need to modify the RegEx a bit to make sure it includes both numbers with and without country dial in codes.

fzaninotto:
Thanks!

fzaninotto:
Tests fail, I'll revert this one.

```
There was 1 failure:

1) Faker\Test\Provider\en_CA\PhoneNumberTest::testPhoneNumberFormat
Failed asserting that '529-411-8627 x175' matches PCRE pattern "/^((([0-9]{1})*[- .(]*([0-9]{3})[- .)]*[0-9]{3}[- .]*[0-9]{4})+)*$/".

/Users/francois/git/Faker/test/Faker/Provider/en_CA/PhoneNumberTest.php:32
```

Newman101:
Interesting, when I did run the tests it actually passed. I'll need to re-check this one. Thanks for notification.

fzaninotto:
I also partially reverted #767 for the same reason (see 0456c4efc8faffe9359cc75b6896bcc92d458ef6)

Newman101:
OK. I need to re-check these commits. As I mentioned, the build passed while committing.

PR #775, opened 2015-11-13
Added unit test folder for en_UG provider and AddressTest.php

Newman101:
Added unit test folder for en_UG provider and AddressTest.php. 

Thank you.


fzaninotto:
Thanks

Issue #527, opened 2015-02-28
Text Provider ensure text starts with an uppercase letter


Antonimo:
Many Languages don't have capital/uppercase letters (like hebrew).
Also I think people that use Faker to fill random data don't really care about that.
I suggest make it optional, so we could do something like this:

```php

class HebrewTextProvider extends \Faker\Provider\Text
{
    protected $ensureTextStartsWithUppercase = false;

```

fzaninotto:
Great idea. Would you like to work on a PR for that?

Antonimo:
just add an if!

Antonimo:
https://github.com/fzaninotto/Faker/pull/528

fzaninotto:
Fixed by #778

PR #778, opened 2015-11-17
Text Provider ensure text starts with an uppercase letter #527

LagunaJavier:
hi @fzaninotto !

I decided to push the changes requested in the issue https://github.com/fzaninotto/Faker/issues/527.
The code guidelines should be correct and tests passes.

Let me know if that is ok or should I update something.

best,
Javier


fzaninotto:
Thanks!

PR #779, opened 2015-11-18
Fixed a lot of semicolon missings

aniruddha-chakraborty:
There was a lot of ";" semicolon missings in readme.md file . I fixed those


vlakoff:
Semicolons aren't required when there is only one statement inside the PHP tags. Personally I prefer to omit them. Furthermore, I'd remove the semicolons from the `<?php endif; ?>` for lightness and consistency. But it's no big deal.

However, the last two lines of this PR are appropriate for sure.

fzaninotto:
I agree with @vlakoff, only the last two lines could get merged. Please update your PR accordingly.

PR #781, opened 2015-11-21
Add Swiss names

z38:
Added first & last names for each Swiss region (`de_CH`, `fr_CH` and `it_CH`).


ravage84:
:+1: His sources are legit.

vlakoff:
If I were expecting a child I'd use Faker to determine their name.
The first generated result, no negotiation.

fzaninotto:
Thanks!

And @vlakoff if that actually happens, please notify me, I'll be very happy!

PR #782, opened 2015-11-21
Improve Swiss phone numbers

z38:
Fixed the wrong international prefix and added valid mobile number formats.


fzaninotto:
Thanks

PR #783, opened 2015-11-26
ne_NP Added phone numbers & fixed provider country to Nepal only

kbirmhrjn:


src/Faker/Provider/ne_NP/Address.php
```diff
@@ -62,33 +62,7 @@ class Address extends \Faker\Provider\Address
         'Yala', 'Yalbang', 'Yamphodin', 'Yrsa',
     );
 
-    protected static $country = array(
```
fzaninotto: Why? This formatter is used to generate a random country name. A single country isn't random.

fzaninotto:
The CI complains about coding standards, check the report.

src/Faker/Provider/ne_NP/Address.php
kbirmhrjn: Just to let you know because, 
https://github.com/fzaninotto/Faker/blob/master/src/Faker/Provider/en_NZ/Address.php#LC53
has a single country.

kbirmhrjn:
I hope its good for merging into the master now.

src/Faker/Provider/ne_NP/Address.php
fzaninotto: That's a mistake in `en_NZ`.

fzaninotto: It's now fixed in `en_NZ`

kbirmhrjn:
Okay, its been fixed. can you look at it now again.

PR #785, opened 2015-11-28
Added provider for en_IN (India)

kartiksomani:
Added provider for Indian data. Contains Address, Person, Phone number and Internet data.

### DIFF
--- a/src/Faker/Provider/en_IN/Address.php
+++ b/src/Faker/Provider/en_IN/Address.php
@@ -1,0 +1,134 @@
+<?php
+
+namespace Faker\Provider\en_IN;
+
+class Address extends \Faker\Provider\Address
+{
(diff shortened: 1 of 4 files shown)


src/Faker/Provider/en_IN/Address.php
```diff
@@ -0,0 +1,134 @@
+<?php
+
+namespace Faker\Provider\en_IN;
+
+class Address extends \Faker\Provider\Address
+{
+
+    protected static $city = array(
```
fzaninotto: Could you provide a source URL in a phpDoc block?

fzaninotto: Nevermind, I'll go with your PR as is. Thanks!

PR #786, opened 2015-11-28
Code adjustments in unique() generator

vlakoff:
Highlights:

* When unique value found, serialize once instead of twice. If unique values are likely to be found on first try, this results in twice less serialize calls. Refs #749.
* Use `isset()` instead of `array_key_exists()`, the former being a language construct and not a function, it is significantly faster. This change also allows code simplification by removing the "intermediary index creation" step at beginning of method. Refs #491.

### DIFF
--- a/src/Faker/UniqueGenerator.php
+++ b/src/Faker/UniqueGenerator.php
@@ -37,9 +37,6 @@ class UniqueGenerator
      */
     public function __call($name, $arguments)
     {
-        if (!isset($this->uniques[$name])) {
-            $this->uniques[$name] = array();
-        }
         $i = 0;
         do {
             $i++;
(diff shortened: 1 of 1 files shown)


fzaninotto:
Did you benchmark the speedup?

vlakoff:
Just did some quick xdebug profiling, and the speedup about `serialize()` and `array_key_exists()` appears to be rather insignificant… (still, it can't harm, let's call this a little code polishing ;)

Though, the `call_user_func_array()` calls are *very* costly, they are a bottleneck (roughly 10% execution time of my test script, top 5 of most costly functions). They can be avoided [as Laravel does](https://github.com/laravel/framework/blob/8a2a2d5d9102ecab0cd4a69ffc073f09f03760e1/src/Illuminate/Support/Facades/Facade.php#L211), and I think the added bit of complexity is worth it.

After this PR is merged (if you wish so ;) I'll submit an other one to avoid calling this hugely costly `call_user_func_array()`.

fzaninotto:
If you submit a perf patch, you must be able to prove that it works. So please provide an execution time before/after on a series of calls, describing the conditions of the test. If the gain is significant, I'll merge. Otherwise, I won't. It's as simple as that.

vlakoff:
Please consider this PR as just some code adjustments, merge it or not, your pick :)

If I submit a PR about `call_user_func_array()`, I'll provide some numbers along.

vlakoff:
Rebased and simplified a little bit.

Despite the diffs, the changes are really straightforward. It just brings a cleaner code.

Could you please merge or close? your pick :)

PR #788, opened 2015-11-29
Added some fixes to Armenian locale

mhamlet:


src/Faker/Provider/hy_AM/PhoneNumber.php
```diff
@@ -4,34 +4,30 @@
...
+        '+374 {code} {numberFormat}',
     );
+
+    public function phoneNumber()
+    {
+        $format = static::randomElement(static::$formats);
+        $code = static::randomElement(static::$codes);
+        $numberFormat = static::randomElement(static::$numberFormats);
+
+        $number = str_replace(array('{code}', '{numberFormat}'), array($code, $numberFormat), $format);
```
fzaninotto: You should not need to `str_replace`, `parse()` would do it for you if there were related formatters.

fzaninotto:
Thanks!

PR #791, opened 2015-12-10
Wraps Faker\Provider\fa_IR\Text

yaim:
Wraps Text for fa_IR Provider to 72 char length lines for a better code reading.


fzaninotto:
That's not code that you wrapped - that's text. Also, your code editor probably has a wrap option.

PR #792, opened 2015-12-14
added state property

mseshachalam:


fzaninotto:
no, `$state` isn't present in all locales.

PR #793, opened 2015-12-17
Create Text.php

yzm629:


EmanueleMinotto:
Probably needs tests (under `test/Faker/Provider/zh_CN/TextTest`).

fzaninotto:
I think it doesn't make sense to use a translated piece of litterature (Alice in Wonderland) as basis for a text provider. Please find a source originally written in modern Chinese instead.

PR #795, opened 2015-12-21
Added the car feature

prio101:
Added a car feature . I mean random car names and codes.


fzaninotto:
Interesting, but a bit too specific for Faker Core. I suggest you create a third-party provider and host it - I'll be happy to reference it in Faker's README.

PR #796, opened 2015-12-23
Add Spot ORM

urisavka:
Adding [Spot2](https://github.com/vlucas/spot2) ORM support.

### DIFF
--- a/src/Faker/ORM/Spot/ColumnTypeGuesser.php
+++ b/src/Faker/ORM/Spot/ColumnTypeGuesser.php
@@ -1,0 +1,77 @@
+<?php
+
+namespace Faker\ORM\Spot;
+
+use Faker\Generator;
+
+class ColumnTypeGuesser
+{
+    protected $generator;
+
+
+    /**
+     * ColumnTypeGuesser constructor.
+     * @param Generator $generator
+     */
+    public function __construct(Generator $generator)
+    {
+        $this->generator = $generator;
+    }
+
+    /**
+     * @param array $field
+     * @return \Closure|null
+     */
+    public function guessFormat(array $field)
+    {
+        $generator = $this->generator;
+        $type = $field['type'];
+        switch ($type) {
+            case 'boolean':
+                return function () use ($generator) {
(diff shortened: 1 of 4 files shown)


fzaninotto:
Thanks!

src/Faker/ORM/Spot/ColumnTypeGuesser.php
```diff
@@ -0,0 +1,77 @@
...
+                return function () use ($generator, $size) {
+                    return $generator->randomNumber($size + 2) / 100;
+                };
+            case 'smallint':
+                return function () {
+                    return mt_rand(0, 65535);
+                };
+            case 'integer':
+                return function () {
+                    return mt_rand(0, intval('2147483647'));
```
TimWolla: Why don't you use `$generator->numberBetween()` here?

src/Faker/ORM/Spot/ColumnTypeGuesser.php
```diff
@@ -0,0 +1,77 @@
...
+                };
+            case 'bigint':
+                return function () {
+                    return mt_rand(0, intval('18446744073709551615'));
+                };
+            case 'float':
+                return function () {
+                    return mt_rand(0, intval('4294967295'))/mt_rand(1, intval('4294967295'));
```
TimWolla: `$generator->randomFloat()`. In fact I am not entirely sure whether your current method generates a uniform distribution (I suspect not).

src/Faker/ORM/Spot/ColumnTypeGuesser.php
urisavka: Copied it from Doctrine column type guesser. Will fix both today. 

I think ``$generator->randomNumber(10);`` is even more correct solution here...

src/Faker/ORM/Spot/ColumnTypeGuesser.php
urisavka: Same as above. Thanks for noticing!

PR #799, opened 2015-12-29
Fix incorrect mb_substr arguments

BePsvPT:


src/Faker/Provider/zh_TW/Text.php
```diff
@@ -109,7 +109,7 @@ protected static function appendEnd($text)
     {
         // extract the last char of $text
         if (function_exists('mb_substr')) {
-            $last = mb_substr($text, mb_strlen($text)-1, 'UTF-8');
+            $last = mb_substr($text, mb_strlen($text)-1, NULL, 'UTF-8');
```
EmanueleMinotto: Contributions should follow PSR2, so the third argument must be in [lower case](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#25-keywords-and-truefalsenull).

fzaninotto:
Thanks!

PR #800, opened 2015-12-30
Init new Geo provider for geographical and spatial data. Move longitu…

Mevrael:
Init new Geo provider for geographical and spatial data. Move longitude() and latitude() from Address to Geo.


.gitignore
```diff
@@ -1,2 +1,3 @@
+.idea
```
EmanueleMinotto: This should be in your global .gitignore instead.

src/Faker/Provider/Geo.php
```diff
@@ -0,0 +1,57 @@
...
+        return [static::longitude(), static::latitude()];
+    }
+
+    /**
+     * @example 'POINT(77.147489, 86.211205)'
+     * @return string An SQL value command used to insert longitude and latitude into POINT column.
+     */
+    public static function sqlGeoPoint()
+    {
+        return 'POINT(' . static::longitude() .', ' . static::latitude() .')';
```
EmanueleMinotto: imho this isn't a responsability of this library

Mevrael:
If not for seeding DB, generating random values then what is a responsibility of this library?

P.S. this is only the beginning, there would be new generators associated with spatial data. I made this because currently there are no Geo provider at all. Talking about long/lat, their location in address/translations is not the best place also. Now there is primitive Geo provider.

There is similar repository with geo provider https://github.com/bobthecow/Faker where you can see more generators but there are still much more functionality exists in GIS.

src/Faker/Provider/Geo.php
fzaninotto: I agree

Mevrael: @fzaninotto could you comment on this https://github.com/fzaninotto/Faker/pull/800#issuecomment-168022382 and give me a clear vision of what is "reposnibility of this library"?

fzaninotto:
Faker takes care of generating raw data. The actual formatting of the data is up to the user. Whether it's XML, JSON, SQL, or whatever format out there, it' not Faker's responsibility.

If you want to make a SQL formatter for all of Faker's data, be my guest. This will just have to live outside of the Faker Core.

PR #801, opened 2015-12-30
Update Image.php to allow generation of gray images using lorempixel

neutralrockets:
Add ability to generate gray images by providing "true" as the final parameter to imageUrl().


fzaninotto:
Thanks!

PR #804, opened 2016-01-05
Added formatters "state" and "stateAbbr" support in Simplified Chinese(zh_CN).

ezburg:
Also formatted the random data.


fzaninotto:
Thanks!

PR #805, opened 2016-01-06
Greek Text

hootlex:
Created Text.php for el_GR provider.


EmanueleMinotto:
I'm not sure if the `@licence` annotation could be used there (this library is under MIT).

hootlex:
I saw this annotation here: https://github.com/fzaninotto/Faker/blob/master/src/Faker/Provider/ru_RU/Text.php

fzaninotto:
Thanks!

PR #809, opened 2016-01-09
Update zh_TW text provider

BePsvPT:
1. Update coding style
2. Refactor text appendEnd method
3. Add text test


fzaninotto:
Thanks!

PR #812, opened 2016-01-21
Update Person.php

Papisou:
Hi 
I just added two names Florent and Valentin i think they are pretty frequent
:)


fzaninotto:
Low PR value, sorry. Faker doesn't need to have the names of all the people who star the repo on GitHub. Just a reasonable amount to guarantee that the data it generates looks legit.

PR #814, opened 2016-01-26
Update imageUrl method documentation

jonwurtzler:
Adding the $word property to method documentation to match actual method implementation.


fzaninotto:
thanks!

PR #818, opened 2016-02-02
Improved mod97 method of Iban class

Newman101:
Good morning.

I moved variable ```$size``` above for loop. The reason for this is that in the previous form the method calculates ```$size``` variable on every step of the loop. I changed this to be pre-calculated, which improves the efficiency.

Thank you.


okj579:
Actually it doesn't. If you look closely, you'll see that they're functionally equivalent

PR #820, opened 2016-02-02
Use generator to populate values for numeric fields for Spot provider.

urisavka:
I can do the same changes for Doctrine, BUT it will change values for existing seeds (randomFloat has different functionality than it's there now).

Given that Spot provider is new, I think we can afford changes here. 

So, what do you think about Doctrine?


fzaninotto:
I'll leave the Doctrine builder to the Doctrine maintainer.

PR #822, opened 2016-02-07
Latvia phone number format

veisis:
Phone numbers length for Latvia is 8 digits plus country code. And sometimes divided with spaces.


fzaninotto:
I don't see much difference with the previous format. Can you point me to a reference?

veisis:
https://en.wikipedia.org/wiki/Telephone_numbers_in_Latvia
And rel life experience :)
I am from Latvia.

fzaninotto:
Thanks

Issue #361, opened 2014-07-01
Improve german first names


mikehaertl:
As a native I have to say, that the german first names don't look very german to me :).

Here's a list of the most frequently used first names: 
- http://de.wiktionary.org/wiki/Verzeichnis:Deutsch/Liste_der_h%C3%A4ufigsten_m%C3%A4nnlichen_Vornamen_Deutschlands
- http://de.wiktionary.org/wiki/Verzeichnis:Deutsch/Liste_der_h%C3%A4ufigsten_weiblichen_Vornamen_Deutschlands

Findus23:
I known I'm a bit late, but would it be possible to improve the last names too?
I can't really believe that these are the most common last names. Especially names like `auch Schlauchin` don't make sense.
@mikehaertl Would it be possible to parse the names from https://de.wiktionary.org/wiki/Verzeichnis:Deutsch/Liste_der_h%C3%A4ufigsten_Nachnamen_Deutschlands ?
And while we're on it, is it possible to also update the Austrian Names from https://de.wiktionary.org/wiki/Verzeichnis:Deutsch/Liste_der_h%C3%A4ufigsten_Nachnamen_%C3%96sterreichs (or at least copy them from Germany)

mikehaertl:
@Findus23 It was mainly lack of time - and that I didn't need faker anymore so far. So feel free to send any updates you like. Also consider to update the first names like I've mentioned above. Most of them are not really sounding very german (yes, we have "multikulti" ;) -  but still I think the "classical" german names are most common).

Findus23:
I'll see if I am able to update them.
Do you still have the script to parse the list and create the array?

mikehaertl:
No, sorry. I think I've just copy & pasted this into Vim and then used some regex to extract the names.

Findus23:
I have made the pull requests #825 and #826.

PR #825, opened 2016-02-11
improved german names

Findus23:
I took the Top 500 last names from https://de.wiktionary.org/wiki/Verzeichnis:Deutsch/Liste_der_h%C3%A4ufigsten_Nachnamen_Deutschlands
and kept only the Top 500 first names.

-> #361


fzaninotto:
Thanks!

PR #826, opened 2016-02-11
improved Austrian names, states and realtext

Findus23:
- added states
- Top 500 last names from https://de.wiktionary.org/wiki/Verzeichnis:Deutsch/Liste_der_h%C3%A4ufigsten_Nachnamen_Deutschlands

- As there is no source of Austrian first names, I took the 30 most popular names in 1985, 1995, 2005 and 2015 from the Austrian statistical office to get a representative list of typical Austrian names.
http://www.statistik.at/wcm/idc/idcplg?IdcService=GET_PDF_FILE&RevisionSelectionMethod=LatestReleased&dDocName=021130


src/Faker/Provider/de_AT/Person.php
```diff
@@ -24,81 +24,79 @@ class Person extends \Faker\Provider\Person
...
+    'Isabella',
+    'Jacqueline', 'Jana', 'Jasmin', 'Jennifer', 'Johanna', 'Julia',
+    'Karin', 'Katharina', 'Katrin', 'Kerstin',
+    'Lara', 'Laura', 'Lea', 'Lena', 'Leonie', 'Lisa', 'Luisa',
+    'Magdalena', 'Manuela', 'Maria', 'Marie', 'Marlene', 'Martina', 'Melanie', 'Mia', 'Michaela', 'Michelle', 'Miriam',
+    'Nadine', 'Nicole', 'Nina',
+    'Petra',
+    'Sabine', 'Sabrina', 'Sandra', 'Sarah', 'Selina', 'Sonja', 'Sophia', 'Sophie', 'Stefanie',
+    'Tanja', 'Theresa',
+    'Valentina', 'Vanessa', 'Verena', 'Viktoria',
```
EmanueleMinotto: missing indentation

Findus23: Thanks, I missed that.

fzaninotto:
Interesting, but there aren't enough first names for the AT locale. Can you add more?

Findus23:
I have now added the whole list (Top 60). Now it contains the names of 60 to 80% of the population. (Apparently we aren't that creative when it comes to names)

fzaninotto:
I'm really surprised. Every other locale has several hundreds of names. Perhaps you need a better data source? 

Also, could you rebase now that the other PR is merged?

Findus23:
Unfortunately there is no official list of names like in Switzerland and it's hard to use baby names like in Great Britain because they are only published as an (ugly) PDF: http://www.statistik.at/wcm/idc/idcplg?IdcService=GET_PDF_FILE&RevisionSelectionMethod=LatestReleased&dDocName=057950 ([Overview](https://translate.google.at/translate?sl=de&tl=en&js=y&prev=_t&hl=de&ie=UTF-8&u=http%3A%2F%2Fwww.statistik.at%2Fweb_de%2Fstatistiken%2Fmenschen_und_gesellschaft%2Fbevoelkerung%2Fgeborene%2Fhaeufigste_vornamen%2Findex.html&edit-text=&act=url))

Most (European) countries I checked have about the same number of first names. (e.g France https://github.com/fzaninotto/Faker/blob/master/src/Faker/Provider/fr_FR/Person.php)

fzaninotto:
OK, fair enough.

Findus23:
Thanks for merging.

PR #827, opened 2016-02-11
Added Dutch text

rolfdenhartog:
I hope everybody will enjoy the Dutch text :)


fzaninotto:
Please use a native Dutch text, not a translation. The purpose of the realText provider is to generate really native looking text, it won't be the case with Alice in Wonderland.

PR #834, opened 2016-02-22
Add nl_NL base text

endroid:


fzaninotto:
Please provide a native Dutch text, not a translation from another language.

Issue #567, opened 2015-05-03
Length modifier for generators


hendore:
More of a feature request than an issue. I'm seeding entities which have constraints such as a username cannot exceed 24 characters etc... It would be great if we could specify this constraint when invoking a generator, similar to `unique` and `optional`. They could even be chained, not sure if `optional` and `unique` are chainable at the moment, not tested. It could look something like:

`$faker->length($min = 0, $max = 24)->unique()->userName`

Would this be a possible feature for a future release?

Diving through the src, looks like we could introduce a `LengthGenerator` class, and then a `length($min = 0, $max = null)` method to the Base provider. Can anyone see any implications on doing this?

ikwattro:
The problem is that a length constraint can not be implemented to all providers, and thus can't be applied at a global stage for chaining. For e.g. it can have undesired effetcts to use a length constraint on a Date or EAN provider.

hendore:
ikwattro, thank you for such a quick response, that makes sense.

I've just noticed some formatters such as those defined in the Lorem provider allow number of characters to be specified, however those in the Internet provider don't. Would it be more of a case of updating the formatters to take into account a length argument?

ikwattro:
Yes it would. Personally, I would : 

* Or Create my own provider on top of the base formatter present in the library (which is so easy to do)
* Or ask to implement `pre` and `post` generation events that would allow you to implement logics based on the formatter called.

fzaninotto:
That being said, I like the idea of a general purpose modifier, taking a callable as parameter. If the callable returns true, the value is valid, otherwise the formatter should generate another one.

fzaninotto:
Fixed by #836

PR #836, opened 2016-02-23
[RFR] Introducting the valid() modifier

fzaninotto:
```php
<?php

// valid() only accepts valid values according to the passed validator functions
$values = array();
$evenValidator = function($digit) {
	return $digit % 2 === 0;
};
for ($i=0; $i < 10; $i++) {
	$values []= $faker->valid($evenValidator)->randomDigit;
}
print_r($values); // [0, 4, 8, 4, 2, 6, 0, 8, 8, 6]

// just like unique(), valid() throws an overflow exception when it can't generate a valid value
$values = array();
try {
  $faker->valid($evenValidator)->randomElement(1, 3, 5, 7, 9);
} catch (\OverflowException $e) {
  echo "Can't pick an even number in that set!";
}
```

Refs #567

### DIFF
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -5,6 +5,7 @@ namespace Faker\Provider;
 use Faker\Generator;
 use Faker\DefaultGenerator;
 use Faker\UniqueGenerator;
+use Faker\ValidGenerator;
 
 class Base
 {
@@ -519,7 +520,7 @@ class Base
         if ($weight > 0 && $weight < 1 && mt_rand() / mt_getrandmax() <= $weight) {
             return $this->generator;
         }
-        
+
         // new system with percentage
         if (is_int($weight) && mt_rand(1, 100) <= $weight) {
             return $this->generator;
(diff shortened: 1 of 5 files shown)


src/Faker/Provider/Base.php
```diff
@@ -551,4 +552,9 @@ public function unique($reset = false, $maxRetries = 10000)
 
         return $this->unique;
     }
+
+    public function valid($validator = null, $maxRetries = 10000)
```
EmanueleMinotto: missing phpdoc?

src/Faker/ValidGenerator.php
```diff
@@ -0,0 +1,57 @@
...
+class ValidGenerator
+{
+    protected $generator;
+    protected $validator;
+    protected $maxRetries;
+
+    /**
+     * @param Generator $generator
+     */
+    public function __construct(Generator $generator, $validator = null, $maxRetries = 10000)
```
EmanueleMinotto: What about forcing `$validator` to be a `\Closure`?
Because without forcing it, functions not returning a boolean could be used.

fzaninotto: I want `$validator` to have a default value of accepting all values (the true function), and PHP only accepts scalar values as default arguments.

I added a check later to make sure the parameter is a callable.

PR #630, opened 2015-07-07
Add jobTitle generator

gregoryduckworth:
To address #626

PR #839, opened 2016-02-26
en_US locale provider for jobTitle added.

thudfactor:
Added a method to the en_US/Company provider to give access to the job title array instead instead of a "lorem" word.

### DIFF
--- a/src/Faker/Provider/en_US/Company.php
+++ b/src/Faker/Provider/en_US/Company.php
@@ -90,4 +90,12 @@ class Company extends \Faker\Provider\Company
 
         return join($result, ' ');
     }
+
+    /**
+     * @example 'Hand Trimmer'
+     */
+    public function jobTitle()
+    {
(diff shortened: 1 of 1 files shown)


fzaninotto:
No, it doesn't make sense to override the parent method in this class and all the others just for phpDoc.

thudfactor:
I don't understand the criticism; phpDoc had nothing to do with why I was making the pull request. 

Using the en_US region, if I call jobTitle() without making this change, all I get is a latinish word. Pull request #630 includes a jobTitle array but it doesn't appear to be accessed by anything; I thought I was following similar code elsewhere. 

It's my first pull request to this project so I probably don't sufficiently understand where something like this should go -- can you point me in the right direction?

fzaninotto:
It's because the en_US locale didn't use the right variable name. It's now fixed (see 416a132).

thudfactor:
Thanks! I think I see how that's working now.

PR #841, opened 2016-02-27
Add world countries to he_IL provider addresses

yonirom:
Added world countries in Hebrew (was TODO)
A bit of code alignment for array data, removed very long lines.


fzaninotto:
Is the datasource for name and places the same that's mentioned in the phpDoc comments?

yonirom:
Yes, Lists have not been changed. I just broke the long lines into ~200 chars.

fzaninotto:
Thanks!

PR #844, opened 2016-03-02
Removed function from loop in zh_TW provider

Newman101:
Removed  a function from loop in file ```src/Faker/Provider/zh_TW/Text.php```. In order to avoid the overhead of executing the function multiple times, I made a change to pre-calculate it before the loop begins. This increases efficiency. Thank you.


fzaninotto:
Thanks!

PR #846, opened 2016-03-03
Added unit tests for en_IN Address.php

Newman101:
Hello.

Added unit test kit for file ```Address.php``` in ```en_IN``` provider. 

Thank you.


fzaninotto:
Verifying the type of result isn't a good enough test. Please add more tests than that if you want this to be merged.

PR #848, opened 2016-03-03
docs: Clarify that randomElements() does not repeat input elements

sustmi:
It was not clear for me whether randomElements() repeats input elements or not until I looked into the implementation.

### DIFF
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -156,7 +156,7 @@ class Base
     }
 
     /**
-     * Returns random elements from a provided array
+     * Returns randomly ordered subsequence of $count elements from a provided array
      *
      * @param  array            $array Array to take elements from. Defaults to a-f
      * @param  integer          $count Number of elements to take.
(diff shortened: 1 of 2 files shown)


src/Faker/Provider/Base.php
```diff
@@ -156,7 +156,7 @@ public static function randomAscii()
     }
 
     /**
-     * Returns random elements from a provided array
+     * Returns random elements from a provided array (not repeating input indices)
```
fzaninotto: I don't understand your comment

sustmi: What about this?
> Returns randomly ordered subsequence of $count elements from a provided array

I want for the other users to be clear that the function does not repeat input elements. The current comment does not say it.

fzaninotto: ok

sustmi:
I updated the pull request with the new sentence.

fzaninotto:
Thanks!

PR #849, opened 2016-03-04
Added en_IN unit test for Address.php

Newman101:
Added unit test kit for file ```Address.php``` in ```en_IN``` provider.

The unit testing process goes as follows: 
1) Test that the variable is not empty
2) Test that internal type is string
3) Test that the string starts with upper-case letter (RegEx match)

Thank you.


fzaninotto:
Thanks

PR #850, opened 2016-03-07
Added Spanish text so we can have some nice sentences from "El Quijote".

HectorOrdonez:
It's quite simple.


fzaninotto:
I'd prefer a more recent text. 17th century Spanish may look strange in an e-commerce website.

HectorOrdonez:
The english version uses Alice in Wonderland (1865). That would look totally find in an e-commerce website, huh? :)

PR #859, opened 2016-03-22
Add PHP-CS-Fixer and bump the minimal PHP version

afurculita:
This PR adds Php-cs-fixer as a dev requirement and applies the Symfony level of fixes.

It also bumps the minimal required version of PHP to 5.5.9


fzaninotto:
No, thanks. See #728 for details.

PR #862, opened 2016-03-23
Added unit testing for en_CA provider

Newman101:
Added ```test/Faker/Provider/en_CA/AddressTest.php``` unit test class to test ```en_CA``` provider. Unit test class tests province, province abbreviation and post code letters in three assertions: 

1) Variable is not empty
2) Variable type is a string 
3) Variable matches required RegEx pattern

Thank you.


fzaninotto:
Thanks!

PR #865, opened 2016-03-29
Added AddressTest.php to ar_JO unit tests

Newman101:
Hello.

Added ```AddressTest.php``` to ar_JO provider unit tests. This test class performs the following tests:

1) Tests that the generated variable is not empty.
2) Tests that the generated variable is a string.
3) Tests that the generated string is in Arabic. 

Thank you.

### DIFF
--- a/test/Faker/Provider/ar_JO/AddressTest.php
+++ b/test/Faker/Provider/ar_JO/AddressTest.php
@@ -28,6 +28,14 @@ class AddressTest extends \PHPUnit_Framework_TestCase
         $this->assertRegExp('/\p{Arabic}/u', $cityPrefix);
     }
 
+    public function testStreetPrefix()
+    {
+      $streetPrefix = $this->faker->streetPrefix();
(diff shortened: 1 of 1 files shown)


Newman101:
ImageTest again: 

```
There was 1 failure:
1) Faker\Test\Provider\ImageTest::testDownloadWithDefaults
Failed asserting that null matches expected 640.
```

fzaninotto:
@Newman101 This is at least the 5th PR on the same subject. Extending code coverage is nice, but not if it requires my attention every week for 5 lines. I suggest you group all similar editions into one single PR. 

So I'll close this one, expecting a larger and definitive PR on the same subject, but for all locales. I'd be happy if it came from you.

Newman101:
Sure. I'll do more comprehensive PR. The reason for these PR has been to create one pull request per one issue. If making more comprehensive PR's (where multiple locales are being improved at the same time), I'll change the approach.

PR #869, opened 2016-03-30
Add greek mobile and toll-free phone numbers

sebdesign:
Source: https://en.wikipedia.org/wiki/Telephone_numbers_in_Greece


fzaninotto:
Can you add the link to the Wikipedia reference in a phpDoc comment in the class you modified?

sebdesign:
@fzaninotto Sorry for the late reply, I hope it's acceptable now.

fzaninotto:
Thanks!

PR #879, opened 2016-04-12
add Provider of Word,Sentence,Paragraph and Text for zh_CN

lbp0200:
add Provider of Lorem for zh_CN


fzaninotto:
No, Lorem isn't localizable. Use the realText formatter for that purpose. I know it doesn't allow the same granularity as the Lorem provider, but that's a limitation I prefer to live with rather than translating Lorem in all languages (which doesn't make sense).

PR #880, opened 2016-04-12
Fix typo

xelan:
Dorins is most likely a typo. While Doris is a common name in Austria, Dorins isn't :smile:


fzaninotto:
Thanks.

PR #883, opened 2016-04-16
Improved Hungarian provider

balping:


fzaninotto:
What's a Male Ne? Could you please add a description?

balping:
It's a special name form for wives. Like Mrs. title in English. For example: if Kiss Piroska (female) marries Kovács József (male) then she has the choice of changing her name to Kovács Józsefné. As you can see, the wife is called on the husband's name, but with a -né suffix. It's a bit old-fashioned nowadays, but still many women do this after marriage.

fzaninotto:
Thanks.

PR #887, opened 2016-04-21
Remove jobTitle from readme

crossRT:
Unknown formatter "jobTitle"
README remains the 'jobTitle' property which is already removed from company provider.


fzaninotto:
I don't understand, `jobTitle` [is in the company provider](https://github.com/fzaninotto/Faker/blob/master/src/Faker/Provider/Company.php#L38-L43).

PR #888, opened 2016-04-21
Add job titles for fr_FR locale

PedroTroller:
French set of job titles from http://www.emploi.com/emploi/index-metiers.html


fzaninotto:
The source is © Keljob, I can't put that in a MIT project.

Issue #881, opened 2016-04-13
License Contamination


y-takshima:
Based on the LICENSE and the CONTRIBUTING.md, the software seems to be under the MIT license. But when I search "license" string within this repository, there are some files described as a creative commons.

https://github.com/fzaninotto/Faker/search?utf8=%E2%9C%93&q=CREATIVE+COMMONS

Should I consider all the repository files under the MIT license or not?

Newman101:
Looks like some of the texts are being taken from Ivan Franko's poem "Zakhar Berkut". Ivan Franko died 1916 so as such, at least that text should be under public domain. If for some reason it's not under public domain, then the text can be replaced with a text which is.

fzaninotto:
No, some portions are Creative Commons, although they all concern public domain novels. Do you have any advice on how to present that?

y-takshima:
I believe that it is better not to use CC BY-SA content inside OSS. It may not be re- distributed as the MIT license since ShareAlike terms affects.

> ShareAlike
 — If you remix, transform, or build upon the material, you must distribute your contributions under the same license as the original.

okj579:
It is possible to distribute software under a mixed license. However, it would probably be easier to make sure everything is either public domain or MIT.

The Khazakh text is a Wikipedia article and truly CC. It should be probably be replaced anyway.

The Russian and Ukranian texts are in the public domain. They were probably copied from Wikipedia/Wikisource, but that doesn't matter if they don't have any effective copyright themselves. The copyright notices can be removed, as they are incorrect.

The Farsi text is fairly recent and is not public domain (till at least 2019 when it passes into the public domain in Iran and some other countries). As far as I can tell, it has not been released under CC or any other permissive license, so its current use on Wikisource and here wouldn't be legal.

fzaninotto:
Fine by me. Could you open a pull request with these changes ?

Newman101:
I can make the PR. Before that, I need to google some public domain resources for Kazakh and Farsi texts. Any ideas where to start?

Newman101:
Also, are there any length requirements?

okj579:
I started looking and found a book by Abay Qunanbayuli (Kazakh) on
Wikisource. So far, I've been unable to find modern Persian prose old
enough to be in the public domain.

fzaninotto:
If you can't find any, then simply remove the one published in Faker.

Newman101:
[...]

PR #892, opened 2016-04-23
uk_UA & ru_RU: Removed @license from Text.php

Newman101:
Removed ```@license``` from Text.php in both ```uk_UA``` and ```ru_RU``` locales. Since the texts are in public domain no license notice is required. 

This should be the last part of solution to ticket #881.


fzaninotto:
Thanks

PR #897, opened 2016-04-27
Improve german titles

christianbartels:
Always include "Herr" or "Frau" in the title (formal salutation always includes the gender in German)


fzaninotto:
Thanks!

PR #898, opened 2016-04-27
New method in Internet Provider - youtube()

giordanolima:
New method in Internet Provider - youtube() - Return a Youtube url of a random video (Using randomyt.com service)


fzaninotto:
Thanks, but I'll pass. I'm not sure this is very useful in the core. You're welcome to create a third-party provider, though ; I would include it in the README.

PR #899, opened 2016-04-28
Add job titles for fr_FR locale

PedroTroller:
@fzaninotto Same things than https://github.com/fzaninotto/Faker/pull/888 but under CC license.

PR #900, opened 2016-04-29
Revert "Add job titles for fr_FR locale"

fzaninotto:
Reverts fzaninotto/Faker#899, because the list is of bad quality (some English names, duplicates, non-job titles, etc.)


fzaninotto:
See the PR description

PR #904, opened 2016-05-02
Fix French Cell phone number format

lbeauvisage:
All numbers starting by 07 are not valid.
According to http://www.arcep.fr/?id=8146, only number starting by 073 to 079 are valid


fzaninotto:
That's not the right way to do, you're giving 07 numbers a very high chance of being generated.

lbeauvisage:
OK @fzaninotto , I understand the problem, Is there any way to use REGEX ? Can you provide me an example ?

fzaninotto:
yep, see `regexify()`

lbeauvisage:
ok @fzaninotto, I make some test and I think I have the good regex:

`(\+33 |0)(6( )?[0-9]{1}|7( )?([3-9]){1})([0-9]{1})( )?([0-9]{2})( )?([0-9]{2})( )?([0-9]{2})`

But regexify() does not support the `|` in `(6( )?[0-9]{1}|7( )?([3-9]){1})` as I am expecting
Any idea ?

fzaninotto:
yes, `regexify` has limiter regex support - and very slow.

I guess you could try generating 07 ## ## ## ## in a `while (true)` loop, exiting as soon as a valid phone numer comes out (based on your regexp).

fzaninotto:

PR #905, opened 2016-05-02
Add Indonesia vehicle license (police) number

matriphe:
This method will produce something like: `B 1234 XYZ`, `AD 1234 VW`, `RI 1`, or `CD 53`

### DIFF
--- a/src/Faker/Provider/id_ID/Person.php
+++ b/src/Faker/Provider/id_ID/Person.php
@@ -236,7 +236,7 @@ class Person extends \Faker\Provider\Person
     );
 
     private static $vehicleThirdSuffixCode = array(
-        '', 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M',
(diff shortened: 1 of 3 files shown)


src/Faker/Provider/id_ID/Person.php
```diff
@@ -244,13 +285,68 @@ public static function lastNameFemale()
...
+    }
+
+    /**
+     * Return vehicle license (police) number.
+     *
+     * @return string vehicle license number
+     */
+    public static function personalLicenseNumber()
+    {
+        $license = static::randomElement(static::$vehicleAreaCode);
```
fzaninotto: better: use `$this->generator->parse()` with a format string. See other providers as a usage example.

fzaninotto:
Can you please rebase?

matriphe:
Oops.. This branch had left so far behind. Close it for now.

matriphe:
I think it would be better to go different provider, for example `Vehicle` provider.

PR #906, opened 2016-05-02
Added common free E-mail provider

ludo237:


fzaninotto:
Too low value. Each time you update the dataset, it breaks unit tests for users relying on seeding. It should only be done for a good reason.

PR #908, opened 2016-05-08
Fix: Prefer dependencies installed from dist

localheinz:
This PR

* [x] uses the `--prefer-dist` option when installing dependencies in `Makefile` target

💁 It's faster, and it's also more in line with `.travis.yml`


fzaninotto:
Thanks!

PR #909, opened 2016-05-08
Fix: Do not pick a random float less than minimum

localheinz:
This PR

* [x] fixes an issue where a value generated by `randomFloat($nbMaxDecimals, $min)` can actually be less than `$min`

### DIFF
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -110,6 +110,9 @@ class Base
 
         if (null === $max) {
             $max = static::randomNumber();
+            if ($min > $max) {
+                $max = $min;
+            }
         }
 
         if ($min > $max) {


src/Faker/Provider/Base.php
```diff
@@ -110,6 +110,9 @@ public static function randomFloat($nbMaxDecimals = null, $min = 0, $max = null)
 
         if (null === $max) {
             $max = static::randomNumber();
```
localheinz: @fzaninotto 

`$max` can be `0` here.

src/Faker/Provider/Base.php
```diff
@@ -110,6 +110,9 @@ public static function randomFloat($nbMaxDecimals = null, $min = 0, $max = null)
...
+            if ($min > $max) {
+                $max = $min;
+            }
         }
 
         if ($min > $max) {
```
localheinz: @fzaninotto 

Without the change proposed in this PR, `$min` would then be set to `0`, and eventually return `0`, too, in rare cases.

fzaninotto:
Thanks!

PR #911, opened 2016-05-10
Add Romanian Text

thadumi:


fzaninotto:
The referenced text is copyrighted, Faker can only bundle MIT-compatible content.

PR #912, opened 2016-05-12
Add ISIN (International Securities Identification Number)

dirkolbrich:
Could be the start of a financial section for Faker.


fzaninotto:
Hi, and thanks for your PR. I don't think this provider is generic enough to be added to the Faker core, but you're welcome to create a custom repository for this provider, and I'll gladly link to it in the README.

PR #913, opened 2016-05-14
ka_GE: overall improvements to ka_GE locale

hertzg:
General improvemens were made to existing Provides as well
as added Company and Lorem providers (generators)

ka_GE\Address:
* trimed items in arrays
* added multiple address formats (full, short and etc.)
* added cityFormats
* added regionGenitiveForm (gramar)

ka_GE\Company:
* generates names using popular name elements with affixes

ka_GE\Internet:
* added items to freeEmailDomain array
* added all supported Georgian TLDs

ka_GE\Lorem:
* added about 300 localized lorem words

ka_GE\Person:
* added titles for genders


fzaninotto:
Lorem is never localized, please remove this class.

src/Faker/Provider/ka_GE/Company.php
```diff
@@ -0,0 +1,59 @@
+<?php
+/**
+ * Created by IntelliJ IDEA.
```
fzaninotto: please remove this phpDoc

hertzg:
@fzaninotto Is there any specific reason why Lorem is never localized? Because when I'm using `$faker->word`, `$faker->sentence`, `$faker->paragraph` and friends on a localized `Generator` it makes sense for the result to be localized. It's also very useful when testing specific localized fonts (in PDF's for example or some other generated content as well as website).

okj579:
There's a provider for that. It's called Text.

hertzg:
@okj579 The `Text` provider provides data only for `$faker->realText()` which generates text up to `$maxNbChars` as opposed to `$faker->word`, `$faker->sentence`, `$faker->paragraph` and similar functions do.

To rephrase: When I want to generate a random localized word, sentence or paragraph the most intuitive would be to call the respective methods on Generator instance. 

There's no good way of doing those using realText as it generates a text up to $maxNbChars. Also realText could return newlines, parts of a word, punctuations and what not.

So again: @fzaninotto Could you please guide me why "Lorem is never localised"?

fzaninotto:
Lorem is for text that is in no language. For real language, use the Text provider.

If you need words and sentences with realtext, then we should add new methods (like `realWord`) to the Text provider, not localize Lorem.

hertzg:
@fzaninotto Thanks for the explanation.

hertzg:
@fzaninotto I've removed the lorem provider.

> If you need words and sentences with realtext, then we should add new methods (like realWord) to the Text provider, not localize Lorem.

Yes. I definitely need a way to generate dummy localized `words`, `sentences` and `paragraphs` (similar to what `Lorem` provides, but localized).

Are you willing to add such functionality to Faker and if so how do you think would be the best way to implement it?

fzaninotto:
Thanks for this PR. For localized text, I suggest you work on the base `Text`provider, and add `realWord`, `realSentence`, and `realParagraph`formatters - that should be enough to make them work in all languages. Then, if you're satisfied with the result, please open a new PR.

PR #917, opened 2016-05-19
Fix: fix invalid parameter of mb_substr()

tkawaji:
In this PR, fix bug which call mb_substr() with invalid parameter in ja_JP Text.

error message when call $faker->realText:
`
mb_substr() expects parameter 3 to be integer, string given
`


vlakoff:
We could do more simply:

    mb_substr($text, -1, null, 'UTF-8');

fzaninotto:
Thanks!

tkawaji:
@vlakoff  Thank you for your advice! I will fix them next time.

@fzaninotto  Thank you for merging.

PR #918, opened 2016-05-20
Harmonize fr_*\Company

Max13:


src/Faker/Provider/fr_CA/Company.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Faker\Provider\fr_CA;
+
+class Company extends \Faker\Provider\fr_FR\Company
+{
+    //
```
fzaninotto: please remove this line

src/Faker/Provider/fr_FR/Company.php
```diff
@@ -13,7 +13,6 @@ class Company extends \Faker\Provider\Company
         '{{lastName}} {{companySuffix}}',
         '{{lastName}} {{lastName}} {{companySuffix}}',
         '{{lastName}}',
-        '{{lastName}}',
```
fzaninotto: Duplicates have a function in format lists: making the duplicated patter occur more often. Please remove this change.

src/Faker/Provider/fr_FR/Company.php
```diff
@@ -166,4 +165,16 @@ protected static function isCatchPhraseValid($catchPhrase)
 
         return true;
     }
+
+    /**
+     * @link http://www.emploi.com/emploi/index-metiers.html
```
fzaninotto: this content is `© Keljob.com tous droits réservés` and can't be reused in a MIT library.

src/Faker/Provider/fr_CA/Company.php
Max13: Are you talking about line 7 ?

src/Faker/Provider/fr_FR/Company.php
Max13: Got it. I will rebase.

src/Faker/Provider/fr_FR/Company.php
Max13: 😓 I understand that and didn't check. It's actually just a list of job titles in French... How can it be "reserved"?

Anyway, removing the link will make the difference between this list and a list made by hand... But before that, I have asked the site if they can give us the permission to reuse the list. Would it be OK ?

src/Faker/Provider/fr_CA/Company.php
fzaninotto: yes

src/Faker/Provider/fr_FR/Company.php
fzaninotto: nah, we use a public domain / CC license datasource

Max13: Too bad. I got the [permission](http://i64.tinypic.com/2eyj49j.png) and they can't guarantee that this list will still be available in the future, as the website will evolve. Even if I remove this list and make another one by hand, I don't think it's allowed to reserve a list of common names.

Anyway, if it's still not OK, I will make one by hand with just few names. Then I'll be sure there wouldn't be any problem.

fzaninotto: I suggest you use a public list, like [the list of ROME codes from INSEE](http://www.pole-emploi.fr/candidat/le-code-rome-et-les-fiches-metiers-@/article.jspz?id=60702). It might need filtering as it's very large, but it's a good (and public) datasource.

Max13:
@fzaninotto I have amended the commits, thanks

src/Faker/Provider/fr_FR/Company.php
```diff
@@ -166,4 +166,111 @@ protected static function isCatchPhraseValid($catchPhrase)
 
         return true;
     }
+
+    /**
+     * @link http://www.pole-emploi.fr/candidat/le-code-rome-et-les-fiches-metiers-@/article.jspz?id=60702
+     * @note Randomly took 100 from this list
+     */
+    protected static $jobTitleFormat = array(
+        'Elagueur-botteur',
```
fzaninotto: Please sort these alphabetically

src/Faker/Provider/fr_FR/Company.php
```diff
@@ -166,4 +166,111 @@ protected static function isCatchPhraseValid($catchPhrase)
 
         return true;
     }
+
+    /**
+     * @link http://www.pole-emploi.fr/candidat/le-code-rome-et-les-fiches-metiers-@/article.jspz?id=60702
+     * @note Randomly took 100 from this list
```
fzaninotto: feel free to take a bit more (250)

src/Faker/Provider/fr_FR/Company.php
```diff
@@ -166,4 +166,311 @@ protected static function isCatchPhraseValid($catchPhrase)
 
         return true;
     }
+
+    /**
+     * @link http://www.pole-emploi.fr/candidat/le-code-rome-et-les-fiches-metiers-@/article.jspz?id=60702
+     * @note Randomly took 300 from this list
```
Max13: @fzaninotto At that point... Why not make it 300 ?

fzaninotto:
Thanks!

PR #919, opened 2016-05-20
Modified the phone number pattern to match right pattern for bn/BD loc…

prio101:
# Modified the phone number pattern to match up the right pattern for bn/BD .
-  added `static::numerify()`

### DIFF
--- a/src/Faker/Provider/bn_BD/PhoneNumber.php
+++ b/src/Faker/Provider/bn_BD/PhoneNumber.php
@@ -7,7 +7,7 @@ class PhoneNumber extends \Faker\Provider\PhoneNumber
     public function phoneNumber()
     {
         $number = "+880";
-        $number .= static::randomNumber(7);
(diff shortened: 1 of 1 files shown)


src/Faker/Provider/bn_BD/PhoneNumber.php
```diff
@@ -7,7 +7,7 @@ class PhoneNumber extends \Faker\Provider\PhoneNumber
     public function phoneNumber()
     {
         $number = "+880";
-        $number .= static::randomNumber(7);
+        $number .= static::numerify($string = "##########");
```
okj579: I don't think this does what you think it does. PHP doesn't have named parameters, so the `$string = ` is unnecessary.

fzaninotto: better use `static::randomNumber(7, true);`, it's much faster.

fzaninotto:

PR #929, opened 2016-06-03
Remove Gaylord from list of names

thomasedwards:
Because the domain faker uses this list, I ended up with `gaylord.com` – which out of context can look deliberate and when our tests are client-facing, looks derogatory.


fzaninotto:
That's the beauty of randomness! Gaylord IS a valid english name, see https://en.wikipedia.org/wiki/Gaylord. Besides, what's wrong about http://www.gaylord.com?

PR #931, opened 2016-06-07
Improved en_GB and ro_RO 

the-noob:
en_GB
- improved postcode generator although not guaranteed to produce valid codes
- landline formatter for phone numbers

ro_RO
- adding real text

### DIFF
--- a/src/Faker/Provider/en_GB/Address.php
+++ b/src/Faker/Provider/en_GB/Address.php
@@ -86,4 +86,25 @@ class Address extends \Faker\Provider\Address
     {
         return static::randomElement(static::$county);
     }
+
+    /**
+     * Attempt to generate something that looks like valid postcode
+     *
+     * @link http://stackoverflow.com/questions/164979/uk-postcode-regex-comprehensive
+     * @example 'N6 5AA'
+     */
+    public static function postcode()
+    {
+        $attempts = 10;
+        do {
+            $attempts--;
+            $result = parent::postcode();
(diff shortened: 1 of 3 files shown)


src/Faker/Provider/en_GB/Address.php
```diff
@@ -86,4 +86,25 @@ public static function county()
...
+    /**
+     * Attempt to generate something that looks like valid postcode
+     *
+     * @link http://stackoverflow.com/questions/164979/uk-postcode-regex-comprehensive
+     * @example 'N6 5AA'
+     */
+    public static function postcode()
+    {
+        $attempts = 10;
+        do {
```
fzaninotto: I don't like the iterative approach. You should fix the format instead.

src/Faker/Provider/ro_RO/Text.php
```diff
@@ -0,0 +1,153 @@
...
+class Text extends \Faker\Provider\Text
+{
+    /**
+     * Title: Neghiniţă
+     *
+     * Author: Barbu Ştefănescu Delavrancea
+     *
+     * Language: Romanian
+     *
+     * @see http://www.povesti-pentru-copii.com/barbu-stefanescu-delavrancea/neghinita.html
```
fzaninotto: The site mentions a copyright, and we need a public license text. Also, please make a separate PR for that.

the-noob:
I'll implement the suggestions and split the changes into 2 separate PRs

PR #938, opened 2016-06-16
Create Text.php

kongoon:


src/Faker/Provider/th_TH/Text.php
```diff
@@ -0,0 +1,32 @@
+<?php
+namespace Faker\Provider\th_TH;
+class Text extends \Faker\Provider\Text
+{
+    /**
+     * @see http://lorem.in.th
```
fzaninotto: the content of this site is copyrighted, and can't be used in Faker (MIT license)

PR #948, opened 2016-06-27
Add $timeout to Image() parameter

andrey-helldar:
Correct parameters for image() method and added a new parameter (timeout).


andrey-helldar:
Updated requests :-)

fzaninotto:
The `image` formatter should not be an interface for all the possible options of a general purpose HTTP client (proxy, headers, etc). If you have specific needs like a timeout, please write your own implementation.

andrey-helldar:
I am developing a website. When you try to download a large number of images, we have to wait very long. Timeout solves the problem.

```
$users = User::get(); // 200 records in table generated by faker.
$images = [];
$faker = Faker\Factory::create();

foreach($users as $user) {
    for ($i = 0; $i < $faker->numberBetween(1, 12); $i++) {
        $images[] = $faker->unique()->image(storage_path('app/public/tmp/' . $user->id . '/'), 1920, 1080, null, false, true, null, 10); // Timeout 10 sec.
    }
}
```

If you remove a timeout, loading images takes about 40-80 minutes, because 30% of the server is not responding on certain links.

P.S.: I understood you. I think about how you can implement properly.

PR #951, opened 2016-06-29
Fix EAN8 checkSum generator

MatthieuMota:
Fix eanChecksum function who check bad length for input. Correct sequence for ean8 checkSum is 3, 1 and not 1, 3. [See on Wikipedia](https://fr.wikipedia.org/wiki/EAN_8).
@fzaninotto Sorry for #950, it was a PR demo for web training session :smiley:


fzaninotto:
Thanks!

PR #952, opened 2016-06-30
Simplify and optimize randomElements()

vlakoff:
Figures, on 100000 iterations:

* For several random elements: before 590 ms, after 320 ms
* For one random element: before 390 ms, after 120 ms

### DIFF
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -169,27 +169,21 @@ class Base
      */
     public static function randomElements(array $array = array('a', 'b', 'c'), $count = 1)
     {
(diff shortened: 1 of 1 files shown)


vlakoff:
Tests failing. Investigating.

vlakoff:
Fixed an issue with $count = 0, but remaining test failures seem unrelated…

vlakoff:
Looks like it's because of the change from `mt_rand()` to `array_rand()`, which discards the `seed()`. So if you want Faker to be possibly deterministic using a seed, I guess this PR has to be dropped…

PR #954, opened 2016-07-01
Remove 'Stripper' from en_US job titles

amcsi:
This entry that's probably there for laughs is the reason the company I'm working for is making me ditch the entire Faker library.

We're using Faker to generate fake data to show clients what our platform looks like.
I believe this inappropriate entry is doing more harm than you would think.


fzaninotto:
You're right, thanks for the PR.

PR #959, opened 2016-07-07
Adding Text for ro_RO

the-noob:
Same text as in previous PR but this time with proof of public domain copyright and source.

### DIFF
--- a/src/Faker/Provider/ro_RO/Text.php
+++ b/src/Faker/Provider/ro_RO/Text.php
@@ -1,0 +1,158 @@
+<?php
+
+namespace Faker\Provider\ro_RO;
+
+class Text extends \Faker\Provider\Text
+{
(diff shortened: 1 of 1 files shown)


src/Faker/Provider/ro_RO/Text.php
```diff
@@ -0,0 +1,158 @@
+<?php
+
+namespace Faker\Provider\ro_RO;
+
+class Text extends \Faker\Provider\Text
+{
+    /**
+     * From uk.wikisource.org
+     * Text available under Creative Commons Attribution-ShareAlike 3.0 Unported
```
okj579: Since the work is in the public domain (the author died in 1918), this can be removed.

fzaninotto:
Thanks!

PR #960, opened 2016-07-07
adding postcode generator for en_GB

the-noob:


the-noob:
seems that running 3 local tests is not enough :)
will provide a large array with valid codes

PR #961, opened 2016-07-10
Adding valid en_GB postcodes

the-noob:
Using `bothify` provides too many invalid post codes and the validator is too complex for `regexify` (attempts to chunk it failed)
Ended up adding 500 valid postcodes.


fzaninotto:
Thanks!

PR #963, opened 2016-07-13
Update Image.php

gulaandrij:
fix broken image if service unavailable

### DIFF
--- a/src/Faker/Provider/Image.php
+++ b/src/Faker/Provider/Image.php
@@ -76,7 +76,7 @@ class Image extends Base
             curl_setopt($ch, CURLOPT_FILE, $fp);
             $success = curl_exec($ch) || curl_getinfo($ch, CURLINFO_HTTP_CODE) === 200;
 
-            if ($success){
+            if ($success) {
                 fclose($fp);
             } else {
                 unlink($filepath);
--- a/src/Faker/Provider/Image.php
+++ b/src/Faker/Provider/Image.php
@@ -74,9 +74,15 @@ class Image extends Base
             $fp = fopen($filepath, 'w');
             $ch = curl_init($url);
             curl_setopt($ch, CURLOPT_FILE, $fp);
-            $success = curl_exec($ch);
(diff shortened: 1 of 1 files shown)


gulaandrij:
any feedback on this?

fzaninotto:
Thanks!

src/Faker/Provider/Image.php
```diff
@@ -74,9 +74,15 @@ public static function image($dir = null, $width = 640, $height = 480, $category
             $fp = fopen($filepath, 'w');
             $ch = curl_init($url);
             curl_setopt($ch, CURLOPT_FILE, $fp);
-            $success = curl_exec($ch);
+            $success = curl_exec($ch) || curl_getinfo($ch, CURLINFO_HTTP_CODE) === 200;
+
+            if ($success){
+                fclose($fp);
```
ethanclevenger91: Shouldn't this happen regardless of success?

src/Faker/Provider/Image.php
```diff
@@ -74,9 +74,15 @@ public static function image($dir = null, $width = 640, $height = 480, $category
...
+            } else {
+                unlink($filepath);
+            }
+
             curl_close($ch);
-            fclose($fp);
         } elseif (ini_get('allow_url_fopen')) {
```
ethanclevenger91: Shouldn't the same checks go here in case `curl` isn't enabled?

gulaandrij:
> Silently deleting the image seems like unexpected behavior.
if lorempixel is not available then the image is saved as spoiled, 0 kb, so there is no need to store it

ethanclevenger91:
Yeah, I get that much. But if I expect all my images to be generated, some are inexplicably missing, and I'm not versed in the intricacies of this issue, what do I do?

gulaandrij:
in my opinion its a good catch to add additional check 
```
if (!$faker->image()) {

} 
```

ethanclevenger91:
But once again, if I'm a user who has no idea this is an issue, how would I know to do that?

I certainly don't do anything like:
```
if (!$faker->text()) {

}
```

PR #970, opened 2016-07-19
Fix a mixup between male and female last names in Icelandic.

arthur-olafsson:
Last names for females usually end in "dóttir", meaning "daughter".
Last names for males usually end in "son".

See: https://en.wikipedia.org/wiki/Icelandic_name#Typical_Icelandic_naming


vlakoff:
Indeed, mistake introduced in https://github.com/fzaninotto/Faker/pull/114/files#diff-2bb60c1a0d17a1e6ff6c646fc914f3be (page slow to load).

fzaninotto:
Thanks!

PR #936, opened 2016-06-14
Fix to Issue #935 - German Locale

Newman101:
Removed the name with a dot. This fixes the malformed email problem described in #935 for German locale.

Issue #935, opened 2016-06-10
Faker sometimes generates invalid email addresses


prolic:
F.e. this here:

witte.hansd.@feldmann.org

Newman101:
This can be fixed easily. The issue is in variable ```$firstNameMale``` of German locale. Looks like the email formation takes first name as it is and removes capital letters. Since one of the names (Hans D.) includes a dot, result is malformed e-mail address. I'll send a PR to fix this issue.

Newman101:
As a side note, would be good idea to amend unit tests to make sure email addresses in the future are not malformed.

curry684:
Technically that is not an invalid email address.

KaelWD:
[Tell that to the IETF:](https://tools.ietf.org/html/rfc3696#page-6)

>   Without quotes, local-parts may consist of any combination of
>   alphabetic characters, digits, or any of the special characters
>   ``! # $ % & ' * + - / = ?  ^ _ ` . { | } ~``
>   period (".") may also appear, but may not be used to start or end the
>   local part, nor may two or more consecutive periods appear.

curry684:
Granted, but by that RFC no modern mail service is compliant. Pragmatism over idealism: http://girders.org/blog/2013/01/31/dont-rfc-validate-email-addresses/

If someone wants to fix this case sure, but not for the sake of following that RFC, that would require Faker being able to generate a lot of fubar mail addresses.

hypeJunction:
Email should at least pass the ``filter_var($address, FILTER_VALIDATE_EMAIL)`` check

Einenlum:
I think this issue is solved thanks to #975, right?

pimjansen:
I think this would have solved it yes. However i think it is wise to run a cross check to be 100% sure

pimjansen:
I did a basic test with below. 0 occurences on 100k so im closing this issue

```php
<?php
require 'vendor/autoload.php';

$faker = Faker\Factory::create('de_DE');

for ($i=0; $i<100000;$i++) {
    $parts = explode('@', $faker->email);
    $lastChar = substr($parts[0], -1);
    if ($lastChar === '.') {
        die('Invalid email char');
    }
    echo '.';
}

curry684:
Why not the `filter_var($address, FILTER_VALIDATE_EMAIL)` test?

pimjansen:
[...]

PR #975, opened 2016-07-24
Remove trailing dot in username if any

vlakoff:
This change to prevent #935, without having to rely on specific fixes as in #936.

### DIFF
--- a/src/Faker/Provider/Internet.php
+++ b/src/Faker/Provider/Internet.php
@@ -100,6 +100,10 @@ class Internet extends \Faker\Provider\Base
             throw new \Exception('userName failed with the selected locale. Try a different locale or activate the "intl" PHP extension.');
         }
 
+        // clean possible trailing dots from first/last names
+        $username = str_replace('..', '.', $username);
+        $username = rtrim($username, '.');
+
         return $username;
     }
     /**
@@ -133,6 +137,9 @@ class Internet extends \Faker\Provider\Base
         if (trim($lastName, '._') === '') {
             throw new \Exception('domainWord failed with the selected locale. Try a different locale or activate the "intl" PHP extension.');
         }
+
+        // clean possible trailing dot from last name
+        $lastName = rtrim($lastName, '.');
 
         return $lastName;
     }
--- a/src/Faker/Provider/Internet.php
+++ b/src/Faker/Provider/Internet.php
@@ -214,7 +214,7 @@ class Internet extends \Faker\Provider\Base
         }
 
         $transId = 'Any-Latin; Latin-ASCII; NFD; [:Nonspacing Mark:] Remove; NFC;';
-        if (function_exists('transliterator_transliterate') && $transliterator = \Transliterator::create($transId)) {
+        if (class_exists('Transliterator') && $transliterator = \Transliterator::create($transId)) {
             $transString = $transliterator->transliterate($string);
         } else {
             $transString = static::toAscii($string);
--- a/src/Faker/Provider/Internet.php
+++ b/src/Faker/Provider/Internet.php
@@ -93,7 +93,14 @@ class Internet extends \Faker\Provider\Base
         $format = static::randomElement(static::$userNameFormats);
         $username = static::bothify($this->generator->parse($format));
 
-        return strtolower(static::transliterate($username));
+        $username = strtolower(static::transliterate($username));
+
+        // check if transliterate() didn't support the language and removed all letters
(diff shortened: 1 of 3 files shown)


vlakoff:
ping @prolic @Newman101 :)

prolic:
good for me, but I don't maintain this project, cannot merge :)

vlakoff:
There is a TravisCI failure, on PHP 5.3 only. Interesting.

prolic:
PHP 5.3 is no longer supported by internals. So maybe it's time to drop it
for faker, too.

Sascha-Oliver Prolic

2016-07-24 12:55 GMT+08:00 vlakoff <notifications@github.com>:

vlakoff:
The 2 failures seem to be due to transliteration returning empty string.

Probably the 5.3 box is configured differently, anyway it's not caused by this PR.

vlakoff:
@prolic This may be more about a different set of PHP extensions enabled.

But that issue is a different one, unrelated to this PR.

vlakoff:
The issue happens when `intl` PHP extension is not installed, so that `Internet::transliterate()` uses our internal `Internet::toAscii()` instead of the `intl` extension.
In such situation, `userName` can result in a single dot "." if the language is not supported by `toAscii()`.

This "single dot" value is obvisouly wrong, but it passes the `TEST_STRING_REGEX = '/.+/u'` in `ProviderOverrideTest`.
My PR removes this dot so the result becomes an empty string, hence the failing tests.

I can't think of a good way to fix this… suggestions?

vlakoff:
Added some workaround for this. If `transliterate()` ate all letters, we return `$faker->word` as a fallback.

It would be better to generate a new value, but by using the en_US locale. Is this doable?

vlakoff:
Just pushed a better fallback. Generate and return a new value, but using en_US locale.

vlakoff:
Just pushed improvements to the tests, they are much robuster now.

Refs #490, in particular https://github.com/fzaninotto/Faker/pull/490/commits/eeaa79c99f744525bb07d7e096ce36d416a24fb1.

src/Faker/Provider/Internet.php
```diff
@@ -120,7 +130,15 @@ public function domainWord()
     {
         $lastName = $this->generator->format('lastName');
 
-        return strtolower(static::transliterate($lastName));
+        $lastName = strtolower(static::transliterate($lastName));
+        $lastName = rtrim($lastName, '.');
+
+        // fallback in case transliterate() doesn't support the language and removed all letters
```
fzaninotto: I don't like that fallback. If the locale doesn't manage to create username, it's the user's choice to replace it by another locale. 

In case the `$lastName` is empty, I prefer that we throw an exception here. It will break code, but at least that'll let users deal with it the way they want.

vlakoff: Added a commit to throw exception instead. Still, this may be refined.

Please tell me before merging, I'd like to squash the commits a bit once we have settled on the end result.

vlakoff:
Wow, that escalated quickly.
At the beginning I was just adding a `rtrim($name, '.')`… :trollface:

Squashed the commits a bit, hope it will help reading.

test/Faker/Provider/InternetTest.php
```diff
@@ -40,29 +40,65 @@ public function localeDataProvider()
...
      */
     public function testUsernameIsValid($locale)
     {
+        if ($locale !== 'en_US' && !class_exists('Transliterator')) {
+            $this->markTestSkipped('Transliterator class not available (intl extension)');
+        }
+
         $this->loadLocalProviders($locale);
-        $pattern = '/^[A-Za-z0-9._]+$/';
+        $pattern = '/^(?![._])[A-Za-z0-9._]+(?<![._])$/';
```
vlakoff: Alternative: `/^[A-Za-z0-9]([._][A-Za-z0-9]+)*$/`

vlakoff: Alternative: `/^[A-Za-z0-9]+([._][A-Za-z0-9]+)*$/`

vlakoff: PR has been updated with this alternative, which I think is clearer. It's also stricter (forbid consecutive separators).

fzaninotto:
Great! Can you rebase?

vlakoff:
There were no conflicts. Still, rebased.

fzaninotto:
Thanks!

PR #980, opened 2016-07-27
Add "gray" argument to "image" method

unlikenesses:
Allows you to request grayscale images from lorempixel.com when calling the `image` method (previously only possible on the `imageUrl` method).


localheinz:
Duplicate of #1714.

PR #981, opened 2016-07-28
Added 'bank' formatter to Payment provider which returns bank name.

Powerhead13:
Added 'bank' formatter to Payment provider  which returns one of 150 largest US bank names.
Source: http://onlinebankingreport.com/resources/100.html


fzaninotto:
The data isn'y open-source, closing.

Issue #982, opened 2016-07-28
Phpunit tests fails on 64-bit systems


Powerhead13:
I'm seeing 2 tests fails on Win 10 64 bit.
This definitely happens because of PHP integer overload on 64 bit systems.
This can be fixed by adding quotes to dataprovider values in `LuhnTest` to convert em to a strings.
Especially, values `561059108101825 `and `601100099013942`
```
There were 2 failures:

1) Faker\Test\Calculator\LuhnTest::testComputeCheckDigit with data set #9 (561059108101825, '0')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'0'
+'7'

E:\Projects\faker-fork\test\Faker\Calculator\LuhnTest.php:34

2) Faker\Test\Calculator\LuhnTest::testComputeCheckDigit with data set #10 (601100099013942, '4')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'4'
+'3'
```

Powerhead13:
I created PR to fix that https://github.com/fzaninotto/Faker/pull/983

fzaninotto:
Fixed by #983

PR #983, opened 2016-07-28
Fix phpunit tests fail on 64-bit systems #982

Powerhead13:
Fixed issue https://github.com/fzaninotto/Faker/issues/982


fzaninotto:
Thanks!

PR #991, opened 2016-08-14
Add Commerce Provider

ahmadpriatama:
Adding Commerce: Product Name, Department Name, Material Name


fzaninotto:
Duplicate of #259, and same answer: I don't want to integrate that in the Faker core, but you're welcome to publish it as a standalone provider, and I'll reference it in the README.

PR #998, opened 2016-08-22
Add generator of solid images and save on the server

alanfabiano:
I found it interesting to create a method that creates a fake image directly on the server, this facilitates the manipulation of images of records in the database. The lorempixel.com site is a bit slow to download and create the images, large volumes of data becomes impractical to use the download feature.


fzaninotto:
Can you post examples of the generated images?

alanfabiano:
createImage($dir = 'storage/app', $width = 200, $height = 380, $fullPath = false, $word = "image generated with Faker")
createImage($dir = 'storage/app', $width = 640, $height = 480, $fullPath = false, $word = "image generated with Faker")
createImage($dir = 'storage/app', $width = 640, $height = 480, $fullPath = false, $word = null)
createImage($dir = 'storage/app', $width = 100, $height = 100, $fullPath = false, $word = null)
10d45582dd1bc11612283fdc46c2f15f
6300e70833fc6daabdb7ed04b8506b2d
71838b1a20263e3cbdc7d5778e03b58e
887360c619b74ca0f182f304cabeddbb

You can add other font colors and background, but to simplify for now the colors are fixed.

src/Faker/Provider/Image.php
```diff
@@ -91,4 +91,52 @@ public static function image($dir = null, $width = 640, $height = 480, $category
...
+            throw new \InvalidArgumentException(sprintf('Cannot write to directory "%s"', $dir));
+        }
+
+        $name = md5(uniqid(empty($_SERVER['SERVER_ADDR']) ? '' : $_SERVER['SERVER_ADDR'], true));
+        $filename = $name .'.jpg';
+        $filepath = $dir . DIRECTORY_SEPARATOR . $filename;
+
+        $word = isset($word) ? $word : "{$width}x{$height}";
+
+        $im = imagecreatetruecolor($width, $height);
```
fzaninotto: This depends on gd being installed... I'm not a big fan of pushing the base requirements for Faker.

src/Faker/Provider/Image.php
```diff
@@ -91,4 +91,52 @@ public static function image($dir = null, $width = 640, $height = 480, $category
...
+        $black = imagecolorallocate($im, 102, 102, 102);
+        $white = imagecolorallocate($im, 224, 224, 224);
+
+        imagefilledrectangle($im, 0, 0, $width, $height, $white);
+
+        $font = __DIR__.'/../fonts/mplus-1c-medium.ttf';
```
fzaninotto: 600kB just to generate square images...

fzaninotto:
Thanks for your PR, but I'll pass on this one - I don't think it's a good idea to make Faker larger and less compatible just for an image generator, especially considering that there is already one available.

That being said, I recommend you to publish it as a standalone generator, that I could link to in Faker's README.

PR #1004, opened 2016-08-24
Use real Belgian postcodes instead of random number

toonevdb:
We needed real existing postcodes for our acceptance tests. Maybe useful for others too :)


fzaninotto:
Excellent, thanks!

PR #1011, opened 2016-09-02
Adding Lorem Formatter to ja_JP Provider

EvanLovely:
This adds Lorem support to Japanese. Let me know if you see any issues! This was grabbed from: http://generator.lorem-ipsum.info/_japanese


fzaninotto:

EvanLovely:
Could you please reconsider? We're using this for [prototyping](https://github.com/pattern-lab/plugin-php-faker) web sites in different languages and we're trying to get this in for Japanese as it demonstrate kanji characters that may not be in the typefaces were using. We want this for Russian as it demonstrates cyrillic characters for the same reason, and for German as it tends to have some of the longest words around. We have a web site prototype done in english and would like to demonstrate it in those 3 languages as we're working on a large site destined for a 31 language roll out. The reason that the `realText` provider doesn't work too well is that then the English version will use that too and it can distract from design review if the content is readable - reviewers tend to focus on that instead of layout ("the site looks great, but that's not the right copy...." And it's also nice to use the `firstName`, `sentence`, `month`, and other appropriate text placeholders. 

Thanks so much for your work on this and taking the time to reconsider these PRs. Additionally, if there's a different set of copy you'd like me to get in, I'd love to do it; we could even pull in the professional internationalization team to review these to ensure it's appropriate. Thanks!!

fzaninotto:
It makes sense if the words don't mean anything, like in the Latin Lorem formatter. Can you tell me where does the word list come from?

I'm willing to reconsider, but we'll have to find a way to add only one Lorem per alphabet - I don't want more code duplication across locales.

EvanLovely:
Thanks so much for being open to this!! Like I mentioned, we're rolling out a site with 31 languages, and the internationalization team has picked the 3 languages I've done PRs for that highlight the challenging different alphabets to focus on when designing websites that use custom fonts. We really feel that using Faker to demo a wide array of flexible content in our website design is very powerful and really helpful. Here's a breakdown of the languages chosen for PRs to this project:

## Japanese

Japanese (this PR) has the text pulled [from here](http://generator.lorem-ipsum.info/_japanese) and demonstrates [kanji characters](https://en.wikipedia.org/wiki/Kanji) which are [adopted from Chinese characters](https://en.wikipedia.org/wiki/Japanese_writing_system).

## Russian

Russian (PR #1012) has the text pulled [from here](http://generator.lorem-ipsum.info/_russian) and demonstrates the [cyrillic alphabet](https://en.wikipedia.org/wiki/Cyrillic_script).

> Cyrillic is the standard script for writing the following languages:

> - Slavic languages: Belarusian, Bulgarian, Macedonian, Russian, Rusyn, Serbo-Croatian (for Standard Serbian, Bosnian, and Montenegrin), Ukrainian
> - Non-Slavic languages: Abkhaz, Aleut (now mostly in church texts), Bashkir, Chuvash, Erzya, Kazakh, Kildin Sami, Komi, Kyrgyz, Mari, Moksha, Mongolian, Ossetic, Romani (some dialects), Sakha/Yakut, Tajik, Tatar, Tlingit (now only in church texts), Tuvan, Udmurt, Yuit (Siberian Yupik), and Yupik (in Alaska).
> The Cyrillic script has also been used for languages of Alaska, Slavic Europe (except for Western Slavic and some Southern Slavic), the Caucasus, Siberia, and the Russian Far East.

 

## German

German (PR #1013) has the text pulled [from here](http://loremipsumonline.com/en/languages/german-lorem-ipsum-generator?g=de) and even though it uses the Latin alphabet, it's used to demonstrate very long words that can throw off a website design. It also helps us ensure that web font characters are available for the [special characters](https://en.wikipedia.org/wiki/German_orthography#Special_characters): umlauts & ligatures. Additionally, it's the [third most widely used language on websites (after English and Russian)](https://en.wikipedia.org/wiki/German_language), so it's an important one to get right.

EvanLovely:
@fzaninotto Any thoughts on this? Thanks for looking!

fzaninotto:
I like it, except that it doesn't fit with the notion of locales in Faker. I don't want to duplicate the cyrillyc Lorem code in all locales using Cyrillic. 

So I see two options:

- add a `$defaultCharset = 'latin'` in Lorem, and include all the datasets for the other charsets in the base `Lorem.php` file. Every locale can then define a custom `Lorem.php` just changing the `$defaultCharset`.
- move the lorem data sets (including the latin one) to another directory, one file per character set, and load them on demand according to the `$defaultCharset`, in order to avoid forcing everyone to download all datasets when they only need one.

What do you think?

EvanLovely:
I like the first option more. How do you see that being done in the best way? Thanks!

fzaninotto:
I like the second option more, because it doesn't load useless data for every user :)

Would you like to give it a try in a new PR?

EvanLovely:
Sorry for falling of the conversation: we just bought our first house and moved in this last week!! Would love to get this rolling again! I'm not quite sure the best way to implement your suggestion, would you mind collaborating on a PR together? If you could start the foundation, I'd be happy to flesh it out further. Thanks!

fzaninotto:
After digging down a little, I like none of the possible implementations:

* Either we change the signature for all `Lorem` methods to include a `$charset` parameter, but this is VERY dirty. It will also force users to keep passing the `$charset` all the time
* Or we change all `Lorem` methods not to be static anymore, and we pass a `$charset` parameter to the constructor. This charset can be a classname (like `'\Faker\Provider\Charset\Latin'`), mater used to retrieve the word list. But this breaks BC heavily.

So I think the only good solution is to create a new provider called `DummyText`, where nothing is static anymore. 

Thoughts?

EvanLovely:
I'm down with that. Our need for this specific project has passed, but I'm sure that it'd be needed again.

PR #1012, opened 2016-09-02
Adding Lorem Formatter to ru_RU Provider

EvanLovely:
This adds Lorem support to Russian. Let me know if you see any issues! This was grabbed from: http://generator.lorem-ipsum.info/_russian


fzaninotto:

PR #1013, opened 2016-09-02
Adding Lorem Formatter to de_DE Provider

EvanLovely:
This adds Lorem support to German. Let me know if you see any issues! This was grabbed from: http://loremipsumonline.com/en/languages/german-lorem-ipsum-generator?g=de


fzaninotto:

PR #1014, opened 2016-09-02
German Bank Names

okj579:
I've added a list of German banks. Since in Germany it's _very_ common to belong to a co-operative or public bank, which are usually individually quite small (often owned by local government), I've included some of them as well as the largest private banks.


src/Faker/Provider/de_DE/Payment.php
```diff
@@ -16,4 +16,86 @@ public static function bankAccountNumber($prefix = '', $countryCode = 'DE', $len
...
+     * Sources:
+     * The 19 largest German banks by total assets
+     * @see https://de.wikipedia.org/wiki/Liste_der_größten_Banken_in_Deutschland
+     * The 20 largest co-operative banks by branch count
+     * @see https://de.wikipedia.org/wiki/Liste_der_Genossenschaftsbanken_in_Deutschland
+     * The 20 largest public savings banks by branch count
+     * @see https://de.wikipedia.org/wiki/Liste_der_Sparkassen_in_Deutschland
+     */
+    protected static $banks = array(
+        'Deutsche Bank',
```
fzaninotto: Could you please order them alphabetically, one line per first letter? This would ease future updates.

PR #1015, opened 2016-09-02
IBAN Formatters for New Locales

okj579:
I updated the new IBAN locales, so that `bankAccountNumber` is available.


fzaninotto:
Thanks!

PR #1018, opened 2016-09-06
fix mb_substr missing parameter error when generating japanese string with realText method

HejunweiCoder:
fix mb_substr missing parameter error when generating japanese string with realText method

### DIFF
--- a/src/Faker/Provider/ja_JP/Text.php
+++ b/src/Faker/Provider/ja_JP/Text.php
@@ -620,7 +620,7 @@ EOT;
     {
         // extract the last char of $text
         if (function_exists('mb_substr')) {
-            $last = mb_substr($text, 0, mb_strlen($text) - 1, null, 'UTF-8');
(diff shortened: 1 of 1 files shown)


HejunweiCoder:
may be this problem happened in php7

src/Faker/Provider/ja_JP/Text.php
```diff
@@ -620,7 +620,7 @@ protected static function appendEnd($text)
     {
         // extract the last char of $text
         if (function_exists('mb_substr')) {
-            $last = mb_substr($text, mb_strlen($text)-1, null, 'UTF-8');
+            $last = mb_substr($text, 0, mb_strlen($text) - 1, null, 'UTF-8');
```
fzaninotto: I don't understand. `mb_substr()`'s signature has only 4 arguments.

http://php.net/manual/en/function.mb-substr.php

fzaninotto:
Thanks!

HejunweiCoder:
You are welcome,I forget to remove that 'null' parameter...@fzaninotto

PR #1021, opened 2016-09-07
Adding Nationalized Citizens to DNI in Person.php

celisflen-bers:
Add Numbers of Nationalized Citizens in Person.php
Correct EOL Windows to Unix and adapting code a PSR-2


fzaninotto:
Thanks!

PR #1027, opened 2016-09-09
Fixed polish text - change '--' into '-'.

piotrooo:
Fixed '--' to ' - '. Before this `słońca--lśniące` was treated as one word.


fzaninotto:
Thanks!

PR #1031, opened 2016-09-12
Replaces main image funcationality with Unsplash.it to support SSL.

wuori:
Ran into an issue with consuming an API via iOS and learned that [iOS9 will enforce ATS])(http://www.techrepublic.com/article/wwdc-2016-apple-to-require-https-encryption-on-all-ios-apps-by-2017/), as such non-SSL images will not display. 

Since Lorempixel does not support SSL, switching provider to Unsplash.it, while not allowing categories (built-in fallback support for categories to rely on LoremPixel), UnSpash does offer some other features, such as directional cropping and blurring, both of which I've added to the functionality by appending parameters to the imageURL function. 

This fix may better be supported by waiting until new version and dropping LoremPixel (i.e. category support) altogether or choosing another SSL-enabled vendor that provides categories.

Just my $0.02!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -258,16 +258,14 @@ Each of the generator properties (like `name`, `address`, and `lorem`) are calle
 
 ### `Faker\Provider\Image`
 
-    // Image generation provided by LoremPixel (http://lorempixel.com/)
-    imageUrl($width = 640, $height = 480) // 'http://lorempixel.com/640/480/'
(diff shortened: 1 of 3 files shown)


wuori:
Doesn't matter! I want to re-do this a little more completely. Was in a hurry earlier to meet a deadline for something else. Closing and will-reopen when I have enough time.

PR #1033, opened 2016-09-14
Add additional functions and fix function call errors

travoltron:
Calling other functions besides 'sqrt' never worked in application, possibly because I'm on php7. I modified this file and added a couple additional biasing methods to the mix, namely 'square' and 'cube'.


fzaninotto:
The `$function` parameter isn't a function name, but a function value. You're breaking the API, si I can't accept that PR. If there is a bug on PHP7, I'd welcome a bugfix PR.

PR #1045, opened 2016-10-06
[Generator.php] mt_rand() changed in PHP 7.1

oittaa:
https://github.com/php/php-src/blob/php-7.1.0RC3/UPGRADING


fzaninotto:
Awesome, thanks!

PR #1054, opened 2016-10-10
[fi_FI] Improve phone number generator

Kurre:
With this patch, generated phone numbers are more realistic.
It's also easier to improve the formats in the future, if needed.


fzaninotto:
Thanks!

PR #1060, opened 2016-10-16
Add allowDuplicates option to randomElements()

vlakoff:
* "dice rolls" instead of "raffle tickets".
* can also be useful to pick more elements than the array length.

```php
$faker->randomElements(['foo', 'bar', 'baz', 'qux'], 3, true);

// possible result: 'foo', 'bar', 'foo'
```

### DIFF
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -161,18 +161,19 @@ class Base
     /**
      * Returns randomly ordered subsequence of $count elements from a provided array
      *
-     * @param  array            $array Array to take elements from. Defaults to a-f
-     * @param  integer          $count Number of elements to take.
+     * @param  array            $array           Array to take elements from. Defaults to a-f
+     * @param  integer          $count           Number of elements to take.
+     * @param  boolean          $allowDuplicates Allow elements to be picked several times. Defaults to false
      * @throws \LengthException When requesting more elements than provided
      *
      * @return array New array with $count elements from $array
      */
-    public static function randomElements(array $array = array('a', 'b', 'c'), $count = 1)
+    public static function randomElements(array $array = array('a', 'b', 'c'), $count = 1, $allowDuplicates = false)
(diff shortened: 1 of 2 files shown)


fzaninotto:
Isn't it the purpose of the `unique()` modifier?

vlakoff:
That's precisely the contrary. Currently `randomElements()` forces unique values, and I'm adding an option to disable this unicity constraint.

fzaninotto:
OK, now I get you. then you should change the default behavior to allow duplicates, and remove the `allowDuplicates` options (the user can already do that with `unique()`).

vlakoff:
But another confusion, these are two different things:
* Picking several times the same element from the input array (this PR)
* Having different selections returned by `randomElements()` (`unique` modifier)

Also, changing the default mode would definitely be a BC break.

fzaninotto:
I'm OK for a small BC break for this. And I don't get your first point.

vlakoff:
Maybe a bit of code would help to understand that `unique()` is unrelated:

```php
$data = ['foo', 'bar', 'baz', 'qux'];

// can give ['foo', 'bar']
// but not ['foo', 'foo']
$faker->randomElements($data, 2);

// assume this returned ['foo', 'bar']...
$faker->unique()->randomElements($data, 2);
// ... this can't return ['foo', 'bar'] too.
$faker->unique()->randomElements($data, 2);
```

vlakoff:
And I disagree about changing the default mode:
* There is no preferable mode, so why introducing an unneeded BC break?
* PHP's `array_rand()` also doesn't return duplicates.
* Avoiding duplicates shouldn't have many side effects, whereas getting several times the same element could be unexpected and break some apps. So it sounds fine having to explicitly enable duplicates.

fzaninotto:
OK, you've convinced me.

fzaninotto:
Thanks!

PR #1061, opened 2016-10-16
[nl_NL] Make person provider behave more realistically

curry684:
Removed some names that cannot occur or do not follow Dutch conventions, tweaked last name selector to far more accurately reflect the Dutch populace. Chance of selecting a 'noble' name reduced from 1:4 to 1:25, should be more like 1:25000 but it's good to generate them occasionally.


fzaninotto:
Thanks!

casperbakker:
Yes! Thank you @curry684!

PR #1065, opened 2016-10-20
Add Thai providers

tuwannu:


src/Faker/Provider/th_TH/Person.php
```diff
@@ -0,0 +1,166 @@
...
+        '{{firstNameFemale}} {{lastName}}',
+        '{{firstNameFemale}} {{lastName}}',
+        '{{firstNameFemale}} {{lastName}}',
+        '{{firstNameFemale}} {{lastName}}',
+        '{{titleFemale}} {{firstNameFemale}} {{lastName}}',
+    );
+
+    /**
+     * @var array Thai male first names.
+     * @link https://github.com/tuwannu/thai-names/blob/master/FIRST_NAMES_MALE.txt
```
fzaninotto: Sorry, that doesn't count as a trustworthy datasource

fzaninotto:
I suggest you remove the `Person` provider until you can reference a valid datasource.

tuwannu:
Right. Thanks for understanding. We don't really have an official datasource or anything close at all for person's names

tuwannu:
`Person` removed.

fzaninotto:
Thanks!

PR #1068, opened 2016-10-24
Ensure unique IDs in randomHtml

vlakoff:
In case of duplicate IDs the method fails.
Considering the small 5-digit-max range [0,99999] that's quite probable, and I [encountered it](https://travis-ci.org/fzaninotto/Faker/jobs/169501052#L282).
I did not increase the number of digits to be safe concerning BC.

A few notes about why I did the implementation as so:
* Doing `$this->unique()` or `$this->generator->unique()` would "pollute" (or be polluted by) the unique pool for `randomNumber` in other parts of the application.
* We want a fresh unique pool for each call to `randomHtml()`.

<br>ping @rudkjobing :)

### DIFF
--- a/src/Faker/Provider/HtmlLorem.php
+++ b/src/Faker/Provider/HtmlLorem.php
@@ -3,6 +3,7 @@
 namespace Faker\Provider;
 
 use Faker\Generator;
+use Faker\UniqueGenerator;
 
 class HtmlLorem extends Base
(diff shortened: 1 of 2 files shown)


fzaninotto:
Great, thanks!

rudkjobing:
Nice one. Thanks! :)

PR #1072, opened 2016-10-26
Reduced line lengths

jirehstudios:
PSR-2 recommends 80 characters per line or less.


fzaninotto:
No, thanks. The code is good enough.

jirehstudios:
There's a lot of room for improvement regarding PSR-2, but if you say so...

PR #1073, opened 2016-10-26
Updated spacing and line length

jirehstudios:
Reduced some lines to a length of less than 80 characters per PSR-2 standards and added spacing for readability between some lines.


fzaninotto:
No, thanks.

jirehstudios:
There's a lot of room for improvement regarding PSR-2, but if you say so...

PR #1077, opened 2016-10-26
Made unformatted and international formats configurable

smithandre:
Added ability to choose whether one wants a formatted or unformatted phone numbers, which can now also optionally be returned in international format.
- Formatted numbers could contain brackets, dashes and spaces, where unformatted numbers wont have any of those.
  - International format numbers will always start with '+27' instead of '0'
- Added unit tests to test new functionality


fzaninotto:
No thanks, this creates a too large difference between locales. I advise you to create a specialized provider in your onwn repo instead.

PR #1080, opened 2016-10-28
Removed legacy autoloader

sadovnik:
I think it's time to make it. :)


fzaninotto:
Why?

Ricknox:
I am sorry but there are people who use this autoloader. Not everybody can use Composer. Just my two cents.

fzaninotto:
Yep, and @sadovnik didn't elaborate, so... closing.

PR #1082, opened 2016-10-29
Adding in Master of Divinity Suffix

chrispecoraro:
This is important because it is four characters long.


fzaninotto:
Low value, seldom used, closing.

PR #1084, opened 2016-10-31
Created basic vehicle provider

smithandre:
* The idea with this provider is to generate vehicle related information like manufacturer, model, and vin  & engine numbers. But for now it's just manufacturers for UK, Germany & Japan
* Updated documentation


curry684:
Why all the obscure brands in there? Isn't fake data going to be a lot more useful if you're generating a car park full of Peugeot 308, Renault Twingo and Ford Focus?

smithandre:
@curry684 agreed. Just wanted to test the general concept. But I will put more effort into expanding and refining the lists

fzaninotto:
@smithandre Great start, but this is typically something that I prefer out of the core. You are the creator and you will maintain it, so it'd better live in another repository, as a standalone provider. 

I'll mention it in the README section when it's ready!

PR #1085, opened 2016-11-04
[ci skip]added .gitattributes

lichunqiang:


fzaninotto:
No, thanks.

lichunqiang:
Hi, can i get your thought to say no?

fzaninotto:
Can you first explain why you propose that change?

lichunqiang:
Ok. 

Now install via composer will download all the files in the vendor, but always like the `test` folder is no mean for the user.

This change can prevent those files to install to user's vendor.

In my case, our deploy system let us to write all the files which should be deploy online, that's a big work, because lot of depends...

lichunqiang:
for example

https://github.com/laravel/framework/blob/5.3/.gitattributes

lichunqiang:
https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes#Exporting-Your-Repository

PR #1087, opened 2016-11-07
[ja_JP]kana of Japanese name by gender.

itigoppo:
Changed person name for ja_JP
Updated README.md for ja_JP

I divided Japanese kana name by gender.


fzaninotto:
Thanks!

PR #1093, opened 2016-11-20
[ru_RU] Add jobTitle

2bj:


src/Faker/Provider/ru_RU/Company.php
```diff
@@ -25,6 +25,61 @@ class Company extends \Faker\Provider\Company
         'Текстиль'
     );
 
+    protected static $jobTitleFormat = array(
```
fzaninotto: what's the datasource? Is it MIT-compatible?

2bj: List compiled myself from different sources

fzaninotto: Then you must list the datasources and they must all be MIT compatible

fzaninotto:
No news, closing.

PR #1094, opened 2016-11-27
Change postal code format for ko_KR

coozplz:
Korean postal codes have been changed from the old xxx-xxx format to xxxxx on August 1, 2015


fzaninotto:
Thanks!

PR #1097, opened 2016-11-28
Add northern german cities Flensburg and Kiel

Konafets:


fzaninotto:
Low value PR. Please find an open-source reference of major German cities and replace the entire list (with a comment pointing to the data source). Otherwise, anyone can decide to add or remove a city because they want to.

Konafets:
Of course. Please look at #1102.

PR #1103, opened 2016-11-30
Update MasterCard BIN Range

andysnell:
MasterCard has announced that they are increasing the number of potentially valid credit card numbers by adding new BINs (bank identification numbers, i.e., the first six digits of the credit card number). Effective January 2017, banks may begin to issue MasterCard credit cards starting with "2" instead of the usual "5".   This major change was announced last year, and MasterCard is expecting acceptance compliance by June 2017.  Including the new BIN range in Faker should help developers make this transition. 

This PR updates the valid MasterCard card number ranges, adding entries to support 16-digit card numbers starting with 2221-2720.  Additionally, it updates the regular expression used by the test function in PaymentTest.php to match on these ranges.

Additional information on the BIN range update can be found at: 
https://www.mastercard.us/en-us/issuers/get-support/2-series-bin-expansion.html


fzaninotto:
Thanks, could you add a comment containing the link you mentioned in the source, so that future developers know where it comes from?

andysnell:
Added as requested.

fzaninotto:
Excellent, thanks!

PR #1107, opened 2016-12-01
Add lorem files, modify data, change the format

AmiBang:
Why before inspections department through?Submit this time?


AmiBang:
Compiled automatically checks strictly, really helpless

fzaninotto:
Lorem is never localized ; don't add one or two elements in a data set. Instead, replace the entire dataset, but reference an external (and open-source) dataset.

AmiBang:
Have changed, thank you for your prompt.Really want to do some contribution!

AmiBang:
Thank you, very clear tell my mistake, I will carefully read contributors guide. I will also continue to contribute.

PR #1110, opened 2016-12-02
Lower column name in name guesser later

guidocella:
Moved the toLower() call on the column name in the Name guesser after the regular expressions that check for upper case characters.


fzaninotto:
Zero interest, closing.

PR #1111, opened 2016-12-02
Added a generator so its possible to get a associative out from a associative array 

lsv:
As my model required a associative array, like this ```['a' => 'A', 'b' => 'B']``` out from a larger array, I added a generator for this.

Example

```php
$input = array('a' => 'A', 'b' => 'B', 'c' => 'C', 'd' => 'D', 'e' => 'E');
$output = $generator->randomKeys($input, 2);
```

Output is now a associative array

```
Array
(
    [b] => B
    [a] => A
)
```

I thougt about the name ```associativeElements``` instead of ```randomKeys``` but I sticked with ```randomKeys```


fzaninotto:
No, thanks. It's so easy to do an `array_keys` then use Faker to find a random element.

PR #1112, opened 2016-12-02
Improve Internet::transliterate performance

dunglas:
This patch improves the performance of the `Internet::transliterate` method by not using the transliterator if the current string is already in ASCII.

It improves the speed of Alice's benchmark of 37%.

/cc @theofidry


theofidry:
Link to the blackfire profile: https://blackfire.io/profiles/compare/d27f37a2-acf2-4da0-852b-ef307872f084/graph

dunglas:
(travis error not related)

fzaninotto:
Awesome, thanks!

PR #1114, opened 2016-12-10
Created physics class to generate a temperature degree number

tperrelli:
added Physics class to provider temperature generator (C - celsius | F - Fahrenreit | K - Kelvin)

### DIFF
--- a/src/Faker/Factory.php
+++ b/src/Faker/Factory.php
@@ -6,7 +6,7 @@ class Factory
 {
     const DEFAULT_LOCALE = 'en_US';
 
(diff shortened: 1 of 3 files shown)


tuwannu:
This is an interesting idea! Although I see some issues in the PR that I would love if they are fixed/implemented.

- Max/min values are fixed to -300 to 300 even though there are arguments for them
-`temperature()` returns `10°C` string, which is good for displaying, but not ideal for further temperature calculations. I would consider `humanTemperature()` instead.

If this PR is accepted, perhaps it could be extended to do e.g. `Physics::temperatureOf('water', 'liquid', 'C')` or `Physics::celsiusOf('water', 'liquid')` to give value between 0-100C?

tperrelli:
Yeah! I also thought something about calculations, but the main ideai is only for generate some random numbers. I agree with you when you suggest humanTemperature() 👍 
I don't know if converting values is within the fzaninotto proposal..

fzaninotto:
Nice idea! I suggest you make it a standalone provider package and publish it on composer. I'll be happy to reference it in the README.

tperrelli:
@fzaninotto  @tuwannu 
https://github.com/tamperdata/exiges

PR #1120, opened 2016-12-17
Link Eloquent Populator

guidocella:
Hi. I created an adapter/Populator for Laravel's Eloquent ORM, but I released it as a [separate package](https://github.com/guidocella/eloquent-populator) because I ended up adding lots of stuff. So I added a link to it in the readme if it's ok. I don't know if I should add Eloquent to the list of supported ORMs.


fzaninotto:
Great, thanks! And feel free to add Eloquent to the list of supported ORMs.

PR #1121, opened 2016-12-20
Added missing $state property

enricobono:
This file misses the $state property, this is just a quick fix of the phpdoc block


fzaninotto:
Duplicate of #792

PR #1122, opened 2016-12-20
Image placeholder

biserantonov:
Generate image with solid background and text. Inspired by https://dummyimage.com/


fzaninotto:
This has already been done, see #712. Feel free to publish it as a standalone provider.

PR #1132, opened 2017-01-04
Updating ninth digit on whole country

gpressutto5:
Finally, every cellphone number on Brazil has 9 digits. We don't need to check its area anymore.


fzaninotto:
Thanks!

PR #1137, opened 2017-01-25
New Male and Female names

Purus:


fzaninotto:
Low value PR. Which names should be there and which ones should not be there? If you feel that the list isn't good enough, please find an external datasource with good statistical accuracy, and an open-source license, then replace the entire list.

PR #1138, opened 2017-01-25
Indian city names for en_IN Locale

Purus:


fzaninotto:
Low value PR - see my comment in your other PR.

PR #1141, opened 2017-02-03
remove bad phone number from ru_RU PhoneNumber

shatilov:


fzaninotto:
Can you link to a public resource explaining the formats of the legal phone numbers in the phpDoc?

pwsdotru:
May be this: https://en.wikipedia.org/wiki/Telephone_numbers_in_Russia

fzaninotto:
...in the phpDoc, so future developers understand your code

pimjansen:
@shatilov bump

PR #1144, opened 2017-02-11
Separated male and female names for sr_RS locale.

bogdanpet:
I was free to separate those names in sr_RS locale, it is 99% correct, even though I am Serb some of the names are very rare and dead for centuries so there might be a mistake or two


fzaninotto:
I don't understand, male and female first name were already separated. What did you do exactly?

bogdanpet:
if you see Person.php file in sr_rs locale, there is line
`// below are unknown gender firstnames`
and a bunch of both gender names appearing in both male and female first names, I just classified them. See the changes.

fzaninotto:
Thanks!

PR #1151, opened 2017-02-27
Add prefixes for nl_NL

hyperized:
Based on: https://www.vernoeming.nl/alle-333-voorvoegsels-tussenvoegsels-in-nederlandse-achternamen

Fixed a few consistency issues along the way with handling single quotes.

Added the `prefix` method.

Due to the nature `$commonDutchLastNames` and `$dutchLastNames` are set up (already including the prefixes) I did not want to interfere with the current formats, but simply add additional functionality.

### DIFF
--- a/src/Faker/Provider/nl_NL/Person.php
+++ b/src/Faker/Provider/nl_NL/Person.php
@@ -30,6 +30,22 @@ class Person extends \Faker\Provider\Person
         'BA', 'Bsc', 'LLB', 'LLM', 'MA', 'Msc', 'MPhil', 'D', 'PhD', 'AD', 'B', 'M'
     );
 
+    protected static $prefix = array("'s", "'t", 'a', 'aan', "aan 't", 'aan de', 'aan den', 'aan der', 'aan het',
+        "aan t", 'af', 'al', 'am', 'am de', 'auf', 'auf dem', 'auf den', 'auf der', 'auf ter', 'aus', "aus 'm",
+        'aus dem', 'aus den', 'aus der', 'aus m', 'ben', 'bij', "bij 't", 'bij de', 'bij den', 'bij het', 'bij t',
+        'bin', 'boven d', "boven d'", 'd', "d'", 'da', 'dal', 'dal’', 'dalla', 'das', 'de', 'de die', 'de die le',
+        'de l', 'de l’', 'de la', 'de las', 'de le', 'de van der', 'deca', 'degli', 'dei', 'del', 'della', 'den',
+        'der', 'des', 'di', 'die le', 'do', 'don', 'dos', 'du', 'el', 'het', 'i', 'im', 'in', "in 't", 'in de', 'in den',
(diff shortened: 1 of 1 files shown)


curry684:
Why have them all with and without capitalization? That makes using them in a clean consistent UI highly unpredictable...?

hyperized:
Hi @curry684, I took the data set as-is.

Is there anything specific you want to see changed before this can be merged?

As for the reason why I listed them both is that I can only assume it was listed in this particular matter because they are actually used as such.

Of course, it's possible to reduce the set and introduce a 'capitalize first' function, but that would not work well for `'S` vs `s'` forms (or any other quoted prefix for that matter.

curry684:
The problem is that if you have both the code might end up generating names like "Joran Van der Sloot" which is inconsistent with Dutch conventions. In general something that is not by definition capitalized, like eigennamen, should be generated without caps. If it is to be used in a sentence the invoker can call `ucfirst` himself.

hyperized:
@curry684 Ok, is it mergeable if I remove the capitalisation duplicates?

I was not aware that probability is (or should be?) a factor in faking prefixes and/or names.

Are you suggesting I should overhaul the way the names are rendered and separate the prefixes from the lastNames altogether, or should I just close this PR?

It's hard to find out what would make this mergeable from your response.

curry684:
I'm just a contributor, I have no say in whether it's mergeable - I'm just sharing my opinion as a consumer of the library  :)

fzaninotto:
I'm not able to validate/invalidate this capitalization question. But @curry684's comments make sense!

hyperized:
The dataset provided uses both variants because they are used as such.

Both: 'Pieter De la Bateau' and 'Koen van der Broek' are perfectly valid Dutch names, but let's not overstep the pull request here, I just want be able to get the prefix, not the full name structure, that would require a rewrite.

curry684:
Reference the authority on Dutch language: https://onzetaal.nl/taaladvies/hoofdletters-in-namen-nynke-van-der-sluis-nynke-van-der-sluis

We are actually both right - in Dutch Dutch the prefix is always decapitalized, so the 'Pieter De la Bateau' in your example would always be modified to 'Pieter de la Bateau' in the Netherlands. In Flemish Dutch however you are correct that the official citizen registry is leading, and 'Pieter De la Bateau' would be written as such if the surname were fully registered as 'De la Bateau'.

However the pull request is on the `nl_NL` locale, which is Dutch Dutch, and not Flemish Dutch which is `nl_BE`. Therefore in `nl_NL` the capitalized variants should not be present.

hyperized:
Rebased to remove capitalised duplicates.

hyperized:
Hi @fzaninotto, is this PR something you could include :)?

fzaninotto:
As long as you guys both agree, I'm fine with it ;)

fzaninotto:
Thanks!

PR #1163, opened 2017-03-16
Refactor text method to remove duplication

ankitpokhrel:


fzaninotto:
Why do you create a new method? for me, your `generate()` should be inside `text()`.

ankitpokhrel:
Habbit 😄, will change!

ankitpokhrel:
@fzaninotto updated.

fzaninotto:
Thanks!

PR #1168, opened 2017-03-22
Rewrote deprecated `each()` usage

hboomsma:
See https://wiki.php.net/rfc/deprecations_php_7_2#each
Also allowed using PHPUnit version 5 because version 4
also still uses the deprecated `each()` function. This
way the project will continue to build on future PHP
versions without sacreficing backwards compatibility to
PHP 5.3-5.5 versions. (Would be needed for allowing to
install PHPunit 6).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ php:
     - 5.6
     - 7.0
     - 7.1
+    - nightly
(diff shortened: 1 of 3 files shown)


fzaninotto:
Thanks!

PR #1170, opened 2017-03-26
Typos and doc block fixes

ankitpokhrel:
This PR removes unnecessary vars, fixes typos and adds missing doc block params


fzaninotto:
Thanks

PR #1172, opened 2017-03-26
Turkish Provider for Text.

tkorkunckaya:
Turkish translation of Alice in Wonderland (by Lewis Caroll) by Kismet Burian is used for the text provider.


fzaninotto:
No, please use a native Turkish novel instead of a translation of a foreign text.

PR #1173, opened 2017-03-27
Added hint on state option

bretto36:
Added hint for autocompletion to match documentation


fzaninotto:
No, because we only add phpDoc for formatters existing in the Base providers. `state` exists in some locales, but not all.

PR #1174, opened 2017-03-27
create Text.php for zh_CN

sandysong:
create Text.php for zh_CN, since 'zh_CN' and 'zh_TW' are similar to each other, so the code is just copy from zh_TW


fzaninotto:
Copying code is bad. Why don't you use inheritance?

seekcx:
Simplified and traditional, although the same content, but the characters are not exactly the same, so can not simply use inheritance to solve the problem.

I am also zh-cn no text support and distress, see the commit, should be a solution, hoping to merge.

fzaninotto:
I don't understand your answer, can you elaborate?

fzaninotto:
No feedback for a month, closing

PR #1175, opened 2017-03-27
Added Emoji to Miscellaneous

thomasfdm:
Added the ability to generate random Emoji.

These are Unicode characters from the block U+1F600 to U+1F637. 

Since many messages and text posts online now make use of Emoji (especially if send from a mobile device), I think some may find it useful.

### DIFF
--- a/src/Faker/Provider/Miscellaneous.php
+++ b/src/Faker/Provider/Miscellaneous.php
@@ -4,6 +4,28 @@ namespace Faker\Provider;
 
 class Miscellaneous extends Base
 {
+
+    /**
+     * @link https://en.wikipedia.org/wiki/Emoji#Unicode_blocks
+     * On date of 2017-03-26
+     *
+     * U+1F600 - U+1F637
+     */
+    protected static $emoji = array(
+        "\u{1F600}", "\u{1F601}", "\u{1F602}", "\u{1F603}", "\u{1F604}",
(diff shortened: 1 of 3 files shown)


thomasfdm:
Didn't change any code from 2fd6fe9f19ab9fbb262c9a03f50948767b097824 .

Travis says the build failed because of part of the DataTime tests, should I make an issue for this?

src/Faker/Provider/Miscellaneous.php
```diff
@@ -273,10 +300,21 @@ public static function languageCode()
...
     }
+
+    /**
+     * Returns an encoded Unicode Character between U+1F600 and U+1F637.
+     *
+     * @link https://en.wikipedia.org/wiki/Emoji#Unicode_blocks
+     */
+    public static function emoji()
+    {
+        return json_decode('"' . static::randomElement(static::$emoji) . '"');
```
fzaninotto: why use `json_decode` for that? Why not store unicode directly in the `static::$emoji` source? That would be faster.

thomasfdm: I did that in my original version 62b87b5de007df5dac40594df9258cbf397c4bbc , however the "\u{xxxx}" notation wasn't added to PHP until 7.0, so it failed the 5.x Travis builds.

After research this was apparently the best method to store unicode characters in PHP pre-7.0

fzaninotto:
Then we'll merge it as is. Thank you!

PR #1183, opened 2017-04-10
Fix: Add class-level method annotations for DateTime provider

localheinz:
This PR

* [x] adds missing class-level `@method` annotations from `DateTime` provider


fzaninotto:
Thanks!

localheinz:
Thank you, @fzaninotto!

PR #1185, opened 2017-04-11
Add PhoneNumberFormat for ja_JP.

itigoppo:


itigoppo:
Contains a format that causes an error in input check.
The telephone number in Japan always starts with "0".

fzaninotto:
Thanks!

PR #1186, opened 2017-04-11
use secure https

jpuck:
Because https://lorempixel.com/ uses a valid SSL certificate, this will stop those pesky Mixed Content warnings.

In the good spirit of security, even while in development.

https://youtu.be/PdnpNJZVUE0


fzaninotto:
Thanks!

PR #1189, opened 2017-04-13
Added new biasedElementBetween

axettone:
I needed such a method because I had to seed a database with entries having a certain flag equal to "false" only in a small %.
So you can achieve this with:

`$entry->field = $faker->biasedElementBetween([true,false],[0.9,0.1]);`

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -299,6 +299,9 @@ Each of the generator properties (like `name`, `address`, and `lorem`) are calle
     // get a random number between 10 and 20,
     // with more chances to be close to 20
     biasedNumberBetween($min = 10, $max = 20, $function = 'sqrt')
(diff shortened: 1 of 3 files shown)


axettone:
I renamed the method biasedElementIn (it's more appropriate)

fzaninotto:
Thanks, but I'll pass on this one. Otherwise, what prevents us from re-implementing every formatter in a biaised way? 

I think a better approach would be to inject a custom randomizer in the base object. That would help biaising all formatters at will.

axettone:
Yes, I agree with you. Do you mean we can add an extra-parameter to any method of the Base class so to trigger this thing?

fzaninotto:
No, I was rather thinking about a constructor parameter

PR #1199, opened 2017-04-23
added missing address hints

golubkovden:


fzaninotto:
No, these only exist in some locales, not in the base provider.

Issue #1171, opened 2017-03-26
Localization for Nigeria should be en_NG


ujiriukiri:
Nigerian is an English speaking country and its locale is internationally recognized as "en_NG".

raphaeldealmeida:
This is a valid issue, but it breaks the compatibility. I dont know if the maintainer will merge it now or in a major release

fzaninotto:
Fixed by #1205

PR #1205, opened 2017-05-04
Fix ng_NG to en_NG

raphaeldealmeida:
Localization for Nigeria should be en_NG #1171 Renamed classes and created a readme entry


fzaninotto:
Thanks!

Issue #1191, opened 2017-04-17
middleName consistency


JustBlackBird:
At the moment different providers expose middle names in different ways.

For example, person provider for `ru_RU` locale exposes middle names with two methods: [`middleNameMale`](https://github.com/fzaninotto/Faker/blob/master/src/Faker/Provider/ru_RU/Person.php#L117) and [`middleNameFemale`](https://github.com/fzaninotto/Faker/blob/master/src/Faker/Provider/ru_RU/Person.php#L131). At the same time, person provider for `vi_VN` locale exposes middle names as a single [`middleName`](https://github.com/fzaninotto/Faker/blob/master/src/Faker/Provider/vi_VN/Person.php#L164) method.

I suggest to use only one way of exposing middle names. To be compatible with `lastName` and `firstName` it should be done as in `vi_VN` locale with a single method `middleName` with `$gender` flag.

If the maintainer is interested in it I could provide a PR to fix this.

fzaninotto:
I don't mind inconsistencies in middleNames.

JustBlackBird:
Consider an use case when the lib is used to generate test data for multilingual system. If I use both `ru_RU` and `vi_VN` locales for testing I have to **vary method name** to get middle name. For example:

```php
$faker = \Faker\Factory::create('ru_RU');
$faker->middleNameMale(); // There is no "middleName" method :(
```
but
```php
$faker = \Faker\Factory::create('ru_RU');
$faker->middleName(\Faker\Provider::GENDER_MALE);
```

PR #1209, opened 2017-05-10
Add Person::middleName for ru_RU provider

JustBlackBird:
This PR adds a method for middle name generation compatible (by signature) with `Person::firstName` and `Person::lastName`.

For details about the original problem see #1191


pwsdotru:
Code not worked in PHP 5.3.  Should use array( ) not brackets [ and ] for define array.

JustBlackBird:
@pwsdotru fixed

fzaninotto:
thanks!

PR #1214, opened 2017-05-17
Fix duplication in uk_UA male first names

fre5h:
There two Ukrainian names `Євген` in the list of male first names. One of them should be changed to `Євгеній`, because it is a different name. So after this change list will contain only unique values.


fzaninotto:
Thanks!

pimjansen:
Thanks for the contribution!

PR #1216, opened 2017-05-18
Update country names in Ukrainian language

fre5h:


fzaninotto:
I have no way to validate these changes. You must provide an external and reliable datasource as reference.

fre5h:
@fzaninotto 
About first and second changes.
Last year in Ukraine where renamed a lot of cities according to the decommunization. Here is a link to Wikipedia with a list of all changes https://en.wikipedia.org/wiki/List_of_Ukrainian_toponyms_that_were_changed_as_part_of_decommunization_in_2016
There you can find:

| city |	Dnipropetrovsk |	Dnipropetrovsk city council |	Dnipro |	ПВРУ 1375-VIII |
| city	| Kirovohrad	| Kirovohrad city council |	Kropyvnytskyi	| ПВРУ 1468-VIII |

About last change.
Khmelnytskyi is a city where I live. Very often people forget to add ending `yi` to the end of the city name, like Khmelnytsk. But it is wrong, it is not an official name. Here is also a link to Wikipedia https://en.wikipedia.org/wiki/Khmelnytskyi,_Ukraine

fzaninotto:
Can you find a public list of Ukrainian cities of good quality your could use as a reference?

fre5h:
A set of laws on the administrative structure of Ukraine. This is the official website of the Parliament of Ukraine. But unfortunately this page is only in Ukrainian language.
http://static.rada.gov.ua/zakon/new/NEWSAIT/ADM/zmist.html

fzaninotto:
Then please use that list instead of the current one, and add the url to the phpDoc for future reference.

pimjansen:
@fre5h bump

fre5h:
@pimjansen fixed

src/Faker/Provider/uk_UA/Address.php
```diff
@@ -268,7 +278,6 @@ class Address extends \Faker\Provider\Address
         'Урицького',
         'П. Орлика',
         'Хрещатик',
-        'Артема',
```
fzaninotto: I can't find the list of streets in the URL given in phpDoc. So I can't check that the list is up to date, or that you've removed that name for a good reason. Please explain why you removed that line or revert.

src/Faker/Provider/uk_UA/Address.php
```diff
@@ -11,6 +11,10 @@ class Address extends \Faker\Provider\Address
     );
     protected static $buildingNumber = array('##');
     protected static $postcode = array('#####');
+
+    /**
+     * @see https://uk.wikipedia.org/wiki/%D0%A1%D0%BF%D0%B8%D1%81%D0%BE%D0%BA_%D0%BA%D1%80%D0%B0%D1%97%D0%BD_%D1%81%D0%B2%D1%96%D1%82%D1%83
```
fzaninotto: the content of this page and the content of the country list in this class don't match. The idea of having a datasource is to have an external reference where other people can look for changes and update the dataset in the future.

PR #1217, opened 2017-05-24
Fix spelling errors

driade:
Fixing a lot of spelling errors. Removing just two or three names/surnames which are not Spanish (ex, Mohamed). Would propose, if accepted, removing those names/surnames which came for Latin America, specially those which are just versions of their Spanish originals.


driade:
It seems Travis failed because other reasons...

fzaninotto:
The names come from a source (mentioned in the phpDoc block, http://www.ine.es/daco/daco42/nombyapel/nombyapel.htm), so the fix should be done there first.

driade:
Sorry, http://www.ine.es/daco/daco42/nombyapel/nombyapel.htm doesn't work.

fzaninotto:
Then you need to find another source...

You see, for Faker we must rely on external (and trusted) sources. Otherwise, nothing prevents a given developer from opening a PR to add their name, and this adds no value to Faker.

driade:
I fully understand. The source was http://www.ine.es/dyngs/INEbase/es/operacion.htm?c=Estadistica_C&cid=1254736177009&menu=resultados&idp=1254734710990 though as you may see they offer the data uppercase.

At this moment, I can tell you as a Spanish speaker, looks horrible. We take care a lot about accents.

I'll try to find a source where you can rely on.

fzaninotto:
OK, so if the source is uppercase, I agree that we should add accents here in Faker. Can you update the source in the doc comment?

driade:
Added a comment. I've tried to find any other reliable source for the project, but honestly the three ones I've checked, from different Spanish authorities, are uppercase or lack (!!) of accent marks.

fzaninotto:
Thanks!

PR #1218, opened 2017-05-24
Fix spelling errors

driade:


driade:
Would propose to check if accepted the file. It contains text in other languages, such as Catalonian. As this is es_ES I don't know this texts should be removed to use just Spanish Castilian.

fzaninotto:
I think it's good as is, since street names can be in Catalonian in Catalonia, which is in Spain.

driade:
Not completely agree, sorry. The docs state that the argument passed is the locale, not the country. The Catalonian language has its own locale.

driade:
Also, please, check the changes. I didn't remove any text in Catalonian, just spelling errors.

PR #1224, opened 2017-06-01
Added Animal names provider

edmandiesamonte:
Added Animal Names provider using common animal names
Modified lorempixel URL to avoid ignored test.


fzaninotto:
Nice! But I'm not in favor of adding that to the Faker core. I suggest you release it as a standalone package instead, which I'll link to in the README section for Faker.

PR #1226, opened 2017-06-04
Update Person.php

tamjidmurtoza:
Add some first name and last name.


fzaninotto:
We need much more entropy than that, so a much larger corpus. Please find a MIT-compatible datasource with at least 50 first names for each genre.

Issue #1165, opened 2017-03-18
dateTimeThisMonth generates last month date


dev-seahouse:
faker version: 1.6.0
code
```php
$d = $faker->dateTimeThisMonth();
dd($d);
```
It is march now  
try it a few  times and i got:
```
DateTime {#623
  +"date": "2017-02-24 01:04:43.000000"
  +"timezone_type": 3
  +"timezone": "UTC"
}
```

curry684:
The name is slightly misleading but all `dateTimeThisX` methods actually work as Last X, not This X. So they return a date time within the last week, month, year etc., not within the current week, month, year.

marjetika:
@curry684 - so this issue should be labeled as "it's not a bug it's a feature"

curry684:
More "it won't be changed because of backwards compatibility", aka "better reliably broken than unreliably fixed"   😉

icanhazstring:
Old issue but still valid. This naming is misleading and should be changed in the future.

Maybe we should add proper named methods and add deprecation warnings to the old ones.

```
$faker->dateTimeCurrentX
$faker->dateTImeLastX // current is $faker->dateTimeThisX
```

pimjansen:
@icanhazstring correct, can we make that happen?

curry684:
Before we go into grammar nazi hell again on this - what's the official native English speaker stance on this?

Technically I think the purists maintain that, on August 15th, that "this month" refers to all of August, and "last month" to all of July. So let's doublecheck that before we apply the wrong fix.

pimjansen:
@curry684 agreed, this meaning the current and last the previous.

The current date now is now the 27th of August 2019

dateTimeCurrentMonth (creates a random date between 01-08-2019 and 31-08-2019)
dateTimeCurrentYear (creates a random date between 01-01-2019 and 31-12-2019)
dateTimeCurrentDay (creates random timeslot at 27-08-2019)

For the last one i have no clue if it was implemented but this is what i expect. Overall the methods are not changing however we should create a new method and deprecate the other one with a good explanation on why it is done

icanhazstring:
If not. I am fine raising one.

curry684:
Go for it, I barely have time to respond to issues, I've been coding less than 4 hours per week for months now   😾

icanhazstring:
Ok, so just to wrap this up.

All comments assuming the current date is `2019-08-28`
We *create* the following methods:
[...]

Issue #1078, opened 2016-10-27
Invalid date/time values generated (DST)


telkins:
From time to time we get errors when attempting to insert Faker-generated data into our MySQL database.  We're specifically using `dateTimeThisYear()`.

Here's the error: `SQLSTATE[22007]: Invalid datetime format: 1292 Incorrect datetime value: '2016-03-27 02:08:58'`

As you can see, at first glance, the date/time looks valid, but it's actually between 2 and 3 AM on the early Sunday morning where time leaps forward from 2 to 3 AM.  So, any time between 2 and 3 AM doesn't actually exist.

The column in question is a typical MySQL `datetime` field.

We're running Faker 1.6.0, PHP 7.0.8, Laravel 5.3.19, and MySQL 5.7.13-0ubuntu0.16.04.2, not that it really matters.

I'm not 100% certain that this is a Faker problem because, as I understand it, there are no time changes in UTC, the MySQL server in question is set to use the system timezone, and the system timezone appears to be CEST.  Although, as I mentioned, we're using `dateTimeThisYear()`, so there's no way (as I understand it) to specify the timezone.  Whatever the case, if someone knows better, please let me know.

curry684:
This is not Faker's issue as I see it. Faker is just generating random data within a rigid set of rules, so in the case of a DateTime that means it follows rules like `0 <= hour < 24`. It's completely out of scope for Faker to consider the validity of that time within any more specific situation than Coordinated Time (UTC), which it automatically is according to those constraints. With good reason, as timezone specific handling is a [completely ridiculous field of expertise in IT](http://stackoverflow.com/a/6841479/1729885) with tens of thousands of exceptions in many timezones. Catering to all of them isn't the job of a simple PHP faker library, especially given that there are [wonderful standard libraries](https://packages.debian.org/stretch/tzdata) that contain 2.8MB of data describing [every known exception ever in every timezone](https://packages.debian.org/stretch/all/tzdata/filelist).

So your problem is not in Faker, but in configuring your MySQL not to use UTC. Which is actually not that bad a choice by definition, assuming that MySQL is only ever going to get time data prevalidated to those constraints you imposed on it willingly. That's where this all starts to crash - the point where you the consumer of a generic fake data generator are blindly submitting unvalidated data into your database. You would have much more frequent problems if you didn't happen to live in CE(S)T which is one of the most stable timezones over the past century.

In the end the solution is pretty simple - accept that Faker gives you UTC and convert it to the right zone:
```
[...]

Issue #945, opened 2016-06-24
Timezone conversion issue


thealonekhan:
Hi, i have an issue while converting a giving date or time for a specific **timezone**, using faker **DateTime**.
The issue is: it always convert the date for a given timezone, exact one hour prior.
This is my code:

`$date = new \DateTime($start_time, new \DateTimeZone('GMT')); //start_time will be db pulled time $date->setTimezone(new \DateTimeZone('EDT'));
 $start_time = $date->format('H:i');` 

Now the $start_time will be one hour earlier than given timezone **Eastern Daylight Time(EDT)**.
Any suggestion?

fzaninotto:
fixed by #1231

PR #1231, opened 2017-06-14
Add fix for default time zone

jpamental:
There is a PHP error we see that indicates problems when no default time zone is specified. This adds UTC as the default. This may well address a number of dateTime related issues like #945 #1165 #1078 and possibly others.


fzaninotto:
Setting a default timezone is the #1 thing every developer should do on a new environment. It even raises a warning on newer versions. I don't want to go the route of dealing with noob mistakes.

PR #1238, opened 2017-06-21
Add the option to generate HTML fragments and the body of an HTML document

WillGibson:
Takes work done by @rudkjobing to allow for generating HTML without it needing to be a complete HTML document.

Then issue of it sometimes creating HTML with no human readable text content was addressed and builds got passing again on the various versions of PHP.


WillGibson:
@fzaninotto I appreciate that the commit history has gotten itself a bit loose in here. I'll have a go at rebasing this on your master maybe to see if I can straighten that up.

src/Faker/Provider/HtmlLorem.php
```diff
@@ -129,25 +181,56 @@ private function addRandomAttribute(\DOMElement $node)
...
+        }
+    }
+
+    /**
+     * @param \DOMElement $node
+     * @return void
+     */
+    private function addUniqueIdAttribute(\DOMElement $node)
+    {
+        for ($x = 1; $x <= 10000; $x++) {
```
fzaninotto: Better use a `while` than a `for` that you'll `break`. Oh, and you're free to use recursion by the way.

WillGibson:
I just stumbled across this PR and it's so old I can't even remember anything about it's context, so I'm closing it.

clarkewing:
Hey @WillGibson, any chance of reviving this?
I've got quite a few use cases for generating HTML fragments, and my current implementation using regex to extract everything between the `<body>` tags is... iffy.

WillGibson:
I'm afraid I don't have time to look at it and, if I'm being honest, aside from it being do long ago I can't remember it, I'm not using this library for anything at the moment, or even PHP just now.

PR #1240, opened 2017-06-22
Update Faker/src/Faker/Provider/de_DE/PhoneNumber.php

Windowsfreak:
Increase frequency of mobile numbers and commonly used syntax such as slashes and hyphens


fzaninotto:
This will break existing tests using seeding for an added value that I don't see very high. Thanks, but I won't merge this one.

PR #1242, opened 2017-06-26
Use static instead of self

Stichoza:
Because `$textStartsWithUppercase` was always set to false no matter what you had set in localized class.


fzaninotto:
Thanks!

PR #1247, opened 2017-06-29
Use remote fopen() to retrieve the file if curl fails

Charliiee:
Use method `copy` to retrieve the file from lorempixel.com when `$success` is false even when function curl exists as a workaround to the problem of "_Network unreachable_" when trying to get contents via curl.


fzaninotto:
No, thanks. There are a dozen ways to request a distant resource in PHP, I don't want to deal with more than one. It's not Faker's job.

PR #1248, opened 2017-07-03
Update Composer File

vinkla:
Update the Composer file: bumped the branch alias and added consistent version constraints.

This way way we can test the new version by requiring the `^1.7` development version.


fzaninotto:
Thanks!

PR #1251, opened 2017-07-07
Update ISO 4217 active codes

eidng8:
Contains only active codes.


fzaninotto:
Thanks!

eidng8:
yw

btw, when will it be available though packagist?

fzaninotto:
When it's released!

PR #1252, opened 2017-07-09
init address test for Vietnam locale

mnking:


fzaninotto:
Sorry, these tests have too low value.

PR #1258, opened 2017-07-20
rectify ISO 4217 codes

eidng8:
removed codes that are not "money", such as special drawing rights, fund codes, gold units, etc.


fzaninotto:
Could you provide a URL as a reference?

eidng8:
It's still the same wiki page. In the active codes list, there are notes that say "(funds code)", "(one troy ounce)", etc.

fzaninotto:
Thanks!

PR #1260, opened 2017-07-28
Add Internet Format for ja_JP.

itigoppo:
Unified Internet and Person format.

Because the internet class format is small and there are many duplications.


fzaninotto:
Thanks!

PR #1265, opened 2017-08-08
Added more Ukrainian streets and removed irrelevant names. Added more Ukrainian mobile formats

iamraccoon:
Removed some names of streets, which renamed after the revolution and added several new.
Added more Ukrainian mobile formats.


fzaninotto:
Thanks!

PR #1274, opened 2017-08-15
Adding zh_CN Lorem Provider

SpiderEvgn:


jonasva:
Need this as well

fzaninotto:

PR #1275, opened 2017-08-17
Replaced array() with [].

djh101:
Replaced using the following perl script:
`perl -i -pe 'BEGIN{undef $/;} s/([^a-z0-9_])array\(([^\(\)]*)\)/\1\[\2\]/smgi' */*/*.php`

Breaks support for php < 5.4. I'm guessing that may be why this change hasn't been made?


fzaninotto:
Exactly.

PR #1237, opened 2017-06-20
Fake html body

rudkjobing:

PR #971, opened 2016-07-20
HTML Lorem

rudkjobing:
A provider for faking HTML that should be easy to expand and improve.

Let me know if I have overlooked anything.

Regards Steffen.

PR #1277, opened 2017-08-21
Create HtmlBlock.php

Jazz-Man:
Added generates of html blocks. It will be useful to generate article content in a blog or something similar


Jazz-Man:
Fixed

fzaninotto:
What's wrong with HTML provider (cf #971)? If you just want the HTML *body*, then you'd better update the existing provider rather than creating a new one.

fzaninotto:
Besides, #1237 seems to go in the right direction already.

PR #1284, opened 2017-08-28
Get a random user from https://randomuser.me

guiwoda:
Hi there!
I added a `user()` function to call on https://randomuser.me in the simplest of ways. I was thinking of improving on it by adding some more API features (gender, nationality, etc.) but wanted to check if you'd be interested in adding that to Faker before continuing.

Test depends on availability of randomuser API, so I added a catch block to skip testing if its down.

Let me know if you're open for more integration to this service!


fzaninotto:
No thanks, all the fields can already be generated by Faker. Besides, we don't know anything about the `randomuser.me` random generator, and it adds an HTTP dependency (which we know is a bad idea, see the image generator).

PR #1287, opened 2017-09-01
Tax Id for companies and new formats for es_VE

DIOHz0r:
New function to generate tax ids for companies and the option to have a format for those ids.


fzaninotto:
Thanks! Can you add unit tests for the generators?

fzaninotto:
Thanks!

PR #1290, opened 2017-09-04
Fix CURL status code in ImageTest.php

Sanfra1407:
### Description
`http://lorempixel.com/` is being skipped because it responds with 301 instead of 200

### Code
`$url = "http://lorempixel.com/";   // line 48`
```
if ($httpCode < 200 | $httpCode > 300) {   // line 57
    $this->markTestSkipped("LoremPixel is offline, skipping image download");
}
```

### File
`test/Faker/Provider/ImageTest.php`


fzaninotto:
Thanks!

PR #1292, opened 2017-09-05
Update zh_TW text provider

s9801077:
Change the source of $baseText


fzaninotto:
Why?

s9801077:
I think it is necessary to increase the length and authenticity of $ baseText.
So I changed to use 『Call to Arms』, It is prose rather than ancient literature

fzaninotto:
OK, so to summarize:

- longer text leads to more variety
- 20th century work leads to up-to-date language

Thanks!

PR #1295, opened 2017-09-06
Added a Document Number for Dutch citizens

DiedeGu:
This adds the possibility to generate a valid Dutch Document number, which is placed in dutch passports. Reason for adding it was the recently added BSN number, to which the document number is related. 

Currently the function uses a private function randomCappedLetterNotO within the Dutch person, I would like to be advised on where to place the this function since it is a more general function, yet it might have only the current use case (Dutch Document numbers may not contain O's so that may not be confused with the number zero).

### DIFF
--- a/src/Faker/Provider/nl_NL/Person.php
+++ b/src/Faker/Provider/nl_NL/Person.php
@@ -278,6 +278,8 @@ class Person extends \Faker\Provider\Person
         return $lastName;
     }
 
+
+
     public function title($gender = null)
     {
         return static::randomElement(static::$title);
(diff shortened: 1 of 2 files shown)


fzaninotto:
Please add a mention of this formatter in the locale-specific section of the README

pimjansen:
@DiedeGu can you review the comments? Maybe also nice to link the actual spec in the phpdoc as well so we can easily verify later

pimjansen:
I am going to close this PR as stale. If this still needs to be merged, please feel free to reopen it.
Thanks for contributing

PR #1296, opened 2017-09-06
Allow children classes to access self::$suffix

greg0ire:
suffix() uses late static binding, and no longer works when called from
an extending class. Another solution would be to use self instead of
static, but other properties in that class are protected too, so this
should make it look more consistent.


greg0ire:
ping

fzaninotto:
Thanks!

PR #1297, opened 2017-09-06
NL Persons -> improved BSN function

JBtje:
Lets see if it works now.
Improved the Burger Service Number (BSN) function, both the generator and test.

### DIFF
--- a/src/Faker/Provider/nl_NL/Person.php
+++ b/src/Faker/Provider/nl_NL/Person.php
@@ -323,16 +323,16 @@ class Person extends \Faker\Provider\Person
     public function BSN()
     {
         do {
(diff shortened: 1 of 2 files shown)


fzaninotto:
Sorry, this is a breaking change, I won't merge it. Besides, you use `rand`, which is forbidden in Faker. You should definitely read the [CONTRIBUTING.md](https://github.com/fzaninotto/Faker/blob/master/CONTRIBUTING.md) guide.

artorozenga:
Maybe add the bsn function as alias?
`
    /**
     * Alias dutch personal identifier number
     */
    public static function bsn()
    {
        return self:: idNumber();
    }`

PR #1300, opened 2017-09-08
Semantic Versioning (semVer) provider

stevegrunwell:
This PR adds a Provider for software versions that follow [Semantic Versioning](http://semver.org/) (that is, the form of `MAJOR.MINOR.PATCH`).

The ranges used are 1-50, 0-50, and 0-999 for the major, minor, and patch releases, respectively.

Example output can include:

```
1.0.0
14.6.232
48.0.82
```


fzaninotto:
Nice idea. I suggest that you make it a standalone provider, though. Feel free to open a PR in Faker's README to mention it when it's published.

PR #1301, opened 2017-09-12
Language provider

espenkn:
This PR adds a Provider for languages. It has en_US and nb_NO language names, both taken from a list on wikipedia.

Example output can include:

> Afrikaans
> English
> Norwegian


fzaninotto:
Nice work. Although I advise that you create a specialized provider for that, and publish it as a standalone package.

PR #1302, opened 2017-09-12
Fixed type-o in readme under section about Language specific formatters

espenkn:
Fixed type-o in readme under section about Language specific formatters where it listed a language as no_NO not nb_NO.


fzaninotto:
Thanks!

PR #1306, opened 2017-09-25
Fix #1305 realText in some cases breaks last character

### LINKED ISSUE #1305: realText generates string with broken last character in some cases
When `realText` generates a string that ends with some specific character (for example, `р` (U+0440) in `ru_RU` or `uk_UA` locale; `є` (U+0454) in `uk_UA` locale) this character becomes broken. 
And when I trying to insert generated string into mysql database I get an error: 
`SQLSTATE[HY000]: General error: 1366 Incorrect string value: '\xD1.'`

Here is how to reproduce it:

```
$faker = Faker\Factory::create('ru_RU');
$faker->seed(208);
$title = $faker->realText(32);  // returns 'Какой-то этакой характе<broken_char_here>.'
```

Looks like this happens because of `rtrim($text, ',— ')` in `appendEnd` method. 
Em dash `—` is not an ASCII character so it can cause problems there. 

`rtrim` doesn't understand `—` as one character, but 3 bytes (`E2 80 94`) and it will trim any of those bytes individually. 
So, when text ends with `р` character (it has hex representation `D1 80`) `rtrim` cuts 'half a char' from it: `D1`~~`80`~~ and it becomes broken.

I think changing `rtrim` to `preg_replace` will solve the issue.

apotheosis91:
Changed `rtrim` to `preg_replace`


fzaninotto:
Could you add a unit test that proves the fix?

apotheosis91:
Sure, I've added unit test. 
Is it ok or should I test realText method with some seed values that cause broken char?
Something like this: 
```
/**
 * @dataProvider provideLocaleSeedAndExpectedResult
 */
public function testAppendEnd($locale, $seed, $expected)
{
    $faker = Factory::create($locale);
    $faker->seed($seed);
    $this->assertEquals($expected, $faker->realText(32));
 }

public function provideLocaleSeedAndExpectedResult()
{
    return array(
        array('ru_RU', 208, 'Какой-то этакой характер.'),
        array('uk_UA', 106, 'У мене свої закличники і своє.'),
    );
}
```

test/Faker/Provider/TextTest.php
```diff
@@ -46,9 +46,29 @@ public function testTextMinIndex()
...
+     * @dataProvider provideTextAndExpectedResult
+     */
+    public function testTextAppendEnd($text, $expected)
+    {
+        $appendEndMethod = new \ReflectionMethod('Faker\Provider\Text', 'appendEnd');
+        $appendEndMethod->setAccessible(true);
+        $this->assertEquals($expected, $appendEndMethod->invokeArgs(null, array($text)));
+    }
+
+    public function provideTextAndExpectedResult()
```
fzaninotto: please rename to `TextAppendEndDataProvider` (also in the phpDoc above)

fzaninotto:
I prefer the test as it's currently committed, it's more unitary than using `realText`

fzaninotto:
Superseded by #1380

PR #1309, opened 2017-09-28
Fix with incorrect name city.

zzenmate:
My hometown has incorrect name. I am very upset by this.
Correct name is 'Хмельницький'.
[Link to wiki about correct name.](https://uk.wikipedia.org/wiki/%D0%A5%D0%BC%D0%B5%D0%BB%D1%8C%D0%BD%D0%B8%D1%86%D1%8C%D0%BA%D0%B8%D0%B9)


fzaninotto:
No reason to be upset, it's now fixed. Thanks!

PR #1310, opened 2017-09-29
Added German and Dutch names

Richard-NL:


fzaninotto:
From the contributing guide:

PR #1314, opened 2017-10-04
Update pt_BR names

thiagotalma:
Change Hispanic names by genuinely Brazilian names.
( because Brazil is a Latin country, but it is not Hispanic, it is Portuguese 😉 )

### DIFF
--- a/src/Faker/Provider/pt_BR/Person.php
+++ b/src/Faker/Provider/pt_BR/Person.php
@@ -37,55 +37,41 @@ class Person extends \Faker\Provider\Person
     );
 
     protected static $firstNameMale = array(
-        'Aaron', 'Adriano', 'Alan', 'Alexandre', 'Alonso', 'Anderson', 'Andres', 'Antônio', 'Benjamin', 'Bruno', 'Camilo', 'Carlos', 'Christian',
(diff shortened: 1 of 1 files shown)


fzaninotto:
Please find an external datasource for these arrays, with an open-source licence. Otherwise every dev in the world can say they have better data and I have no way to tell which one is right.

thiagotalma:
I understand your preoccupation.
However I checked in the change history that there was no major change to this file. No Brazilian native had the job of writing Brazilian names. I am the first to do this.

Why not accept my suggestions and create difficulties for the next PR who try to recreate everything again?

Or you may even have the job of correcting this manually and closing the PR.
Use http://www.fakenamegenerator.com

What can not do is continue to give Hispanic names to Brazilians 😉

fzaninotto:
> Why not accept my suggestions and create difficulties for the next PR who try to recreate everything again?

Because it's not because I made mistakes in the past (by accepting data without a proper source) that I want to continue making more.

dknacht:
Hi, I can guarantee that those name he erased were **Hispanic**, not Brazilian at all.
What type external of source would you accept? I'll try to find one, but while I search for a trusted and open one, I think that the names he wrote, even if are not 100% correct are much better than the ones are being used actually.

fzaninotto:
Any MIT-compatible datasource will fit.

In the meantime, I won't merge this PR. I will close it in a few weeks if there is no update.

thiagotalma:
There is no update I can do, you can close now.

I would gain nothing at all with this, my intention was only to collaborate.

PR #1315, opened 2017-10-05
South African licence codes

royalmitten:
Added licenceCode method in the en_ZA person provider, that generates a random valid South African licence code


fzaninotto:
Thanks!

PR #1090, opened 2016-11-16
Introduced the ability to specify the timezone for dateTimeThis*() methods

telkins:
Introduced the ability to specify the timezone for the following methods:
* dateTimeThisCentury(),
* dateTimeThisDecade(),
* dateTimeThisYear(), and
* dateTimeThisMonth().

Updated readme and introduced new unit tests.

PR #1316, opened 2017-10-07
Introduced the ability to specify a default timezone...

telkins:
Inspiration: @curry684 

This PR introduces the ability to set a default timezone that will be used when no timezone is provided by a caller.

It does *not* break existing code, to the best of my knowledge, because the default value is `null` and when it's set to `null`, then it gets the timezone from `date_default_timezone_get()`.  (One of the new unit tests covers this scenario.)

Users can set the default timezone via `$faker->setDefaultTimezone('Europe/Copenhagen')` and can see its current value via `$faker->getDefaultTimezone()`.

If `$faker->setDefaultTimezone()` is called, then it effectively "resets" the default timezone value to `null`, which means that timezones will again be determined by `date_default_timezone_get()` when the caller does not provide a timezone.

A little refactoring was done in the `DateTime` provider, moving the timezone resolution logic into one place instead of having it spread out.

The `DateTime` provider test class was updated a bit, making use of the new functionality in the `setUp()` and `tearDown()` methods.  Additionally, two new tests were added to test some of this new functionality.

The documentation was updated a bit to provide a little explanation on how the timezones would be determined an how users could get and set the timezones using both the new and old functionality.

Finally, the Romanian `Person` test class was also updated a bit, also making use of the new functionality.  It was unclear to me, however, that the timezone changing in that class had any impact.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -180,17 +180,17 @@ Each of the generator properties (like `name`, `address`, and `lorem`) are calle
 ### `Faker\Provider\DateTime`
 
     unixTime($max = 'now')                // 58781813
-    dateTime($max = 'now', $timezone = date_default_timezone_get()) // DateTime('2008-04-25 08:37:17', 'UTC')
-    dateTimeAD($max = 'now', $timezone = date_default_timezone_get()) // DateTime('1800-04-29 20:38:49', 'Europe/Paris')
+    dateTime($max = 'now', $timezone = null) // DateTime('2008-04-25 08:37:17', 'UTC')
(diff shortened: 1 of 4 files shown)


telkins:
@curry684 I tried to add you as a reviewer.  You suggested this nearly a year ago: https://github.com/fzaninotto/Faker/pull/1090

Was this what you had in mind, more or less...?

curry684:
I don't have time for a thorough review right now but yes this is what I was hinting at   :)

telkins:
@curry684 Thanks for responding.  No worries.  I just wanted to make sure that I was in the right ballpark.  :-)

Please let me know if you get time and have any suggestions, questions, whatever.  TTYL....

fzaninotto:
Thanks!

PR #1319, opened 2017-10-11
Fix "Resource temporarily unavailable"

eberkund:
When trying to unlink the file after unsuccessful download, a warning is thrown because file is still open


fzaninotto:
Thanks!

PR #1321, opened 2017-10-13
Added male and female titles for the en_ZA locale

ViGouRCanberra:
- The new titles are for Revenant (Rev.) and Honourable (Hon.)
- Remove unused use import for Faker\Provider\DateTime


fzaninotto:
Thanks!

PR #1323, opened 2017-10-17
Real Dutch postal codes

ametad:
The included postal codes are randomly taken from a database with real Dutch codes.


fzaninotto:
No thanks, I prefer a generator.

ametad:
Hello fzaninotto,

What has changed? I saw a similar PR for the Belgian postal codes that you accepted: https://github.com/fzaninotto/Faker/pull/1004

In our tests we also need real existing codes and I would be very grateful if this could be implemented in your Faker lib.

fzaninotto:
in #1004, I understood that these were the only postal codes in Belgium (my mistake?), so there was no need for more randomness. From your PR, I understand that you're including only a subset of the possible post codes. So in effect, you're reducing the possible values of the Dutch post codes.

If some of the generated post codes aren't valid, maybe you can update the formatter to match a certain format?

ametad:
Thank you for your explanation.

There are a lot more postal codes in Belgium! And indeed my PR is also a subset of all the postal codes in the Netherlands. 

The thing is; I need codes that are really in use. And this is not something that can be calculated. Although the format "%### ??" is correct, not all possible outcome is (already) in use today, but could be in use tomorrow if a new part of a city is build (suburb with some streets). You see, in the Netherlands every street part (even side, uneven side, high number, low numbers) has it own postal code! So in the same street the houses can have two, three, more, different postal codes.

For my tests, I need postal codes that are in use today. Because it is compared with an external services, so that input forms only accept existing codes (validation) and therefore are real addresses. 

Will you accept my PR?

ametad:
I could expand the list to (say) double the codes, so there is more randomness.

fzaninotto:
OK, I'll reopen this. Please triple the list size.

ametad:
Will do triple the size and check back when ready. Thank you.

ametad:
Tests are green :)

fzaninotto:
Thanks!

PR #1325, opened 2017-10-18
Add numberBetweenAndUnqiue function

tuanphpvn:
Add handy function numberBetweenAndUnique.

Usage:

numberBetweenAndUnique(1,10000, 1);  // for instance 1
numberBetweenAndUnique(1,10000, 2); // for instance 2


tuanphpvn:
@fzaninotto All checks have passed. Please review.

francislavoie:
I'm not sure I see the point of this. You can just do `$faker->unique()->numberBetween(1, 10000);`. If you need two collections of uniques, just make two instances of Faker and use them individually. Barring that, probably a better solution is implementing separate unique instances on the unique validator itself.

fzaninotto:
I agree with @francislavoie, I don't see the interest of this.

PR #1327, opened 2017-10-19
Doctrine : use words when string length < 5

pierre-H:
Since text doesn't support when length < 5.


pierre-H:
Sorry I confused word and letter ...

PR #1328, opened 2017-10-19
Update city name

s9801077:
On December 25, 2014, Taoyuan County (桃園縣) was from a county to upgrade a special municipality named Taoyuan City (桃園市).


fzaninotto:
Thanks!

PR #1330, opened 2017-10-20
Add missing --dev flag

tuanphpvn:
Used library for development only.


fzaninotto:
Unless some developers want to use Faker in test environment... Let them decide.

tuanphpvn:
Environments, whereas the dependencies declared in the require-dev section are typically dependencies which are required in:

developing
testing

PR #1331, opened 2017-10-20
[sl_SI] Created provider for Company

alesvaupotic:
Companies in Slovenija always have a suffix. Also, "and" is translated to "in".


fzaninotto:
Thanks!

PR #1333, opened 2017-10-22
Change symbol 'minus' with code 226 to 'minus' with code 45

Negasus:
With utf-8 charset call trim with symbol "- (code 226)" work uncorrect.

Example with Russian string
```
$s = 'Пример';
$res = trim($s, '—');
var_dump($res);

// result: string(11) "Приме�"
```


fzaninotto:
Thanks!

PR #1335, opened 2017-10-24
Fix typo

davidporos92:
Add more names


fzaninotto:
Low value PR. See the Contributing.md:

PR #1336, opened 2017-10-24
Add two new company type in the Swiss Provider

pvullioud:


fzaninotto:
I'd prefer if you added all the possible suffixes, instead of adding just a few ones.

pvullioud:
it's the exhaustive list of legal forms for company in Switzerland :)

fzaninotto:
Then Thanks

PR #1337, opened 2017-10-25
Prevent problems in the url

seltix5:
this will convert the text to url safe.
for example, convert spaces into "%20"

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1358,6 +1358,15 @@ echo $faker->bankAccountNumber; // "PL14968907563953822118075816"
 echo $faker->taxpayerIdentificationNumber; // '165249277'
 ```
 
+### `Faker\Provider\pt_PT\PhoneNumber`
+
+```php
+<?php
+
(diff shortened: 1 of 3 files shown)


fzaninotto:
This PR does not what its description says.

seltix5:
hi,
should I divide the file changes or not interested?
thanks

fzaninotto:
Start by explaining the problem(s) you want to solve. Reading code without knowing why it's there isn't my favorite hobby.

seltix5:
the origin pull request (Prevent problems in the url) was just for the first commit (https://github.com/fzaninotto/Faker/pull/1337/commits/4710ac644a45f47d4c4a25698fc416e0eb834eaf) where i only add the rawurlencode() to encode spaces and other special chars.
In the second file commit I just add a method to return the categories from the Image class.
The 2 next commits I added "only landline numer" option in my country (PT) PhoneNumber class and updated readme.md.
In the last comits I just added a multiple engine possibility for the Image class.

PR #1340, opened 2017-10-30
Added new ZA gTLD domains

leepelser:


fzaninotto:
Sorry, too low value. Each modification of the data breaks existing tests using seeded results, so I only accept larger modifications than this one.

PR #1341, opened 2017-10-31
Added Vehicle provider

ViGouRCanberra:
- Added a few basic models that are pretty generic internationally

### DIFF
--- a/src/Faker/Factory.php
+++ b/src/Faker/Factory.php
@@ -6,7 +6,7 @@ class Factory
 {
     const DEFAULT_LOCALE = 'en_US';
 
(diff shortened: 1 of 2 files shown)


ViGouRCanberra:
I'll add Unit Tests for this later today, just first wanted to check if this would be a wanted feature for Faker?
The plan is to extend this to Vehicle license plate generator, which is a feature I'd very much like from Faker :D

fzaninotto:
Nice, but I prefer if you ship it as a standalone provider. When it's released, please open a PR on Faker's README to mention it!

ViGouRCanberra:
Will do, thank you.

PR #1343, opened 2017-11-02
Fix FR_fr 07 prefix mobile number generation

svanpoeck:
Bonjour François,

For the FR_fr provider, the `$faker->mobileNumber` does not call  the `phoneNumber07` and `phoneNumber07WithSeparator` methods because method `mobileNumber` does not invoke the generator but instead calls `static::numerify()` directly.
Also, not all of the 07xx formats have been transposed to `static::$mobileFormats`, and not all 07 formats call upon `phoneNumber07` or `phoneNumber07WithSeparator`, whereas they should.

I've shamelessly taken PR https://github.com/fzaninotto/Faker/pull/1304 from ctessier to fix the inners of the `mobileNumber()` method, and have replaced the `#` with `{{phoneNumber07[WithSeparator]}}` where approriate.
I've also added some tests (only tested the regular expressions, did not actually run the tests as I don't have PHPUnit installed).

Hoping you can merge this in soonish, our database seed fails because we use https://github.com/giggsey/libphonenumber-for-php to check phone number validity and currently part of the 07xx ones generated by Faker don't pass the isValid test.

### DIFF
--- a/src/Faker/Provider/fr_FR/PhoneNumber.php
+++ b/src/Faker/Provider/fr_FR/PhoneNumber.php
@@ -52,12 +52,14 @@ class PhoneNumber extends \Faker\Provider\PhoneNumber
     // Mobile phone numbers start by 06 and 07
     // 06 is the most common prefix
     protected static $mobileFormats  = array(
+        '+33 (0)6 ## ## ## ##',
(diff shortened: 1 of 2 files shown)


fzaninotto:
Thanks!

PR #1344, opened 2017-11-04
Use PSR-1 for PHPUnit TestCase

gabriel-caruso:
Use [PSR-1](http://www.php-fig.org/psr/psr-1/#namespace-and-class-names) while extending PHPUnit TestCase class. This will help us when to migrate to PHPUnit 6, that [no longer support snake case namespaces](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#600---2017-02-03).

### DIFF
--- a/test/Faker/Calculator/IbanTest.php
+++ b/test/Faker/Calculator/IbanTest.php
@@ -3,8 +3,9 @@
 namespace Faker\Test\Calculator;
 
 use Faker\Calculator\Iban;
+use PHPUnit\Framework\TestCase;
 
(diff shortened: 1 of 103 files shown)


fzaninotto:
Thanks!

MarioBlazek:
Btw, just for the sake of security update PHPUnit version constraint to `"phpunit/phpunit": "^4.8.35 || ^5.0",` in composer.json as this namespaced `TestCase` was backported to v4 in `4.8.35`.

PR #1348, opened 2017-11-10
only test available date range on 32-bit

remicollet:
Related to #1347

### DIFF
--- a/test/Faker/Provider/fi_FI/PersonTest.php
+++ b/test/Faker/Provider/fi_FI/PersonTest.php
@@ -44,18 +44,25 @@ class PersonTest extends TestCase
 
     public function testPersonalIdentityNumberGeneratesCompliantNumbers()
     {
-        for ($i = 0; $i < 10; $i++) {
-            $birthdate = $this->faker->dateTimeBetween('1800-01-01 00:00:00', '1899-12-31 23:59:59');
-            $pin = $this->faker->personalIdentityNumber($birthdate);
-            $this->assertRegExp('/^[0-9]{6}\+[0-9]{3}[0-9ABCDEFHJKLMNPRSTUVWXY]$/', $pin);
+        if (strtotime('1800-01-01 00:00:00')) {
+            $min="1900";
+            $max="2099";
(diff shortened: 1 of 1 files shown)


test/Faker/Provider/fi_FI/PersonTest.php
```diff
@@ -44,18 +44,25 @@ public function testPersonalIdentityNumberUsesBirthDateIfProvided($seed, $birthd
...
+        if (strtotime('1800-01-01 00:00:00')) {
+            $min="1900";
+            $max="2099";
+            for ($i = 0; $i < 10; $i++) {
+                $birthdate = $this->faker->dateTimeBetween('1800-01-01 00:00:00', '1899-12-31 23:59:59');
+                $pin = $this->faker->personalIdentityNumber($birthdate, NULL, true);
+                $this->assertRegExp('/^[0-9]{6}\+[0-9]{3}[0-9ABCDEFHJKLMNPRSTUVWXY]$/', $pin);
+            }
+        } else { // timestamp limit for 32-bit computer
+            $min="1902";
```
fzaninotto: why not 1900?

remicollet: because strtotime("1900...") return false on 32-bit

remicollet: # php -r 'var_dump(strtotime("1900-01-01"));'
bool(false)
# php -r 'var_dump(strtotime("1902-01-01"));'
int(-2145916800)
# php -r 'var_dump(strtotime("2038-01-01"));'
int(2145916800)
# php -r 'var_dump(strtotime("2039-01-01"));'
bool(false)

fzaninotto: Understood. According to http://php.net/manual/en/function.strtotime.php, the safe limits are 1902 and 2037. So why 2036 in your code?

remicollet: 2036 was a typo ;)

fzaninotto:
Thanks!

PR #1350, opened 2017-11-11
nb_NO provider updates

ahjorth:
* Added more and more common business suffixes*
* Added 1500 random jobs from the official government statistics bureau

\* `$companySuffix` data set was changed, but no values were removed, only added. If this is a problem, feel free to revert it to the original `'AS', 'DA', 'NUF'`


fzaninotto:
Thanks!

PR #1353, opened 2017-11-14
Add --dev flag to readme

shrikeh:
There are almost no circumstances where `faker` should be a requirement for a production environment. Instructions on installation tweaked to add the `--dev` flag to the `composer require` instruction to aid copy-paste.


fzaninotto:
duplicate of #1330

PR #1354, opened 2017-11-16
Add Lorem for zh_TW

asika32764:
This is a Lorem Provider for Traditional Chinese (zh-TW)

I use I Ching words to make the lorem text. (Which is an ancient Chinese divination text and the oldest of the Chinese classics)

See https://en.wikipedia.org/wiki/I_Ching

### DIFF
--- a/src/Faker/Provider/zh_TW/Lorem.php
+++ b/src/Faker/Provider/zh_TW/Lorem.php
@@ -1,0 +1,198 @@
+<?php
+
+namespace Faker\Provider\zh_TW;
+
+use Faker\Provider\Base;
+
+class Lorem extends Base
+{
+    /**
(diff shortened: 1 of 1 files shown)


fzaninotto:
We need a contemporary work, in order to generate normal-looking text.

asika32764:
Sorry I'm not fully understand about `contemporary` means since I'm not primary on English.

But as my knowledge, English lorem is also has long history and not a modern text, it's also not looking like a normal text.

`I Ching` uses the Chinese words that Chinese prople all can read it, but after randomize, it can be a text which is readable but meaningless.

fzaninotto:
It means recent enough (less than a century) for the language to be common.

PR #1360, opened 2017-11-23
Get currency codes from library to stay up to date

tpraxl:
This is a proposal to stay up to date with changing currency codes. 

Background:
In our current project, we perform validation of currency codes. In UnitTests, these codes are generated by Faker and produce randomly failing tests due to a divergence between actual currency codes and the ones that Faker produces. 

This applies to the following codes, which are not considered valid by our application that uses `alcohol/iso4217` to retrieve the codes:

* ZWL
* SVC
* BYN

Using the lib makes it more likely to stay up to date with the changes.

If you don't like the idea of adding that lib, please let me know. I would update the static list of currency codes to reflect the current state as well.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,8 @@
         }
     ],
     "require": {
-        "php": "^5.3.3 || ^7.0"
+        "php": "^5.3.3 || ^7.0",
+        "alcohol/iso4217": "^3.1"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.8.35 || ^5.7",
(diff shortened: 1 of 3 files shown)


tpraxl:
Sorry, just noticed that my PR breaks on php < 5.5:

* alcohol/iso4217 requires php 5.4
* I used array_column which requires php 5.5

@fzaninotto Please let me now if you're interested in the approach at all. If so, please let me know if you are willing to raise the php version requirement to 5.4 or even 5.5. Otherwise, I would create another PR with an update of the static list of currency codes.

fzaninotto:
I'm reluctant to add a dependency just for that. People use Faker without any additional a-installation setup ; breaking that expectation requires a serious reason, ans I don't think this is one.

Also, we won't change the PHP version requirement, for BC purposes.

tpraxl:
Thanks for your feedback. I proposed another approach to the issues: #1373

PR #1361, opened 2017-11-24
adds rijksregisternummer

ppelgrims:
This will generate a Belgian social security number. I've added tests as well.


ppelgrims:
I'm adding documentation

fzaninotto:
Thanks!

PR #1363, opened 2017-11-27
Prevent errors on private methods

petecoop:
Use `is_callable` to prevent errors if a method is private.

### DIFF
--- a/src/Faker/ORM/Doctrine/EntityPopulator.php
+++ b/src/Faker/ORM/Doctrine/EntityPopulator.php
@@ -211,7 +211,7 @@ class EntityPopulator
                 }
                 // Try a standard setter if it's available, otherwise fall back on reflection
                 $setter = sprintf("set%s", ucfirst($field));
-                if (method_exists($obj, $setter)) {
+                if (is_callable($obj, $setter)) {
                     $obj->$setter($value);
                 } else {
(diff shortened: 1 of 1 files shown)


fzaninotto:
in which case does the current code produce an error?

petecoop:
If for whatever reason in the codebase the setter is a private method, this will still try and call it producing an error

fzaninotto:
Still unclear to me. Can you add a failing test case to demonstrate the problem ?

petecoop:
Are there any tests around doctrine population at all?

fzaninotto:
probably not. But please paste a code sample here for me to understand.

petecoop:
so take some Doctrine model like the below, for some reason within the codebase the `setName` method has been set to private - perhaps it's only ever called in the constructor and not meant to be called outside of it. This will currently cause an error when using `Faker\ORM\Doctrine\Populator` as it is automatically counting a method beginning with `set` as publicly available.

```php
<?php
class Product
{
    /**
     * @var int
     */
    protected $id;
    /**
     * @var string
     */
    protected $name;

    public function getId()
    {
        return $this->id;
    }

    public function getName()
    {
        return $this->name;
    }

    private function setName($name)
    {
        $this->name = $name;
    }
}
```

fzaninotto:
Thanks!

Issue #1346, opened 2017-11-10
Test failure on BigEndian


remicollet:
See Fedora scratch build (version 1.7.1 against PHP 7.2.0RC6)
https://koji.fedoraproject.org/koji/taskinfo?taskID=23038022

BigEndian (ppc64 and s390x):

```
There was 1 failure:
1) Faker\Test\Provider\UuidTest::testUuidExpectedSeed
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'8e2e0c84-50dd-367c-9e66-f3ab455c78d6'
+'840c2e8e-dd50-3c46-9e66-f3ab455c78d6'
/builddir/build/BUILD/Faker-d3ed4cc37051c1ca52d22d76b437d14809fc7e0d/test/Faker/Provider/UuidTest.php:20
```

fzaninotto:
OK, then let's just say that Faker doesn't support BigEndian OSes.

remicollet:
Only for a single test ?
Especially as there is code to handle bidendian ?
=> https://github.com/fzaninotto/Faker/blob/master/src/Faker/Provider/Uuid.php#L27

fzaninotto:
yes, because I don't want to download an entire OS to test a patch.

PR #1365, opened 2017-11-29
skip test failing on bigendian

remicollet:
According to your answer in
https://github.com/fzaninotto/Faker/issues/1346#issuecomment-343504794


fzaninotto:
Thanks

PR #1368, opened 2017-11-30
Wtoalabi patch 1

wtoalabi:
I would like to work on the Nigerian version of the locale.


fzaninotto:
From the Contributing guidelines:

PR #1369, opened 2017-11-30
default value for Payment::iban() country code

madmanmax:
matching the function signature with the doc as defined in [`Faker\Generator`](https://github.com/fzaninotto/Faker/blob/9cda1a418ca4b2b71c9b921ebaea6c2e61e270b4/src/Faker/Generator.php#L48):

` @method string iban($countryCode = null, $prefix = '', $length = null)`


fzaninotto:
Thanks

PR #1371, opened 2017-11-30
Test against PHP 7.2

carusogabriel:
With the release of [`PHP 7.2`](http://php.net/archive/2017.php#id2017-11-30-1), would be nice to test Faker against it.


fzaninotto:
Thanks!

PR #1373, opened 2017-12-01
Remove obsolete currency codes

tpraxl:
This is an alternative approach to solve the issue described in #1360.

I removed SVC and ZWL from the list of currency codes.

While still in the list of official [ISO 4217 currency names](https://en.wikipedia.org/wiki/ISO_4217#Active_codes), these currencies are obsolete:

SVC has been replaced by the USD in 2001:
https://en.wikipedia.org/wiki/Salvadoran_col%C3%B3n

ZWL has been suspended since 2009:
https://en.wikipedia.org/wiki/Zimbabwean_dollar

Using them in Faker can cause 3rd party validations to fail.


fzaninotto:
Thanks!

PR #1375, opened 2017-12-03
Refactoring tests

carusogabriel:
I've refactored some tests, using:
- `assertInternalType` instead of `is_*` methods;
- `assertNull` and `assertNotNull` instead of strict comparisons with `null` keyword;
- `assertTrue` instead of strict comparisons with `true` keyword;
- `assertFalse` and `assertNotFalse` instead of strict comparisons with `false` keyword;
- `assertGreaterThan`, `assertGreaterThanOrEqual`, `assertLessThan` and `assertLessThanOrEqual` for mathematical comparisons;
- `assertSame` and `assertNotSame` instead of strict comparisons;
- `assertContains` instead of `in_array` method.


fzaninotto:
Thanks!

PR #1376, opened 2017-12-04
Add Chinese (simplified) Lorem and Text provider

ganlvtech:


ganlvtech:
Lorem has moved to another repo https://github.com/ganlvtech/faker-chinese-lorem .
even after reading #1011 , I still don't know how to provide a Lorem provider gracefully.

The base text was written in 1926, and the previous link was wrong by my mistake. I have fixed it. Now, it reference to a correct location.

readme.md
```diff
@@ -1539,6 +1539,16 @@ echo $faker->personalIdentityNumber('female') // '950910-0781'
 echo $faker->bank; // '中国建设银行'
 ```
 
+### `Faker\Provider\zh_CN\Text`
+
+```php
+<?php
+
+// Generate a random single Chinese character
+echo $faker->realText;
```
fzaninotto: no need to mention providers available in all languages.

src/Faker/Provider/zh_CN/Text.php
```diff
@@ -0,0 +1,176 @@
...
+     * Language: Simplified Chinese
+     * Written on: Sept 18, 1926
+     *
+     * Notice: Some punctuation has been removed. Some paragraph and quotations have be joined together.
+     *
+     * @see https://zh.wikisource.org/zh-hans/%E4%BB%8E%E7%99%BE%E8%8D%89%E5%9B%AD%E5%88%B0%E4%B8%89%E5%91%B3%E4%B9%A6%E5%B1%8B
+     * @var string
+     */
+    protected static $baseText = <<<'EOT'
+我家的后面有一个很大的园，相传叫作百草园。现在是早已并屋子一起卖给朱文公的子孙了，连那最末次的相见也已经隔了七八年，其中似乎确凿只有一些野草；但那时却是我的乐园。
```
fzaninotto: This text is too short imo, the result won't be random enough. for a Markov chain generator to work, each word / character needs to appear many times in the base text.

ganlvtech:
[x] Readme changed.
[x] Text long enough.
[x] No lorem provider.
[x] Test case added.

src/Faker/Provider/zh_CN/Text.php
ganlvtech: I have select more text from a collection. I use 4 articles together as base to make it long enough.

src/Faker/Provider/zh_CN/Text.php
```diff
@@ -0,0 +1,1063 @@
...
+                // Unexpected first byte of UTF-8 character
+            }
+        }
+        return $count;
+    }
+
+    /**
+     * Get UTF-8 string characters array
+     *
+     * TODO: this function is not used
```
pimjansen: Why is this method in here when it is not used? And in general speaking i dont think such a method even belongs in a locale. If it belongs somewhere then it would even be in the core itself

src/Faker/Provider/zh_CN/Text.php
```diff
@@ -0,0 +1,1063 @@
...
+    /**
+     * UTF-8 strlen
+     *
+     * @see https://en.wikipedia.org/wiki/UTF-8#Description
+     *
+     * @param string $text
+     *
+     * @return int
+     */
+    protected static function utf8Strlen($text)
```
pimjansen: wont the php native mb functions handle this properly nowadays?

ganlvtech: I'm happy to change all this messy code to `mb_*` function. But there is no `ext_mbstring` in package.json. So I don't use these functions.

src/Faker/Provider/zh_CN/Text.php
ganlvtech: This function `utf8Chars` is a fallback for `explode`. I don't know if `preg_split` with `/u` works well for utf8 text. So I provide a fallback function. But it's truly not used.

ganlvtech:
@pimjansen What do think about adding `ext-mbstring` to `package.json`

pimjansen:
@ganlvtech not a fan of this since it would mean a breaking change. We are not allowing it till we start with Faker 2.0 later this year. 

So for 1.x we cant

singleseeker:
Any update？

PR #1382, opened 2017-12-14
Clean elses

carusogabriel:
I've removed the `else`s when we have already returned something :put_litter_in_its_place:


fzaninotto:
Thanks!

PR #1385, opened 2017-12-18
update th_TH mobile number format

earthpyy:
update new Thailand (th_TH) mobile number format
- 06x xxx xxxx
- 06 xxxx xxxx


fzaninotto:
Thanks

PR #1389, opened 2017-12-18
allow `randomElements` to accept a Traversable object

browner12:
Often times when using Faker in our seeders, we will have Collection objects, not simple arrays. This PR allows the `randomElement()` and `randomElements()` methods to accept both arrays **and** objects that implement the `Traversable` interface.

rather than adding a bunch of conditionals to deal with arrays or objects, I decided to just normalize all Traversable objects to arrays at the start.  this lets the existing code operate as before.

### DIFF
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -168,8 +168,19 @@ class Base
      *
      * @return array New array with $count elements from $array
      */
-    public static function randomElements(array $array = array('a', 'b', 'c'), $count = 1, $allowDuplicates = false)
+    public static function randomElements($array = array('a', 'b', 'c'), $count = 1, $allowDuplicates = false)
     {
+        if ($array instanceof \Traversable) {
+
+            $return = [];
+
+            foreach($array as $element){
+                $return[] = $element;
+            }
+
+            $array = $return;
+        }
+
         $allKeys = array_keys($array);
         $numKeys = count($allKeys);
 
@@ -206,7 +217,7 @@ class Base
      */
     public static function randomElement($array = array('a', 'b', 'c'))
     {
-        if (!$array) {
+        if (!$array || ($array instanceof \Traversable && !count($array))) {
             return null;
         }
         $elements = static::randomElements($array, 1);
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -2,8 +2,8 @@
 
 namespace Faker\Provider;
 
-use Faker\DefaultGenerator;
 use Faker\Generator;
+use Faker\DefaultGenerator;
 use Faker\UniqueGenerator;
 use Faker\ValidGenerator;
 
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -2,8 +2,8 @@
 
 namespace Faker\Provider;
 
-use Faker\Generator;
 use Faker\DefaultGenerator;
+use Faker\Generator;
 use Faker\UniqueGenerator;
 use Faker\ValidGenerator;
 
@@ -174,7 +174,7 @@ class Base
 
             $return = [];
 
-            foreach($array as $element){
+            foreach ($array as $element) {
                 $return[] = $element;
             }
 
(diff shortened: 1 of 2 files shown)


src/Faker/Provider/Base.php
```diff
@@ -168,8 +168,19 @@ public static function randomAscii()
...
     {
+        if ($array instanceof \Traversable) {
+
+            $return = array();
+
+            foreach ($array as $element) {
+                $return[] = $element;
+            }
+
+            $array = $return;
```
fzaninotto: bad practice to update a parameter. Prefer using a new variable with a ternary expression

browner12:
updated.  is this what you're thinking?

fzaninotto:
PLease use a new variable to avoid updating a variable passed as parameter

browner12:
I am not following what you mean by that. Can you provide an example?

fzaninotto:
Instead of

```php
$array = isset($traversables) ? $traversables : $array;
```

Write 

```php
$arr = isset($traversables) ? $traversables : $array;
```

To avoid modifying `$array`, which is passed to the function.

browner12:
I don't really see the benefit to this change, and would actually argue it's better the other way, but here's the update.

 what we're doing is standardizing the input. we accept both arrays and traversable objects as our parameter, but we want to write code that only deals with arrays, so we standardize everything to an array.

am i missing something?

src/Faker/Provider/Base.php
```diff
@@ -168,9 +168,17 @@ public static function randomAscii()
      *
      * @return array New array with $count elements from $array
      */
-    public static function randomElements(array $array = array('a', 'b', 'c'), $count = 1, $allowDuplicates = false)
+    public static function randomElements($array = array('a', 'b', 'c'), $count = 1, $allowDuplicates = false)
```
fzaninotto: can you read the type hint?

src/Faker/Provider/Base.php
```diff
@@ -168,9 +168,17 @@ public static function randomAscii()
...
     {
-        $allKeys = array_keys($array);
+        if ($array instanceof \Traversable) {
+            foreach ($array as $element) {
+                $traversables[] = $element;
```
fzaninotto: You need to set `traversables` to an empty array first I guess.

src/Faker/Provider/Base.php
```diff
@@ -168,9 +168,17 @@ public static function randomAscii()
...
+            }
+        }
+
+        $arr = isset($traversables) ? $traversables : $array;
```
fzaninotto: Don't check `isset` but zero length

src/Faker/Provider/Base.php
browner12: i don't know what you mean by this.

src/Faker/Provider/Base.php
browner12: i could, but this is why i use the `isset` below.  it's 2 ways to accomplish the same thing.

src/Faker/Provider/Base.php
fzaninotto: Just revert your changes to this line

src/Faker/Provider/Base.php
fzaninotto: Then use only 1 intermediate variable if you want, but pushing to a non-existing array will cause a warning. If it doesn't, it's a bad practice anyway.

src/Faker/Provider/Base.php
browner12: the whole point of this PR is to accept things other than arrays, so if I add the type hint back, that negates all of the new code.

fzaninotto: ok, you're right, my bad

src/Faker/Provider/Base.php
```diff
@@ -170,13 +170,15 @@ public static function randomAscii()
      */
     public static function randomElements($array = array('a', 'b', 'c'), $count = 1, $allowDuplicates = false)
     {
+        $traversables = [];
```
fzaninotto: you must use `array()` to keep PHP 5.3 compatibility.

browner12:
good call, force of habit.

fzaninotto:
Merged! Thank you very much.

browner12:
glad to help

PR #1392, opened 2017-12-27
Add licenceCode method in the to es_ES person provider

ffiguereo:
Added licenceCode method in the en_Es person provider, that generates a random valid Spain licence code


fzaninotto:
Thanks!

PR #1394, opened 2018-01-04
Added a random alpha numeric string generator

robbytaylor:
This PR adds a generator for arbitrary length alpha numeric strings. I needed this for a project but couldn't find anything which already does it. Apologies, and please let me know, if I've missed something.


robbytaylor:
Ok, so I realised I did miss something quite significant in that what I was trying to do can be achieved easily with the `randomElements()` method.

I'll leave this open though in case it might be useful as an alias for `randomElements()`

fzaninotto:
Why can't you use `asciify`?

robbytaylor:
asciify includes special characters. I just wanted alpha numeric characters.

PR #1395, opened 2018-01-06
Remove unused variable inside closure

carusogabriel:


fzaninotto:
I think the mistake is rather on line 160, where `$this->locator` should be `$locator`

fzaninotto:
Thanks!

PR #1400, opened 2018-01-11
VISA retired the 13 digit PAN moved to new cardParams

hppycoder:
VISA retired the 13 digit PAN and 468 exist as of Dec 11, 2013. I have moved the retired card length into its own cardParams ("Visa Retired") to allow backward compatibility and allow the card length to still be generated if needed.


fzaninotto:
Thanks!

PR #1404, opened 2018-01-15
TH color name

Naruedom:


Naruedom:
?

ppelgrims:
There is a whitespace at the end of line 9, if you ran `make sniff test` you would've found out before the CI script did, it runs the same command.

fzaninotto:
Thanks!

PR #1405, opened 2018-01-15
Update Person.php

oversluizen:
Some names added


fzaninotto:
Low added value PR. Every change in an array of data breaks all the tests depending on the related formatter. So changing the data should only be done for a very good reason (e.g. vastly more representative list), which is not the case of your PR.

PR #1408, opened 2018-01-18
Fix for Spot2 ORM EntityPopulator

michal-borek:
In Spot2 you don't need to create getId methods for your entities and Faker requires from you to have such. Changing it like this will not force developer to create getId methods in Spot2 entity classes.


fzaninotto:
Thanks!

michal-borek:
My pleasure :+1:

PR #1411, opened 2018-01-24
Added "blockchain" to en-us company provider catchPhrase method

samoldenburg:
In this day and age, this seems required.


fzaninotto:
👍

fzaninotto:
I wonder if it'd not be better un the bullshit words lists? 🤓

PR #1412, opened 2018-01-29
Normalize name of variable

eaglewu:
Invalid character

The character `ο` is not equal to `o`   //  it seems the same

`'ο'`   ---to int->  206 // Not Letter
`'o'`   ---to int->  111   // Letter 'o'


eaglewu:
image

fzaninotto:
Thanks

PR #1414, opened 2018-01-31
fix return type of year from string to int

JordiCabezudo:
this is just a possible fix, because year field is a int property but the return type of method year() is string


fzaninotto:
Sorry, that's a breaking change, I won't merge it into master.

PR #1416, opened 2018-01-31
Fix the problem with image load.

SiroDiaz:
The problem: Too many people requesting images to LoremPixel.
I have been searching for a similar website and i found RandImg that seems the most similar to LoremPixel.
Maybe this update suppose a big change for this release but i recommend you to consider this option as a valid starting point.


fzaninotto:
Sorry, I'm not interested in adding every image generator out there in Faker core. But feel free to create your own standalone package, and mention it in Faker's README.

PR #1419, opened 2018-02-02
Update PHPDoc of the DateTime provider.

### LINKED ISSUE #1418: the type of variable year doesn't match
the type of year field and year method doesn't match, and require a cast on my tests:

Generator.php

`@property int $year`

`@method string year($max = 'now')`

tomzx:
Should fix #1418.

I've gone over all the methods in the DateTime provider to make sure they were accurate.


fzaninotto:
Thanks!

PR #1421, opened 2018-02-09
Added randomHtmlFragment method to generate html strings without html…

dvdmarchetti:
Alternative solution to PR #1238 to generate random HTML blocks.


dmnc:
I was just looking at doing this also. It's really useful when using Faker to populate synthetic content in a CMS.

src/Faker/Provider/HtmlLorem.php
```diff
@@ -66,7 +66,25 @@ public function randomHtml($maxDepth = 4, $maxWidth = 4)
...
+        $this->idGenerator = new UniqueGenerator($this->generator);
+
+        $container = $document->createDocumentFragment();
+        $this->addRandomSubTree($container, $maxDepth, $maxWidth);
+
+        $document->appendChild($container);
+        return $document->saveHTML();
+    }
+
+    private function addRandomSubTree(\DOMNode $root, $maxDepth, $maxWidth)
```
pimjansen: Missing phpdoc

src/Faker/Provider/HtmlLorem.php
```diff
@@ -199,7 +217,7 @@ private function addRandomSpan(\DOMElement $element, $maxLength = 10)
         $element->appendChild($node);
     }
 
-    private function addLoginForm(\DOMElement $element)
+    private function addLoginForm(\DOMNode $element)
```
pimjansen: No php doc

PR #1422, opened 2018-02-09
Add YouTube provider link

aalaap:
Add link to the [aalaap/faker-youtube](https://github.com/aalaap/faker-youtube) provider.


fzaninotto:
Thanks. I suggest you don't use `str_shuffle` in your lib though, but Faker's `shuffle()` method instead. It supports seeding, and uses a better randomizer algorithm.

aalaap:
Brilliant! I was a little hesitant to use `str_shuffle` myself, but I didn't want to over-complicate things, so I stuck to basics. I didn't know about the `shuffle` method! I'm going to use it right away!

PR #1429, opened 2018-02-19
Add DateTime formatter for fr_FR and fr_BE

jmsche:


ppelgrims:
I'm not sure if I see any use in this? Why wouldn't you use the [locale](http://php.net/manual/en/function.setlocale.php) on your server with [stftime](http://php.net/manual/en/function.strftime.php)?
```php
$faker = Faker\Factory::create();
setlocale(LC_TIME, 'fr_BE.utf8');
echo strftime("%b", $faker->unixTime()); // mar
setlocale(LC_TIME, 'nl_BE.utf8');
echo strftime("%b", $faker->unixTime()); // mei
echo strftime("%a", $faker->unixTime()); // do
```
Or `$faker->randomElement` with the same arrays?

Anyway, I'm not the maintainer of this project but I reckon that a PR would get more easily accepted with unit tests.

pimjansen:
Indeed like @ppelgrims suggest there is actually no need to extend it in the core where it is relaying still a lot on timestamp (also due to older PHP support)

PR #1433, opened 2018-02-22
Facebook doesn't allow names with 3 or more consecutive identical characters

ktzouno:
Facebook rejects "Hilll" as a name because it has 3 consecutive identical characters.
In this PR I remove that name and add a unit test that tests all locales' Person classes' name-properties for such names.


pimjansen:
Im pretty ok in removing that name from the list if it is causing issues though. I dont know any usecase where 3 of the same letter could match a valid name.

However i think the tests where we are scanning with glob and running reflection on php code goes a bit too far

ktzouno:
Thanks for your answer. Removing it is good enough. I'd like to make sure that nobody adds it back again, but I'll survive :)

pimjansen:
@ktzouno could you rebase it with master so we can prep it for merge later?

ktzouno:
I had deleted my fork, so I opened a new pull request: https://github.com/fzaninotto/Faker/pull/1791

PR #1435, opened 2018-02-26
Remove provocative lastname

DrowningElysium:
One of our clients didn't like the name 'dick in son'.

Therefor the request to delete this name from the array


svanpoeck:
I sincerely hope you're joking.
Let alone for this reference : https://en.wikipedia.org/wiki/Emily_Dickinson

I'm not the maintainer of this tool though so I'll let it up to him to accept this PR or not.
But I'm firmly against it. Tell your client to seek help.

wtoalabi:
What makes you think I WANT the name in all my projects?
I guess it boils down to you forking the repo and do as you like with it....
Besides..this is a testing tool...hopefully you are not using it in
production to generate fake data?

DrowningElysium:
I fully agree that it is a normal name. But some people who don't have the knowledge about people with that last name will read it as 'dick in son' and will make problems of of it.

And no we used it as test data in a demo solution. Which we shouldn't have done it seems.

nextlevelshit:
The name is `Dickinson` and not `Dick in son`.

fzaninotto:
Dickinson is a usual and normal English name. End of discussion.

PR #1440, opened 2018-03-02
Removed unknown postcode

ayanozturk:
This postcode is unknown according to https://www.royalmail.com/find-a-postcode


LegendOfGIT:
Imo, it would be really helpful to add an expiration-date of fake information.
So you can generate fake information for items, which where created in a specific timespan (e.g. an purchase order)

fzaninotto:
thanks!

PR #1446, opened 2018-03-08
change PHPUnit namespace, set coverage setting

peter279k:
As title.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,6 @@
 language: php
 
-dist: precise
-
 php:
-    - 5.3
     - 5.4
     - 5.5
     - 5.6
@@ -12,7 +9,10 @@ php:
     - 7.2
     - nightly
 
-sudo: false
+matrix:
+  include:
+    - php: 5.3
+      dist: precise
 
 cache:
   directories:
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -9,4 +9,10 @@
             <directory>./test/Faker/</directory>
         </testsuite>
     </testsuites>
+
+    <filter>
+        <whitelist>
+            <directory suffix=".php">src/Faker/</directory>
+        </whitelist>
+    </filter>
 </phpunit>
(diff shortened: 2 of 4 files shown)


pimjansen:
@peter279k is this still required? Since we did some changes to the CI pipeline recently already.

peter279k:
Hi @pimjansen, thanks for your reply.

After looking at the [.travis.yml](https://github.com/fzaninotto/Faker/blob/master/.travis.yml) setting on master, it seems that we don't have to let all PHP versions be `precise` dist.

I think we can consider using the `matrix` setting (it's like my approach on this PR) to dispatch different dist for different PHP versions.

pimjansen:
@peter279k i dont really see a problem why they cant all be at precise though?

peter279k:
> @peter279k i dont really see a problem why they cant all be at precise though?

As far as I know, the pre-built PHP versions will be different on Travis CI build.

For example, the `precise` includes the `php-5.3`.

And the `trusty` includes the `php-5.4` to `php-7.3` possibly.

I'm not sure that the internal Travis CI build for PHP versions will be changed and now it's fine for this Travis CI setting at this moment.

Maybe Travis CI will drop the old dist (e.g. precise) in the future.

pimjansen:
@peter279k agree but i dont think we will much longer support those older versions where the builds probably will be executed in isolated docker containers anyways.

5.6 is already EOL and no longer supported so for future versions this will probably be dropped completely

peter279k:
If we drop unsupported PHP versions, we will consider using the `xenial` dist in the future.

pimjansen:
@peter279k well nothing is final actually so i would like to leave that as is for now where the builds run fine.

peter279k:
@pimjansen, thanks for your reply. I think it's time to close this PR until the Travis CI build is failed :).

pimjansen:
@peter279k no worries! Thanks for contributing and thinking with us

PR #1037, opened 2016-09-25
make sure in case of female pesel 10th number always divides by 2

Nuffic:

PR #1449, opened 2018-03-14
Fix Polish PESEL faker

Dartui:
Hi, polish PESEL faker has bug while generating PESEL for women.

10th digit in PESEL determines person sex. Even value mean woman, whether odd meas man. For now, tests sometimes fails, sometime passes. It's because of weird calculation using XOR.

This PR fixes this bug and add some tests to PESEL faker. In the past there were similar PR, but without tests: https://github.com/fzaninotto/Faker/pull/1037.

To run tests use this line: `phpunit --filter PersonTest test/Faker/Provider/pl_PL/PersonTest.php`


fzaninotto:
Thanks!

PR #1450, opened 2018-03-15
Fixing rare iin with 13-digits.

vadimonus:
According to documentations, when both checksum returns 10, Individual Identification Number cannot be used.
Example of such bad iin: Birth - 1989-04-02, Male, Population - 1584


fzaninotto:
Thanks

PR #1451, opened 2018-03-20
add birthplace support for ID nik generator

mikk150:
It turns out that Indonesian NIK has birthplace code in it.. add all valid codes to `Faker\Provider\id_ID\Person`


mikk150:
> Could we add some tests too?

@pimjansen yes

fzaninotto:
Thanks!

Issue #1452, opened 2018-03-21
[Bug] regexify does not allow nested alternations


mlambley:
Faker\Provider\Base::regexify does not recognise nested alternations, or literal brackets in those alternations. See the following:

![image](https://user-images.githubusercontent.com/17264636/37693515-ee7d1e8a-2d0b-11e8-94df-a7b44da531b9.png)

In the third example above, (b|c) should be resolved first, to let's say c, and then it needs to resolve (a|c).
In the fourth example above, the literal bracket inside the alternation should not be considered a closing bracket.

I will submit a PR shortly.

PR #1453, opened 2018-03-21
Allow nested alternations and literal brackets in regexify

mlambley:
See https://github.com/fzaninotto/Faker/issues/1452

I would appreciate some guidance around the test cases. On my machine, the mt_srand calls in Generator->seed completely mess up mt_rand, so phpunit fails tests left right and centre. 

From what I've seen, though, I haven't broken anything. Also, are the test cases I have included enough to cover my changes?

### DIFF
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -348,6 +348,20 @@ class Base
         return $string;
     }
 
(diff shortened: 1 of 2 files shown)


fzaninotto:
Thanks for your patch. 

PR #1454, opened 2018-03-22
Updated ES_es VAT provider, fixed control digit calculation

override2k:
Current ES_es provider faker->vat only generates a random control digit that causes strict vat (CIF) validators to fail.

This version generates a strict VAT number (society type + province + inscription number + control digit) that passes the validation libraries


juanma-lora-pepephone:
Useful fix. Could you merge it, pleae? Thanks!

PR #1455, opened 2018-03-23
floor return type is not integer

steros:
see http://php.net/manual/en/function.floor.php


pimjansen:
@steros nice find. Can we also add a test so we can ensure that a valid value is returned?

pimjansen:
Closing this one since it is stale

PR #1456, opened 2018-03-23
Added emailInDomain method

JGladykowski-modus:
Some corporate environments, require emails from specific domains.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -210,6 +210,7 @@ Methods accepting a `$timezone` argument default to `date_default_timezone_get()
     companyEmail            // 'russel.durward@mcdermott.org'
     freeEmailDomain         // 'yahoo.com'
     safeEmailDomain         // 'example.org'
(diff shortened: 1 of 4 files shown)


ppelgrims:
If you take in consideration how `$faker->freeEmailDomain`, `$faker->domainName`, or `$faker->safeEmailDomain` work in conjunction with the local-part, which you have to provide yourself, what is the added value?
Perhaps it makes more sense to make a local-part formatter that removes whitespaces, a trailing dot, and what not. And use that with your domain, e.g. `$faker->localPart . "@yourdomain.com"`, maybe introduce a parameter to output the `@` as well?
Maybe a combination of these two solutions is a better idea? See #935 as well.

@fzaninotto can probably weigh in on this

fzaninotto:
As @ppelgrims, I don't see much added value here.

PR #1457, opened 2018-03-26
Chancing to personal fictitious numbers

IngWARP:
Considering GDPR, introduced soon, one should use fictitious numbers

### DIFF
--- a/src/Faker/Provider/sv_SE/Person.php
+++ b/src/Faker/Provider/sv_SE/Person.php
@@ -115,6 +115,14 @@ class Person extends \Faker\Provider\Person
 
     /**
      * National Personal Identity number (personnummer)
+     * For example, a number of social security numbers are used in training databases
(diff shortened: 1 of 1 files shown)


IngWARP:
Hi Francois
What happens to this proposal?

fzaninotto:
I don't have a strong opinion on this, but I'm worried because it's the only place we do that. Very few formatters use a "reserved for tests" format - most use a "looks legit". If we opt for "reserved for tests", we should be more systematic about this.

Besides, I don't see what GDPR has to do with this.

IngWARP:
The testpart was added before my change I just changed the number so the test should run.
When I worked with GDPR, I have learned that it is not allowed to use real social security numbers

fzaninotto:
Then I think it should be a new option in the formatter. Please consider people using the formatter in unit tests: your change should not break their tests.

PR #1459, opened 2018-03-28
Fixing sk_SK men last names set

pavoltanuska:
In Slovak language, the suffix "-ová" belongs to female names only - both of these names' (Hečková, Jurinová) male counterparts are already part of the men set (Hečko, Jurina).

However, I'm not sure if I'm not breaking a part of Contributing guidelines with this PR, specifically:

If that's the case, feel free to close and not merge this PR :)


fzaninotto:
Thanks!

PR #1462, opened 2018-04-09
Turkish identity number for tr_TR

aykutaras:
Turkish identity number generator has been added to tr_TR locale's Person entity.
Also a new calculator for generating checksum and validating turkish identity number has been added.
I could not see any unit tests for tr_TR locale so I created some simple unit tests.

Lastly there was a failing unit tests in datetime unit tests which given static date is before current date. I changed that simply to one year later but it can be improved later.

### DIFF
--- a/test/Faker/Provider/DateTimeTest.php
+++ b/test/Faker/Provider/DateTimeTest.php
@@ -236,7 +236,7 @@ class DateTimeTest extends TestCase
 
     public function testFixedSeedWithMaximumTimestamp()
     {
-        $max = '2018-03-01 12:00:00';
+        $max = '2019-03-01 12:00:00';
 
         mt_srand(1);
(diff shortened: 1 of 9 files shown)


test/Faker/Provider/DateTimeTest.php
```diff
@@ -236,7 +236,7 @@ public function providerDateTimeInInterval()
 
     public function testFixedSeedWithMaximumTimestamp()
     {
-        $max = '2018-03-01 12:00:00';
+        $max = '2019-03-01 12:00:00';
```
ppelgrims: This was recently fixed, please remove it from the PR.

aykutaras: @ppelgrims Thank you. I have changed it now.

PR #1466, opened 2018-04-14
Update Address.php

Machou:
Update to news french region

### DIFF
--- a/src/Faker/Provider/fr_FR/Address.php
+++ b/src/Faker/Provider/fr_FR/Address.php
@@ -44,13 +44,26 @@ class Address extends \Faker\Provider\Address
     );
 
     private static $regions = array(
+        // Les treize régions métropolitaines, par ordre alphabétique de leurs noms et codes ISO 3166-2 :
+        'Auvergne-Rhône-Alpes', 'Bourgogne-Franche-Comté', 'Bretagne', 'Centre-Val de Loire', 'Corse',
+        'Grand Est', 'Hauts-de-France', 'Île-de-France', 'Normandie', 'Nouvelle-Aquitaine', 'Occitanie', 
+        'Pays de la Loire', 'Provence-Alpes-Côte d\'Azur', 
+
+        // Les cinq régions ultramarines, par ordre alphabétique des codes ISO 3166-2 :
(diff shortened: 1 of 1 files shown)


src/Faker/Provider/fr_FR/Address.php
```diff
@@ -44,13 +44,26 @@ class Address extends \Faker\Provider\Address
     );
 
     private static $regions = array(
+        // Les treize régions métropolitaines, par ordre alphabétique de leurs noms et codes ISO 3166-2 :
```
ppelgrims: Is there are source for this?

src/Faker/Provider/fr_FR/Address.php
```diff
@@ -44,13 +44,26 @@ class Address extends \Faker\Provider\Address
...
+        'Pays de la Loire', 'Provence-Alpes-Côte d\'Azur', 
+
+        // Les cinq régions ultramarines, par ordre alphabétique des codes ISO 3166-2 :
+        'Guadeloupe', 'Guyane (française)', 'Martinique', 'La Réunion', 'Mayotte',
+
+        // Autres régions
+        'Nouvelle-Calédonie', 'Polynésie française', 'Saint-Barthélémy', 'Saint-Martin', 'Saint-Pierre-et-Miquelon',
+        'Wallis-et-Futuna'
+    );
+        /*
```
ppelgrims: We're using git, so use `git diff` to see the difference between commits. You can safely remove the regions in the comments.

src/Faker/Provider/fr_FR/Address.php
Machou: New region in france in 2017
https://fr.wikipedia.org/wiki/R%C3%A9gion_fran%C3%A7aise

pimjansen:
@Machou bump, could we resolve the comments so we can merge?

src/Faker/Provider/fr_FR/Address.php
Machou: Sorry, but I don't understand what the problem is?

pimjansen:
I am going to close this PR as stale. If this still needs to be merged, please feel free to reopen it.
Thanks for contributing

PR #1469, opened 2018-04-18
Push the max date far into the future so the test can pass

rossmitchell:
There is a fixed date in the \Faker\Test\Provider\DateTimeTest::testFixedSeedWithMaximumTimestamp test that is set to 2018-03-01 12:00:00.

As this is now in the past it is causing the \Faker\Provider\DateTime::dateTimeThisMonth to throw an error which is causing the test to fail.

This increases the max date to 100 years in the future to avoid having to change this again, and allows the tests to pass once more


fzaninotto:
Thanks!

PR #1470, opened 2018-04-18
Update the parameters to check if the setter is callable

rossmitchell:
There was a change made in #1363 to the `\Faker\ORM\Doctrine\EntityPopulator::fillColumns` method to prevent errors if the setter was a private method.

The `method_exists` call was changed to `is_callable`, but the parameters passed to the call were not updated which means it will always return false.

This updates them to check if the setter is callable.

Tests will pass if #1469 if merged in


ppelgrims:
#1469 is merged, [opening and closing this PR should trigger a new build](https://github.com/travis-ci/travis-ci/issues/576#issuecomment-65317915) I didn't test this myself though.

rossmitchell:
Cheers, rebasing and pushing did the trick

fzaninotto:
Thanks!

PR #1473, opened 2018-04-23
Improve en_US Area Code Generation

shawnlindstrom:
The existing PhoneNumber provider does not fully implement the [NANP rules for area codes](https://www.nationalnanpa.com/area_codes/index.html). This can be problematic when running generated phone numbers against validation.

In concert with implementing the full NANP rules, this PR modifies randomDigitNot() such that it will accept an array as an argument. If you prefer, we could implement a distinct method such as randomDigitNotIn() instead. I did this because the first and second digits of the area code cannot be 37 or 96, so passing [3,7] or [9,6] was more convenient.


test/Faker/Provider/BaseTest.php
```diff
@@ -35,9 +35,26 @@ public function testRandomDigitNotReturnsValidDigit()
         }
     }
 
+    public function testRandomDigitNotReturnsValidDigitWhenCalledWithArray()
+    {
+        for ($i = 0; $i <= 9; $i++) {
+            $this->assertGreaterThanOrEqual(0, BaseProvider::randomDigitNot(array($i, $i + 1)));
```
fzaninotto: I don't understand these tests. Please make more obvious tests, with real life arrays.

shawnlindstrom:
@fzaninotto , thanks for the feedback! I've added the new randomDigitNotIn formatter and made the test more obvious. CI is currently breaking for Faker\Test\Provider\BaseTest::testRegexifySupportedRegexSyntax with data set #13 so I'll keep an eye and close/reopen when that's resolved.

shawnlindstrom:
@pimjansen I rebased locally but when I run the test suite, it's still failing for Faker\Test\Provider\BaseTest::testRegexifySupportedRegexSyntax with data set #13. I seem to recall this not happening for a separate PR I did. What should I do?

pimjansen:
@shawnlindstrom to be honest i have no idea where the actual master branch is running fine i expect a wrong rebase/merge or something broken in code though

shawnlindstrom:
Thanks @pimjansen. Yeah, I found the issue that was causing the test to fail. If it's all the same to you, I'm going to close this PR and open a new one. I must have messed something up as the rebase is not going smoothly at all. Thanks for your patience and assistance!

PR #1475, opened 2018-04-25
Fix ODM/ORM associations

### LINKED ISSUE #1474: ODM entity population with nullable fields
The Doctrine ODM EntityPopulator tries to detect optional association by looking at the `nullable` attribute of the association's metadata.

While this is a valid way to detect optional associations, it forces us actually set this attribute for optional associations which has unintended side effects.

# Example 1:
It is perfectly valid to set `$isSuccessorOf` to null, even if `ReferenceOne`/`EmbedOne` has no `nullable` attribute.

```php
class Product
{
    /**
     * @var string
     * @ODM\Id()
     */
    protected $id;

    /**
     * @var Product|null
     * @ODM\ReferenceOne(targetDocument="Meeva\ERP\CoreBundle\Document\Product")
     */
    protected $isSuccessorOf;
}
```

The result in MongoDB is an empty object:
```js
{
    "_id" : ObjectId("5ae097f315fe7b03f260f425")
}
```

# Example 2:
When being explicit about `nullable` the actual semantic seems to be more clear.

```php
class Product
{
    /**
     * @var string
     * @ODM\Id()
     */
    protected $id;

    /**
     * @var Product|null
     * @ODM\ReferenceOne(targetDocument="Meeva\ERP\CoreBundle\Document\Product", nullable=true)
     */
    protected $isSuccessorOf;
}
```

But the result in MongoDB is different to example 1:
```js
{
    "_id" : ObjectId("5ae097f315fe7b03f260f425"),
    "isSuccessorOf": null
}
```

# Conclusion
The DoctrineMongoDBBundle seems to use a different semantic of `nullable`:
- expected: `true` = field can be null / is optional, `false` = field must be set
- actual: `true` = set field to null if not given, `false` = do not persist field

The reason seems to be that MongoDB is NoSQL thus not enforcing any schema.

While this difference seems to be trivial, persisting millions of rows with empty values wastes space and we cannot simply switch all these attributes on/off just for testing purposes/data generation.

Probably checking `@var` for `null` is also feasible as an additional check?

buffcode:
Fixes optional association detection and removes uniqueness from associations (should be required via Generator if necessary).

Fixes #1474

### DIFF
--- a/src/Faker/ORM/Doctrine/EntityPopulator.php
+++ b/src/Faker/ORM/Doctrine/EntityPopulator.php
@@ -2,6 +2,7 @@
 
 namespace Faker\ORM\Doctrine;
 
+use Doctrine\Common\Annotations\AnnotationReader;
 use Doctrine\Common\Persistence\ObjectManager;
 use Doctrine\Common\Persistence\Mapping\ClassMetadata;
 
@@ -92,6 +93,7 @@ class EntityPopulator
      */
     public function guessColumnFormatters(\Faker\Generator $generator)
     {
+        $reader = new AnnotationReader();
         $formatters = array();
         $nameGuesser = new \Faker\Guesser\Name($generator);
         $columnTypeGuesser = new ColumnTypeGuesser($generator);
@@ -124,7 +126,6 @@ class EntityPopulator
                 foreach ($mappings as $mapping) {
                     if ($mapping['targetEntity'] == $relatedClass) {
                         if ($mapping['type'] == \Doctrine\ORM\Mapping\ClassMetadata::ONE_TO_ONE) {
-                            $unique = true;
                             $optional = isset($mapping['joinColumns'][0]['nullable']) ? $mapping['joinColumns'][0]['nullable'] : false;
                             break;
                         }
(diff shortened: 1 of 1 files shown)


buffcode:
Test failure is due to `testFixedSeedWithMaximumTimestamp` and not related to this PR.

src/Faker/ORM/Doctrine/EntityPopulator.php
```diff
@@ -92,6 +93,7 @@ public function mergeModifiersWith(array $modifiers)
      */
     public function guessColumnFormatters(\Faker\Generator $generator)
     {
+        $reader = new AnnotationReader();
```
fzaninotto: unused

src/Faker/ORM/Doctrine/EntityPopulator.php
```diff
@@ -135,8 +136,7 @@ public function guessColumnFormatters(\Faker\Generator $generator)
                 foreach ($mappings as $mapping) {
                     if ($mapping['targetDocument'] == $relatedClass) {
                         if ($mapping['type'] == \Doctrine\ODM\MongoDB\Mapping\ClassMetadata::ONE && $mapping['association'] == \Doctrine\ODM\MongoDB\Mapping\ClassMetadata::REFERENCE_ONE) {
-                            $unique = true;
-                            $optional = isset($mapping['nullable']) ? $mapping['nullable'] : false;
+                            $optional = preg_match("/\*\s*@var\s+(?:[^|]+\|)*?null(\|[^|]+)*$/im", $this->class->reflFields[$fieldName]->getDocComment()) === 1;
```
fzaninotto: please document and test this regexp

pimjansen:
Bump!

pimjansen:
Im going to close this issue because it is stale. Feel free to reopen if required later on.
Thanks for contributing

PR #1485, opened 2018-05-16
Remove the current nl-NL Text provider and replace it with more appropriate text

mlakerveld:
The current nl-NL text can be seen as inappropriate and not of this time.

The replacement text is also from Project Gutenberg and is a translation of Ali Baba and the forty thieves.


ppelgrims:
The problem which caused the tests to fail is fixed. You can build again by closing and reopening this PR, rebasing and pushing, or by empty committing to this branch.

mlakerveld:
Checks now pass, thanks for the tip!

fzaninotto:
Please find a text which is not a translation, but a text written natively in nl_NL.

PR #1486, opened 2018-05-17
fixed realText  generated broken korean text.

lesstif:
realText  generate broken korean text in some cases.

### DIFF
--- a/src/Faker/Provider/ko_KR/Text.php
+++ b/src/Faker/Provider/ko_KR/Text.php
@@ -4,6 +4,13 @@ namespace Faker\Provider\ko_KR;
 
 class Text extends \Faker\Provider\Text
 {
+    public function realText($maxNbChars = 200, $indexSize = 2)
+    {
+        $realText = parent::realText($maxNbChars, $indexSize);
+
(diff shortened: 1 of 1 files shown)


ppelgrims:
The problem which caused the tests to fail is fixed. You can build again by closing and reopening this PR, rebasing and pushing, or by empty committing to this branch.

src/Faker/Provider/ko_KR/Text.php
```diff
@@ -4,6 +4,13 @@
 
 class Text extends \Faker\Provider\Text
 {
+    public function realText($maxNbChars = 200, $indexSize = 2)
+    {
+        $realText = parent::realText($maxNbChars, $indexSize);
+
+        return iconv('UTF-8', 'UTF-8//IGNORE', $realText);
```
fzaninotto: This seems to be an admission of powerlessness. Instead of ignoring the bad characters, why not fix the bug that generates them?

lesstif: that's a good point.
To solve the problem, I will writing *realText* function for  Korean character and resending a PR.

thanks!

Issue #957, opened 2016-07-07
imageUrl lorempixel is down


linxlad:
Is there any chance to introduce a backup when http://lorempixel.com/ is down?

Newman101:
Again? We could always make an additional check to source code where alternative (such as http://placehold.it/) is used if the check fails.

fzaninotto:
On the contrary, I consider removing the image provider altogether, because I don't want to depend on a third-party with no SLA.

okj579:
I think a disclaimer to that effect would be sufficient. I still find the
provider very useful.

Though an external provider (like EmanueleMinotto/PlaceholdItProvider)
might be a better place for it?

damiandennis:
lorempixel seem to be always down at the moment. it is useful though please don't remove it.

atwright147:
How about letting the user select where the images come from rather than making the source an internal variable?

ethanclevenger91:
If you're looking for an immediate solution, I haven't run into this with FillMurray, either. It offers less flexibility, but you can replace the following function in `Providers/Image.php` (obviously this will be lost if you update/start up somewhere else):

```
public static function imageUrl($width = 640, $height = 480, $category = null, $randomize = true, $word = null, $gray = false)
    {
        // $baseUrl = "http://lorempixel.com/";
        $baseUrl = "http://www.fillmurray.com/";
        // $url = "{$width}/{$height}/";
        $url = "{$width}/{$height}";

        // if ($gray) {
        //     $url = "gray/" . $url;
        // }

        // if ($category) {
        //     if (!in_array($category, static::$categories)) {
        //         throw new \InvalidArgumentException(sprintf('Unknown image category "%s"', $category));
        //     }
        //     $url .= "{$category}/";
        //     if ($word) {
        //         $url .= "{$word}/";
        //     }
        // }

        // if ($randomize) {
        //     $url .= '?' . static::randomNumber(5, true);
        // }

        return $baseUrl . $url;
    }
```

gulaandrij:
#963 fix this issue

reno1979:
```public function robohash($name='robot',$width=200,$height=200,$background=FALSE,$secure=TRUE){

                    $baseUrl = ($secure) ? "https://robohash.org/" : "http://robohash.org";
                    $url = "{$name}.png?size={$width}x{$height}";
                    
                    if ($background) {
                        $url .= "&bgset=bg" . $this->randomDigit;
                    }
                   
[...]

PR #1489, opened 2018-05-21
Change image service to picsum.photos

kp-syginteractive:
Recently lorempixel.com has had lengthy downtimes and I have been using picsum.photos as a substitute. Please merge if you think this is appropriate for Faker generally.

If you want to go down a different solution path please let me know and I am happy to try implementing that.

Thanks for Faker btw, it is awesome :)

### DIFF
--- a/src/Faker/Provider/Image.php
+++ b/src/Faker/Provider/Image.php
@@ -30,7 +30,7 @@ class Image extends Base
      */
     public static function imageUrl($width = 640, $height = 480, $category = null, $randomize = true, $word = null, $gray = false)
     {
-        $baseUrl = "https://lorempixel.com/";
+        $baseUrl = "https://picsum.photos/";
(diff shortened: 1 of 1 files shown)


brunogaspar:
This is half working though as the options `category`, `word` and `gray` needs to be properly addressed.

Both the `category` and `word` doesn't seem to be, AFAIK, allowed on picsum.photos.

The `gray` is just a small change on the url so it becomes `https://picsum.photos/g/640/480/`

kp-syginteractiveAU:
Thanks for your input @brunogaspar. The issue with changing services is going to be the inconsistency of options. I normally use only the sizing options in my projects.

The choices here are not great - an unreliable service or a service with lesser/different options.

fzaninotto:
Thanks, but we won't implement an alternative image formatter in the core, see #957

PR #1493, opened 2018-05-24
create `passthrough` function

browner12:
This allows the user to use Faker "modifier" methods with values that are not Faker providers.

For example, I would like to generate an encrypted random string, and use the following code to do so.

```php
$encryptedRandomString = encrypt(str_random(32));
```

Now I would like to make this an optional value, using the Faker modifiers, which I don't believe is currently possible. With this new method, we can.

```php
$encryptedRandomString = $faker->optional()->passthrough(encrypt(str_random(32)));
```


browner12:
hmm..... that was... unexpected.  

I'm a little at a loss why this change would cause so many test breaks. Any thoughts?

ppelgrims:
The tests failed because of a DateTime problem which is fixed by now, and a problem with the image service that is being used.
If you introduce new functionality, please document it in the readme.

I think the name you used is very well picked!

readme.md
```diff
@@ -371,6 +371,12 @@ try {
 }
 ```
 
+If you would like to use a modifier with a value not generated by Faker, use the `passthrough()` method. `passthrough()` simply returns whatever value it was given.
+
+```php
+$faker->optional()->passthrough('my custom value');
```
fzaninotto: please use a better example, e.g. by using `mt_rand` as the custom value generator.

fzaninotto:
Thanks!

browner12:
glad to help

PR #1494, opened 2018-05-24
Fix failing dateTimeThisMonth test

mhlavac:
Not sure if this is perfect solution, but it seems to work for now.
It looks like the date that was picked was reached after some years and the test stopped working. This will always pick a month in the future for the dateTimeThisMonth.


ppelgrims:
Fixed by #1469

PR #1497, opened 2018-05-29
Add fake-car provider link

pelmered:
Full disclosure: I'm the creator of this provider. I hope someone finds it useful.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1661,6 +1661,7 @@ echo $faker->VAT; //23456789
 * [tamperdata/exiges](https://github.com/tamperdata/exiges): Faker provider for generating random temperatures
 * [jzonta/FakerRestaurant](https://github.com/jzonta/FakerRestaurant): Faker for Food and Beverage names generate
 * [aalaap/faker-youtube](https://github.com/aalaap/faker-youtube): Faker for YouTube URLs in various formats
+* [pelmered/fake-car](https://github.com/pelmered/fake-car): Faker for cars and car data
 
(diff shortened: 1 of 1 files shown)


ppelgrims:
I made a package just alike: [FakerCar](https://github.com/ppelgrims/FakerCar), at about the same time as well

pelmered:
@ppelgrims Interesting. There are a few basic car data faker providers that's been lying around for years, and now we created two separate providers the same couple of days.

fzaninotto:
Nice! May I suggest you guys join your efforts and merge your packages?

ppelgrims:
Aside from generating correct models per brand, this doesn't add much value to me:
1. Both door and seat count can be generated with `$faker->numberBetween`
1. Gearbox is just two values: `$faker->randomElement(['automatic', 'manual'])`
1. Vehicle registration number is basically `$faker->regexify` with a default value

pelmered:
@ppelgrims Well, the same could be said about power, torque and engineCapacity as well. It's just a number with a unit. You probably don't want to store the value as a string with unit in the database anyway as you probably would want that kind of data to be stored as an integer or float.

Seat counts, door counts and gearbox are also weighted. You you will for example get 4 doors 75% of the time.

I'm open to merge the packages, but I wouldn't mind if both providers was added neither.

ppelgrims:
@pelmered I agree and I'd prefer a merge, there is no need for two separate packages. How do we go about this?

PR #1508, opened 2018-06-16
[it_IT] Improve vat ID generated using official rules

mavimo:
SSIA.
Italy VAT Code should be:
 - 7 number (random)
 - 3 number (allowed values are city ISTAT code, that is a number between 1 and 121)
 - Luhn control code


fzaninotto:
Thanks!

PR #1511, opened 2018-06-20
Refactor INN Checksum

okj579:
Refactor INN checksum generator with PHP idioms and features to make it more readable.


fzaninotto:
Thanks, but I don't find this new version particularly easier to read.

PR #1514, opened 2018-06-26
Document randomDigitNot base formatter.

shawnlindstrom:
Adds missing documentation for the randomDigitNot base formatter.


fzaninotto:
Thanks!

PR #1515, opened 2018-06-29
fix regexify

cloud-wu:
Fix the bug that when using regexify function if the pattern contain escaping special characters in square brackets such like [\.\+A-z] it would return unmatched string.


cloud-wu:
@ppelgrims 

Thanks for your mention!

I  add the escape characters in `Faker\Test\Provider\BaseTest`

fzaninotto:
Thanks for your patch. 

PR #1516, opened 2018-07-03
Replaced Hilll with Hill

MarkVaughn:
I think this might have just been a human error somehow. 
I'm trying to detect spammy accounts that provide unrealistic names and expected my tests with valid names not to trip it up.
```
        $patterns = [
            '([a-z])\1\1', // Tripple letters
            'hjk',
            'ghj',
            'sdf',
            'zxc',
            'xvc',
            'qwe',
            '\d',
        ];
```
Occasionaly my unit test failed because Hilll was a provided surname. Searching the Providers it was the only case of triple letter other than the III (third) suffix.
image

https://en.wikipedia.org/wiki/Hill_(surname)

https://en.wikipedia.org/wiki/Hilll_(surname) < no such page
Even though there's 300K+ search results for hilll, it seems to be more of a typo. Heck typing hilll causes red squiggly lines making me want to correct it.


fzaninotto:
Thanks!

PR #1517, opened 2018-07-04
Add alphaHexColor

vincentchalamon:
Add hex color with alpha support: https://css-tricks.com/8-digit-hex-codes/


fzaninotto:
Thanks, but this PR adds too low added value. I think developers can do that very easily in userland.

PR #1522, opened 2018-07-09
Switch presentational b and i elements for semantic strong and em

dmnc:
Currently the body HTML generated by `randomHtml` contains elements that some validators object to.

This is slightly complicated again by HTML5 introducing new meaning for the `b` and `i` elements but switching [still looks sensible](http://html5doctor.com/i-b-em-strong-element/).

Merging #1421 too would make this even more useful.

### DIFF
--- a/src/Faker/Provider/HtmlLorem.php
+++ b/src/Faker/Provider/HtmlLorem.php
@@ -24,8 +24,8 @@ class HtmlLorem extends Base
     const UL_TAG = "ul";
     const LI_TAG = "li";
     const H_TAG = "h";
-    const B_TAG = "b";
-    const I_TAG = "i";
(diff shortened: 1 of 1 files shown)


ppelgrims:
Which validators?

I am a bit confused: if semantically they mean different things, why wouldn't we just introduce `strong` and `em`? Add them separately?

pimjansen:
Agree, <b> for example is a style definition and that is not what the HTML spec is about. That is the main reason why there is a <strong> type which indicates on how it should be understood.

I am a bit worried though what this will break for existing implementations

pimjansen:
We are not going to implement this for 1.9 but i do agree that we should have this done for 2.0.
Problem is that it is a breaking change for a lot of implementations

PR #1532, opened 2018-07-24
Return FrozenTime for CakePHP >= 3.3

OneHatRepo:
CakePHP 3.3 introduced the new FrozenTime class, which new Cake apps default to using, rather than the standard PHP DateTime class. This change follows CakePHP conventions and returns the FrozenTime class.

If you want to make it safer, you can first check for the existence of the Configure class:

 ```
                   $datetime = $generator->datetime();
                    if (class_exists('\Cake\Core\Configure') && floatval(\Cake\Core\Configure::version() >= 3.3)) {
                        return new \Cake\I18n\FrozenTime($datetime->format('Y-m-d H:i:s'));
                    }
                    return $datetime;
```

### DIFF
--- a/src/Faker/ORM/CakePHP/ColumnTypeGuesser.php
+++ b/src/Faker/ORM/CakePHP/ColumnTypeGuesser.php
@@ -56,7 +56,11 @@ class ColumnTypeGuesser
             case 'timestamp':
             case 'time':
                 return function () use ($generator) {
-                    return $generator->datetime();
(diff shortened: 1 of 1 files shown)


ppelgrims:
I'm not sure if the check is needed here, wouldn't the dependencies in Cake take care of this? E.g. below version 3.3, it would use the current Faker release? And from 3.3 it would use the next release.

PR #1533, opened 2018-07-26
Add a swiss social security number (AVS13) generator

nhedger:
Hey,

This is a random generator for swiss social security numbers (AVS13). I've made sure the code is correct using `make sniff` and ran the tests against the new code. It should all be good. I've also added an entry in the readme.md file to document the new provider.

Any feedback is welcome.

Regards.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1092,6 +1092,14 @@ echo $faker->nationalId; // 'V11223344'
 echo $faker->taxpayerIdentificationNumber; // 'J1234567891'
 ```
 
+### `Faker\Provider\fr_CH\Person`
+```php
+<?php
+
+// Generates a random AVS13/AHV13 social security number
+echo $faker->navs13; // "756.1234.5678.97"
+```
+
 ### `Faker\Provider\fr_FR\Address`
 
 ```php
--- a/readme.md
+++ b/readme.md
@@ -104,6 +104,7 @@ Each of the generator properties (like `name`, `address`, and `lorem`) are calle
 
     randomDigit             // 7
     randomDigitNotNull      // 5
+    randomDigitsString($length = 1) // '0325698745'
     randomNumber($nbDigits = NULL, $strict = false) // 79907610
     randomFloat($nbMaxDecimals = NULL, $min = 0, $max = NULL) // 48.8932
     numberBetween($min = 1000, $max = 9000) // 8567
--- a/readme.md
+++ b/readme.md
@@ -859,6 +859,15 @@ echo $faker->cvr; // "32458723"
 echo $faker->p; // "5398237590"
 ```
 
+### `Faker\Provider\de_CH\Person`
+```php
+<?php
+
+// Generates a random AVS13/AHV13 social security number
+echo $faker->avs13; // "756.1234.5678.97" OR
+echo $faker->ahv13; // "756.1234.5678.97"
+```
+
 ### `Faker\Provider\de_DE\Payment`
 
 ```php
@@ -1097,7 +1106,7 @@ echo $faker->taxpayerIdentificationNumber; // 'J1234567891'
 <?php
 
 // Generates a random AVS13/AHV13 social security number
-echo $faker->navs13; // "756.1234.5678.97"
+echo $faker->avs13; // "756.1234.5678.97"
 ```
 
 ### `Faker\Provider\fr_FR\Address`
@@ -1198,6 +1207,14 @@ echo $faker->bankAccountNumber; // "HU09904437680048220079300783"
 // first argument is gender, either Person::GENDER_MALE or Person::GENDER_FEMALE, if none specified random gender is used
 // second argument is birth date (DateTime object), if none specified, random birth date is used
 echo $faker->nik(); // "8522246001570940"
+```
+
+### `Faker\Provider\it_CH\Person`
+```php
+<?php
+
+// Generates a random AVS13/AHV13 social security number
+echo $faker->avs13; // "756.1234.5678.97"
 ```
 
 ### `Faker\Provider\it_IT\Company`
--- a/readme.md
+++ b/readme.md
@@ -104,7 +104,6 @@ Each of the generator properties (like `name`, `address`, and `lorem`) are calle
 
     randomDigit             // 7
     randomDigitNotNull      // 5
-    randomDigitsString($length = 1) // '0325698745'
     randomNumber($nbDigits = NULL, $strict = false) // 79907610
     randomFloat($nbMaxDecimals = NULL, $min = 0, $max = NULL) // 48.8932
     numberBetween($min = 1000, $max = 9000) // 8567
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -63,6 +63,33 @@ class Base
     }
 
     /**
+     * Generates a string of random digits
+     *
+     * @param int $length Length of the string to generate
+     * @return string A string of {@see $length} random digits
+     *
+     * @throws \OutOfBoundsException
+     * @throws \InvalidArgumentException
+     */
+    public static function randomDigitsString($length = 1)
+    {
+        if (!is_numeric($length)) {
+            throw new \InvalidArgumentException('the length parameter must be numeric');
+        }
+
+        if ($length <= 0) {
+            throw new \OutOfBoundsException('the length must be greater than 0');
+        }
+
+        $digits = "";
+        for ($i = 0; $i < $length; $i++) {
+            $digits .= mt_rand(0, 9);
+        }
+
+        return $digits;
+    }
+
+    /**
      * Returns a random integer with 0 to $nbDigits digits.
      *
      * The maximum value returned is mt_getrandmax()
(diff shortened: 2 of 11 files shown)


src/Faker/Provider/fr_CH/Person.php
```diff
@@ -87,4 +87,87 @@ class Person extends \Faker\Provider\fr_FR\Person
...
+     *
+     * By default, if no {@see $length} parameter is given, the function will
+     * return a one-digit string.
+     *
+     * @since 0.1.0
+     * @param int $length Length of the string of digits to generate
+     * @return string A string of {@see $length} random digits
+     * @throws \OutOfBoundsException
+     */
+    protected static function generateRandomDigits($length = 1)
```
ppelgrims: I reckon this kind of method should go into `Faker/Provider/Base` but you probably could get the same result using `numberBetween`

src/Faker/Provider/fr_CH/Person.php
```diff
@@ -87,4 +87,87 @@ class Person extends \Faker\Provider\fr_FR\Person
...
+    /**
+     * Computes the checksum of the first 12 digits of an AVS13 number.
+     *
+     * The checksum is computed exactly like the sum of an EAN13 number.
+     *
+     * @see https://fr.wikipedia.org/wiki/EAN_13
+     * @param string $digits
+     * @return int
+     */
+    protected static function computeAvs13Checksum($digits)
```
ppelgrims: I think this would probably fit better in the `Faker/Calculator` namespace, instead of being buried in a localized provider.

src/Faker/Provider/fr_CH/Person.php
```diff
@@ -87,4 +87,87 @@ class Person extends \Faker\Provider\fr_FR\Person
...
+    /**
+     * Generates a valid random AVS13 (swiss social security) number
+     *
+     * This function will generate a valid random AVS13 number and return it
+     * as a formatted string.
+     *
+     * @see https://www.zas.admin.ch/zas/fr/home/partenaires-et-institutions-/navs13.html
+     * @return string
+     */
+    public function navs13()
```
ppelgrims: This should be a static method

readme.md
```diff
@@ -1092,6 +1092,14 @@ echo $faker->nationalId; // 'V11223344'
 echo $faker->taxpayerIdentificationNumber; // 'J1234567891'
 ```
 
+### `Faker\Provider\fr_CH\Person`
```
ppelgrims: Shouldn't this be available in `it_CH` and `de_CH` as well?

src/Faker/Provider/fr_CH/Person.php
ppelgrims: Could you provide a separate test for this as well?

test/Faker/Provider/fr_CH/PersonTest.php
```diff
@@ -0,0 +1,25 @@
...
+    {
+        $faker = new Generator();
+        $faker->addProvider(new Person($faker));
+        $this->faker = $faker;
+    }
+
+	public function testAvsNumber()
+    {
+		$avs = $this->faker->navs13;
+		$this->assertRegExp('/^756\.([0-9]{4})\.([0-9]{4})\.([0-9]{2})$/i', $avs);
```
ppelgrims: Does it make sense to do an assertion about the checksum here? Would that suffice to check the whole number?

src/Faker/Provider/fr_CH/Person.php
nhedger: Great idea. Would it be better to create an Ean13 utility class since it's using the same algorithm ?

src/Faker/Provider/fr_CH/Person.php
nhedger: Thanks, will update that.

readme.md
nhedger: Absolutely, I completely forgot about that. How would you suggest I implement that ? I would like to avoid duplicating the code as much as possible. Should I move all the AVS13 number generation out of the localized provider ?

test/Faker/Provider/fr_CH/PersonTest.php
nhedger: Indeed, a checksum verification should be enough to assert the number is valid.

readme.md
ppelgrims: We're following the PSR-4 standard to auto-load classes. They're static methods, so I'd say you can just call them?

src/Faker/Provider/fr_CH/Person.php
ppelgrims: There is a Luhn (etc.) calculator class, that does sound like a good idea to group it like that.

test/Faker/Provider/fr_CH/PersonTest.php
nhedger: Actually, after a bit of thinking, I believe it's still pertinent to assert the format of the generated number since the format is part of the specification. Sure, the checksum would confirm that the number is a valid EAN-13 number but it could validate any EAN-13 that is NOT an AVS13 number.

nhedger:
Here a brief of the changes I've made :

- Created a new **randomDigitsString** method inside `Faker/Provider/Base`
- Created a new `Ean` class under `Faker/Calculator`. This class allows the validation of EAN-8 and EAN-13 numbers. It is used internally to generate the checkum for the AVS13 number.
- Created alias functions inside the `de_CH` and `it_CH` providers and left the main code inside the `fr_CH` provider.
- Set the methods inside the providers as static, as requested.
- Wrote the tests for everything

nhedger:
Any updates regarding this PR ?

src/Faker/Provider/Base.php
```diff
@@ -62,6 +62,33 @@ public static function randomDigitNot($except)
...
+    /**
+     * Generates a string of random digits
+     *
+     * @param int $length Length of the string to generate
+     * @return string A string of {@see $length} random digits
+     *
+     * @throws \OutOfBoundsException
+     * @throws \InvalidArgumentException
+     */
+    public static function randomDigitsString($length = 1)
```
fzaninotto: how is that different from `numerify('#######')`?

nhedger: I wasn't aware this function existed actually, which is why I made my own. It appears to work similarly and happens to work like intended so I guess I could remove my function. Should I commit my changes ?

fzaninotto: yes, please remove your function from the PR

nhedger: Done

Ph0tonic:
Hi, any news about this PR?

nhedger:
Hi, @Ph0tonic, thanks for your comment. I was waiting for a validation on this PR but I completely forgot about it. @fzaninotto it would be much appreciated if you could take a look at it.

Ph0tonic:
Hi,
Is there any change to make? @fzaninotto I would be gratefull if you could have a look at this PR.

test/Faker/Provider/de_CH/PersonTest.php
```diff
@@ -0,0 +1,34 @@
...
+    private $faker;
+
+    public function setUp()
+    {
+        $faker = new Generator();
+        $faker->addProvider(new Person($faker));
+        $this->faker = $faker;
+    }
+
+	public function testAvs13Number()
```
fzaninotto: indentation seems broken in this file, can you please check

test/Faker/Provider/fr_CH/PersonTest.php
```diff
@@ -0,0 +1,27 @@
...
+    private $faker;
+
+    public function setUp()
+    {
+        $faker = new Generator();
+        $faker->addProvider(new Person($faker));
+        $this->faker = $faker;
+    }
+
+	public function testAvsNumber()
```
fzaninotto: same, seing indentation problems

test/Faker/Provider/it_CH/PersonTest.php
```diff
@@ -0,0 +1,27 @@
...
+    private $faker;
+
+    public function setUp()
+    {
+        $faker = new Generator();
+        $faker->addProvider(new Person($faker));
+        $this->faker = $faker;
+    }
+
+	public function testAvs13Number()
```
fzaninotto: here, too

readme.md
```diff
@@ -104,6 +104,7 @@ Each of the generator properties (like `name`, `address`, and `lorem`) are calle
 
     randomDigit             // 7
     randomDigitNotNull      // 5
+    randomDigitsString($length = 1) // '0325698745'
```
fzaninotto: please remove this line

test/Faker/Provider/de_CH/PersonTest.php
nhedger: Done.

test/Faker/Provider/fr_CH/PersonTest.php
nhedger: Done also.

test/Faker/Provider/it_CH/PersonTest.php
nhedger: Here too :)

readme.md
nhedger: Line removed !

pimjansen:
Thanks for contributing!

PR #1535, opened 2018-07-26
Drop support for legacy PHP versions

simPod:
Hello,

when worked on a new PR trying to implement new generator, I stumbled upon failing builds on legacy PHP versions.

We should drop support for legacy PHP versions in new major release.

secure.php.net/supported-versions.php

By the end of 2018 security support for 5.6 and 7.0 ends. 5.5 and lower are not worth mentioning.

Let's move on. PHP 7.3 will be out soon, PHP 8 is already being [actively discussed](https://twitter.com/nikolaposa/status/1018966511388647429).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,6 @@ language: php
 dist: precise
 
 php:
-    - 5.3
-    - 5.4
-    - 5.5
-    - 5.6
-    - 7.0
     - 7.1
     - 7.2
     - nightly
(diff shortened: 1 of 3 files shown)


fzaninotto:
Thanks, but no new major release is on the horizon

simPod:
Why not make this a major release then? Legacy users can still use old versions but library can progress.

Issue #1352, opened 2017-11-14
Missing UUIDv4


enumag:
`$faker->uuid` returns UUID v3. In my application I need UUIDv4. Could a method like `$faker->uuid4` be added for that?

fzaninotto:
Great idea! Feel free to open a PR for that.

PR #1536, opened 2018-07-26
UUID 4 Generator

simPod:
Adds generator for UUID v4 as the current `uuid()` function generates v3 that is not so widely used nowadays.

Based on https://github.com/fzaninotto/Faker/pull/1535, solves https://github.com/fzaninotto/Faker/issues/1352

### DIFF
--- a/src/Faker/Provider/Uuid.php
+++ b/src/Faker/Provider/Uuid.php
@@ -54,4 +54,13 @@ class Uuid extends Base
 
         return $uuid;
     }
+
+    public static function uuid4()
+    {
+        $data = openssl_random_pseudo_bytes(16);
+        $data[6] = chr(ord($data[6]) & 0x0f | 0x40); // set version to 0100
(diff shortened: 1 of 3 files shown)


fzaninotto:
Can you make a PHP 5.3 compatible PR?

src/Faker/Provider/Uuid.php
```diff
@@ -54,4 +54,18 @@ public static function uuid()
 
         return $uuid;
     }
+
+    /**
+     * Generation code from https://gist.github.com/JanTvrdik/655b31e78e998bb468c7f8f8cfe4b3b6
+     */
+    public static function uuid4() : string
+    {
+        $uuidBin = random_bytes(18);
```
fzaninotto: not good. You have to use the `mt_` random generation functions to allow seeding, or the generators from the `Base` object.

simPod:
Tried but no, cannot. I don't usually develop for legacy versions and don't know proper syntax. https://github.com/fzaninotto/Faker/pull/1535 was closed, can't do much here. We can probably close this one.

PR #1537, opened 2018-07-29
Portuguese providers for Companies and Internet

promatik:
Added the most common terminology for portugueses companies, the most common free mail domains and top level domains.


fzaninotto:
Thanks!

PR #1538, opened 2018-08-01
Update fake information in Vietnam

chiendevit:
I am Vietnamese and I get information from websites that I trust. Some new information (like change of head phone number) is still too few sites talk about it, I added the source


src/Faker/Provider/vi_VN/PhoneNumber.php
```diff
@@ -5,24 +5,29 @@
...
-        26, 710, 511, 500, 510, 230,
-        61, 67, 59, 219, 351, 4,
-        39, 320, 31, 711, 218, 321,
-        8, 58, 77, 60, 231, 25,
-        20, 63, 72, 350, 38, 30,
-        68, 210, 57, 52, 510, 55,
-        33, 53, 79, 22, 66, 36,
-        280, 37, 54, 73, 74, 27,
-        70, 211, 29,
+        // Updated new area code on 17/06/2018
```
ppelgrims: Could this be moved to the DocBlock?

src/Faker/Provider/vi_VN/PhoneNumber.php
```diff
@@ -5,24 +5,29 @@
...
+        // http://vietnamnet.vn/vn/cong-nghe/vien-thong/danh-sach-ma-vung-dien-thoai-moi-cac-tinh-thanh-355763.html
```
ppelgrims: Does this source comply with the license of this package?

src/Faker/Provider/vi_VN/PhoneNumber.php
```diff
@@ -5,24 +5,29 @@
...
+        232, 235, 255, 203, 233, 299,
+        212, 237, 227, 208, 234, 273,
+        294, 207, 276, 270, 216,
+        
         // Mobile
-        96, 97, 98, 162, 163, 164, 165, 166, 167, 168, 169, // Viettel
+        86, 96, 97, 98, 162, 163, 164, 165, 166, 167, 168, 169, // Viettel
         91, 94, 123, 124, 125, 127, 129, // Vinaphone
-        90, 93, 120, 121, 122, 126, 128, // Mobifone
+        // Mobifone --> The network has ceased to exist (fewer users are also considered non-existent)
```
ppelgrims: Removing this could break backward compatibility for other users.

src/Faker/Provider/vi_VN/PhoneNumber.php
```diff
@@ -5,24 +5,29 @@
...
         92, 186, 188, // Vietnamobile
-        99, 199, // Gmobile
-        95, // Sfone
+        // Gmobile --> The network has ceased to exist (fewer users are also considered non-existent)
```
ppelgrims: See previous comment, new releases are minor version releases, which are backwards-compatible.

src/Faker/Provider/vi_VN/PhoneNumber.php
```diff
@@ -5,24 +5,29 @@
...
+        // Sfone --> The network has ceased to exist (fewer users are also considered non-existent)
```
ppelgrims: See previous comment

src/Faker/Provider/vi_VN/Internet.php
```diff
@@ -4,5 +4,346 @@
...
+        'gov',
+        'int',
+        'jobs',
+        'mil',
+        'mobi',
+        'museum',
+        'tel',
+        'travel',
+        
+        'ac',
```
ppelgrims: Please group this alphabetically, e.g. 80 characters on a line so the `git diff` is still readable. See area codes beneath

ppelgrims: I'm not sure if you want to provide all ccTLD? Why not just stick to Vietnamese and gTLD?

src/Faker/Provider/vi_VN/PhoneNumber.php
ChienDevIT: yes

src/Faker/Provider/vi_VN/PhoneNumber.php
ChienDevIT: This source of information is reliable by the state, it is a page that the number of vietnamese access large. And this source of information is shared and reused everywhere

src/Faker/Provider/vi_VN/PhoneNumber.php
ChienDevIT: We should keep the information first, but add a warning later
``
99, 199, // Gmobile --> The network has ceased to exist (fewer users are also considered non-existent)
``

src/Faker/Provider/vi_VN/PhoneNumber.php
ChienDevIT: We should keep the information first, but add a warning later

``
90, 93, 120, 121, 122, 126, 128, // Mobifone --> The network has ceased to exist (fewer users are also considered non-existent)
``

src/Faker/Provider/vi_VN/PhoneNumber.php
ChienDevIT: We should keep the information first, but add a warning later
``

ChienDevIT: We should keep the information first, but add a warning later
``
95, // Sfone --> The network has ceased to exist (fewer users are also considered non-existent)
``

src/Faker/Provider/vi_VN/Internet.php
ChienDevIT: A very good idea. I will try to modify it in near future!

PR #1541, opened 2018-08-05
persian(farsi) lorem

MahdiAbedi:
Persian language was added to lorem


fzaninotto:

MahdiAbedi:
sorry,i don't understant you,why we cannot change lorem file,i just added persian words instead of british's
it's very usefull for iranin developers

PR #1542, opened 2018-08-09
Multiple APIs for faking images

spacedog4:
lorempixel is out of air so I made some change in Faker Image class so I can pass a param and change the base url to loremflickr that works perfectly


loic425:
This pull request solves that issue #1543

loic425:
@spacedog4 All faker image api don't have same path system.
https://picsum.photos/

With this one, you need to add random query parameter.
imho, you should restrict base url to a compatible list.

loic425:
@spacedog4 another feature could be to automatically switch api when one is down.

spacedog4:
@loic425 I've just pushed an update, now you pass the name of the api you want to use, restricted by four, loremixel, loremflickr, picsumphotos and placeholder, each one have a extends a contract to generate a the URL, it means if anyone wants to add a new api just need to create anothe class that extends this contract and create the imageUrl function. If one api is unsuccessful it automatically tries to use the next one #

loic425:
@spacedog4 ok that's far better :)  Now, waiting for a core team member to review that. I can request you some changes but a core team mmeber review seems to be more appropriate.

loic425:
@spacedog4 But you can already fix some codestyle errors looking on travis logs.

src/Faker/Provider/Image.php
```diff
@@ -7,9 +7,14 @@
  */
 class Image extends Base
 {
-    protected static $categories = array(
-        'abstract', 'animals', 'business', 'cats', 'city', 'food', 'nightlife',
-        'fashion', 'people', 'nature', 'sports', 'technics', 'transport'
+
+    protected static $apis = array(
+        'lorempixel'   => ImageContracts\LoremPixel::class,
```
loic425: It's not php5 compatible. Php5 builds fail because of these following lines.

fzaninotto:
I've already said that I won't accept such PRs. I want to remove the image generator, not to add more. See #957 for the rationale.

PR #1547, opened 2018-08-12
Replace LoremPixel with Unsplash

nycooookie:
LoremPixel is down once again, so I thought I'd try to replace it with Unsplash Source.

This would however add breaking changes:
- No categories (instead they use search keywords)
- No grayscale
- No text overlay
- Images are always random

Therefore the method signature would change from

```php
public static function imageUrl($width = 640, $height = 480, $category = null, $randomize = true, $word = null, $gray = false)
```

to

```php
public static function imageUrl($width = 640, $height = 480, $keywords = null)
```
For the `image` method accordingly.

---

You mentioned you would rather drop image support all together, however, for me and probably many other developers, image generation is a valuable part of the package.

Please let me know if you're not interested or if there's something horribly wrong with the PR (it's my first PR, so I'd be glad for feedback!).

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -261,16 +261,13 @@ Methods accepting a `$timezone` argument default to `date_default_timezone_get()
 
 ### `Faker\Provider\Image`
 
-    // Image generation provided by LoremPixel (http://lorempixel.com/)
-    imageUrl($width = 640, $height = 480) // 'http://lorempixel.com/640/480/'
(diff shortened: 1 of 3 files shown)


fzaninotto:
I've already said that I won't accept such PRs. I want to remove the image generator, not to add more. See #957 for the rationale.

PR #1553, opened 2018-08-15
fixed the randomFloat problem

ivancli:
there was a slight chance where randomFloat generates a float smaller than $min when $min was 0.01 and $max was null, result was 0.


ivancli:
Test case has been added. Please check

PR #1558, opened 2018-08-18
Suggest to install this library as a dev dependency

bocharsky-bw:
I think in most cases devs use this library in development, not on production

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -46,7 +46,7 @@ Faker requires PHP >= 5.3.3.
 ## Installation
 
 ```sh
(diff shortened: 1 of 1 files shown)


fzaninotto:
Thanks, but I think developers are mature enough to know if they want it in dev or production?

bocharsky-bw:
Well, I just blindly copied/pasted this command and forgot to add `--dev` - understood it when saw `git diff`... that's why I created this PR - the idea is to help a bit devs. Copy only part of the command easier than write additional options anyways :)

PR #1561, opened 2018-08-20
Make $suffix protected not private

SpadXIII:
When extending \Faker\Provider\nl_NL\Person $suffix cannot be used because it is private. Small fix to make it protected like the other properties.


fzaninotto:
Thanks!

PR #1565, opened 2018-08-22
Add fr_BE national number generator

miclf:
This feature has actually already been implemented in the `nl_BE/Person` provider by @ppelgrims in PR #1361.

Since the output required for an `fr_BE/Person` provider would be exactly the same (as the Belgian database of national numbers does not care about languages), to avoid duplicating code, the method added by this PR simply “proxies” the call to the original implementation in the `nl_BE` provider.
This seemed like the most logical thing to do. Please tell me if, for any reason, code duplication would be preferable.

This PR also adds the provider in the *[Language specific formatters](https://github.com/fzaninotto/Faker#language-specific-formatters)* section of the README.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1074,6 +1074,15 @@ echo $faker->vat;           // "BE 0123456789" - Belgian Value Added Tax number
 echo $faker->vat(false);    // "BE0123456789" - unspaced Belgian Value Added Tax number
 ```
 
+### `Faker\Provider\fr_BE\Person`
+
+```php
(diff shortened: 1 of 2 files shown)


ppelgrims:
Perfect, I completely forgot about it until yesterday when reviewing another PR, but I didn't got a change to get to it. Nice catch!

miclf:
Well, thanks for your work. I think I couldn’t have done an implementation as efficient as yours :)

PR #1568, opened 2018-08-24
Return an it_IT mobile phone number

probfacile:
Extending provider it_IT to generate a mobile phone number (only starting with 3 and 10 digits) almost as in en_HK


ppelgrims:
Please close and reopen the PR in order for Travis to rerun and pass all tests.

PR #1574, opened 2018-09-06
Adding sin number function for Canadian persons

mpge:
Some code leveraged from https://github.com/corbanworks/fng-sin-tools/blob/master/fngsin.class.php


mpge:
@ppelgrims good point! I didn't even see the Luhn class, apologies!

I'll rewrite it using that (since it will be a lot less code). Closing the PR for now until updates are made.

PR #1575, opened 2018-09-06
Prevent Transliterator class autoloading

MontealegreLuis:
Autoloading breaks in Yii 1 applications when checking if the `Transliterator` class exist.

I disabled autoloading in the call to `class_exists`


tance77:
Great patch thank you!

localheinz:
Thank you, @MontealegreLuis!

PR #1578, opened 2018-09-15
Old city names changed to new ones

goszowski:


fzaninotto:
can you explain the change?

goszowski:
@fzaninotto The listed settlements no longer exist in our country. I replaced them with new, valid names.

fzaninotto:
Thanks then!

PR #1580, opened 2018-09-17
Fix Dutch typo

gizburdt:
I saw my surname in the faked data, but it is misspelled. The misspelled version doesn't exist in The Netherlands.


gizburdt:
All clear, thanks for the reply :)

fzaninotto:
@ppelgrims I think this change is valid. It will only break tests in very rare cases (not all cases based on Dutch names). So I'm going to merge it. Thanks!

ppelgrims:
Yeah, I'm not sure how strict this is to be enforced. Anyway, nice work @gizburdt !

PR #1583, opened 2018-09-19
Fix dateTimeBetween() Argument on Generator phpdoc

hisomura:
I'm sorry for my poor English.

Faker\Provider\DateTime::dateTimeBetween() has $timezone argument.
But dateTimeBetween() on Generator phpdoc does not have this argument.
I added it.


fzaninotto:
Thanks.

PR #1588, opened 2018-09-26
The address provider fot the Slovenian language had a wrong address s…

techouse:
…tring printed. I fixed it and wrote some additional tests for the Slovenian language providers.


pimjansen:
Even though i agree on this change, it will be breaking due the fact we change the formatting of the locale. The \n was not respected for a long time so don't think it is wise to implement this for a 1.x release

techouse:
Yes, I see what you mean. :(

PR #1590, opened 2018-10-01
Refactoring of the TextTest class

stevegrunwell:
This PR applies some minor refactoring to the `TextTest` class, making it easier to understand what's happening:

1. Add a new `buildGenerator()` method, which creates a new instance of the text generator at the start of each test and saves it in the private `$generator` property — this saves a few lines of setup for each test method.
2. Instead of a `foreach` loop within `testTextMaxLength()`, use [PHPUnit's data providers](https://phpunit.de/manual/6.5/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers) to explicitly run each scenario as a separate test.
3. For test methods that are using `@expectException`, explicitly add a failure message to the test to describe _why_ the exception should have occurred.


pimjansen:
@stevegrunwell im pretty happy with the change. Is it possible to rebase it with latest so we can do some final checks?

stevegrunwell:
@pimjansen A fresh copy has been pushed. I've also moved the `@expectedException` annotations to `$this->setExpectedException()` as [the annotation version was deprecated in PHPUnit 8](https://github.com/sebastianbergmann/phpunit/issues/3332).

pimjansen:
@stevegrunwell builds are failing, please review

stevegrunwell:
@pimjansen Apologies, type-hints and `::class` resolution in some of the test methods were causing it to fail on old versions of PHP. Everything should be green, now.

fzaninotto:
Thanks!

PR #1592, opened 2018-10-02
add formatter for robot user agents

momorgan7:
Not sure if it should also be added to the random user agent method.


fzaninotto:
Hi, and thanks for your contribution. Crawlers aren't faked that often, so I'll pass on adding this formatter to the core. But you're welcome to publish it as a standalone package, and to add a link to it in Faker's README!

PR #1595, opened 2018-10-08
Fix method name for en_NZ provider in readme

gdhnz:
The method name in the en_NZ provider should be mobileNumber instead of cellNumber.


fzaninotto:
Thanks!

PR #1600, opened 2018-10-14
DateTime provider for pl_PL

jsawo:
added polish names for months and days of the week


pimjansen:
Please set the locale with PHP to get a valid day of the week or month. No need to do a custom implementation on the library it self

PR #1546, opened 2018-08-11
add gender variants for last name [ru_RU]

w1cerg:
Add stand-alone method for last name female

PR #1603, opened 2018-10-18
Fixed female lastName + added tests for ru_RU Person

iGusev:


iGusev:
This is a better variant of https://github.com/fzaninotto/Faker/pull/1546

Issue #1441, opened 2018-03-04
Null is considered unique value to Faker!


AmrIKhudair:
When using `optional()` in conjucation with `unique()`, a `null` is generated just once.

How to reproduce (reproduction rate of 100%):

    $faker = Faker\Factory::create();
    $faker->unique()->optional(0)->word; // null
    $faker->unique()->optional(0)->word; // throws OverflowException

BTW $faker->optional()->unique() yields `null` most of the time which I find a little bit awkward!

ppelgrims:
From the [readme](https://github.com/fzaninotto/Faker#modifiers):
```php
// optional() accepts a weight argument to specify the probability of receiving the default value.
// 0 will always return the default value; 1 will always return the provider. Default weight is 0.5 (50% chance).
```
```php
// optional() accepts a default argument to specify the default value to return.
// Defaults to NULL.
```
Does that help?

AmrIKhudair:
The problem is `null` is considered a value that is unique although `null` essentially means "undetermined value". Even databases that have nullable unique fields allow multiple instances of null.

ppelgrims:
I see what you're getting at, I had something else in mind when I wrote that reply.
What do you propose as a solution?

AmrIKhudair:
That the Faker\UniqueGenerator will refrain from checking uniqueness when the value is `null`.

marjetika:
I agree with @AmrIKhudair

jdferreira:
I know this is a bit old, but I support this change as well.

pimjansen:
@fzaninotto any reasonf or this since it actually functionality wise looks like a bug. However it could have bad sidefx when we will patch this since it will break probably

fzaninotto:
the two modifiers are independent, and there is no way to know that one is used in conjunction with the other. Therefore I don't think that we have a way to fix this. Do we?

pimjansen:
Correct, wanted to know for sure. Think we can highlight this a little in the docs for the end users then in this case

pimjansen:
Did some more digging and actually what `faker` does here is 100% correct. If you ask for an unique record then `null` is still a valid record and therefor not unique. 

I do not really support the idea of skipping `null` on uniqueness where its a valid value itself.

jdferreira:
I still feel that  the Faker library should support faking a unique value that is also nullable.

Consider the database schema below:

```mysql
CREATE TABLE client (
[...]

PR #1610, opened 2018-10-30
Ignore null values when watching uniqueness

davidporos92:
This PR is intended to "fix" issue #1441


pimjansen:
I do not support this feature. The value `null` is a valid value and should be taken into account when an unique value is required. Hence if you allow this the other side effect would be that it generates multiple null values and therefor have an issue on the other side.

fzaninotto:
I agree with @pimjansen. Closing the PR.

PR #1614, opened 2018-11-08
Suggest PSR-4 compliant autoloader in comment

samnela:
- PSR-0 has been marked as deprecated , PSR-4 is now recommended as an alternative. https://www.php-fig.org/psr/psr-0/
 - the ClassLoader component entirely is no longer maintained. https://github.com/symfony/class-loader


samnela:
Is there something wrong? @fzaninotto

fzaninotto:
Thanks!

PR #1615, opened 2018-11-09
Update doc block

ankurk91:
Make `e164PhoneNumber` and `e164PhoneNumber` available to IDEs


fzaninotto:
tollFreePhoneNumber isn't present in the root Phone provider - only in selected locales. So please don't include it in the phpDoc.

fzaninotto:
Thanks!

Issue #1597, opened 2018-10-11
Regexify with special symbols returns wrong output


AxelGreen:
For input like this `[\!\"\#\$\&\(\)\*\+\,\-\.\/\<\=\>\@\{\|\}\~]{40,100}`, `regexify()` sometimes return not only this special characters (notice letters in output). And sometimes string is shorter then 40 symbols. 
Example:
```
$regex = '[\!\"\#\$\&\(\)\*\+\,\-\.\/\<\=\>\@\{\|\}\~]{40,100}';
// 'p' symbol at the end, and not long enough
var_dump($faker->regexify($regex)); // string(39) "|$={!},$}@~-$=*#.i<@,+/"/*-#@*=&>/@~p~/"
// 'Z' and 'D' symbols at the end, and not long enough
var_dump($faker->regexify($regex)); // string(37) "",~->{+..|+$`!<@*<@|"~&&...=Z~|{=.-D="
// this string is ok, but too short
var_dump($faker->regexify($regex)); // string(26) "*@&/!=-|~=><$/!>*&|{&$,~~>"
```

fzaninotto:
Faker's `regexify` formatter is very limited by design, because it's already super slow, and adding support for more RegEx grammar would make it even slower. So we won't address cases like this one. The community can choose to address them in another package.

freearhey:
@fzaninotto This is literally a bug and you do not want to do anything about it?

fzaninotto:
It is not a bug if it is documented as not supported.

https://github.com/fzaninotto/Faker/blob/1d996514052fa7dc75077ce5edba85672eb071be/src/Faker/Provider/Base.php#L439-L465

At least that's my opinion.

freearhey:
@fzaninotto Well, why don't I see the mention of this "feature" in the documentation?

<img width="627" alt="Screenshot 2019-11-06 at 13 19 49" src="https://user-images.githubusercontent.com/7253922/68290139-7a978200-0098-11ea-804d-5cecde5911a8.png">

PR #1624, opened 2018-11-27
Harden regexify() against escaped special characters in input

tlezard:
Addresses issue #1597


fzaninotto:
Thanks for your patch. 

PR #1627, opened 2018-11-30
Add imageLocal method to Image provider

jakebathman:
## Overview

This PR adds a new method to the Image provider called `imageLocal()`.

The idea behind this method is to avoid an HTTP request for an image if any image file will do. It generates an image of a solid, random color of the given `$width` and `$height`. 

Please let me know if there are any questions or issues. Thanks!


pimjansen:
The idea is to completely remove the image provider from the Faker core. Therefor we are not going to accept this.

PR #1630, opened 2018-12-05
Fixed BIC (Swift code) numbers generation

### LINKED ISSUE #1577: Generates invalid IBAN
Hi,

Faker just generated the IBAN "FR4350142857692ZLW2D67JCG24" which is invalid according to https://www.ibancalculator.com/iban_validieren.html and my SEPA provider.

Is it normal? If yes, maybe the documentation should be updated to reflect the fact that the IBAN is only partially valid (only the first checksum).

Thanks,

unrealevil:
Fixes #1577 
Improved BIC-Swift numbers generation with country code


pimjansen:
Maybe we should add a full test to check if the generated ibans are valid? Since there a clear validation rules to verify if an iban is actually valid:

https://en.wikipedia.org/wiki/International_Bank_Account_Number

bjoernffm (approved):

PR #1631, opened 2018-12-06
Minor typo fix

nhedger:
The word **Licence** is written in french. It should be written in english to keep consistency with the other occurrences of the word in this document.


pimjansen:
Nice find, thanks!

localheinz:
Thank you, @nhedger!

Issue #1640, opened 2019-01-04
Feature request, datetime start, end, interval - dateperiod as input


slifin:
I have a use case where I'd like a random day between two points in time, I'd like to express this is a ```DatePeriod``` object and pass it to faker for it to randomly pick a date in the range

perhaps ```dateTimePeriod(\DatePeriod $dateperiod);```

Example call might look like:
```
$faker->dateTimePeriod(
  new \DatePeriod(
    new \DateTime('2018-01-01'),
    \DateInterval::createFromDateString('1 day'),
    new \DateTime('2019-01-01'),
  )
);
```
with a \DateTime object as a result, seems like this form can express every other form in the DateTime API

PR #1642, opened 2019-01-04
Allows datetime selection between start, end across interval (with \DatePeriod)

slifin:
Adds the ability to pick a random datetime object between two points in time across a given interval

In response to this issue (self reported): https://github.com/fzaninotto/Faker/issues/1640

### DIFF
--- a/src/Faker/Provider/DateTime.php
+++ b/src/Faker/Provider/DateTime.php
@@ -141,7 +141,22 @@ class DateTime extends Base
             $timezone
         );
     }
-
+    
+    /**
+     * Get a DateTime object based on a given DatePeriod.
+     *
+     * @param \DatePeriod $period A PHP DatePeriod object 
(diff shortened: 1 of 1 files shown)


slifin:
What are the places where DateTime methods are enumerated? 

I can see:
```
/readme.md
/src/Faker/Generator.php
```

Are there any others?

slifin:
Looks like I'm missing tests too

slifin:
I have a local commit that will handle the docs, but for the tests I'd like to use generative testing, but I can't find any examples of anyone doing that  in the surrounding file:
```/test/Faker/Provider/DateTimeTest.php```

I'd like to randomly generate the DatePeriod object and then compare its outputs against some universal truths for the method that should line up, I'd like to do it in a reproducable way with seeds, like you would in faker but don't know how to do in this testing context

How do I do a generative/property based test in DateTimeTest?

PR #1652, opened 2019-01-22
Fixed `->optional(1.0)` not returning the provider

Seb-C:
When the weight is `1.0`, which should always return the generator (according to the documentation), we never actually get the generator. The second test does not pass either because it is a float.


fzaninotto:
Could you please add a unit test showing your fix?

Seb-C:
@fzaninotto Done.

src/Faker/Provider/Base.php
```diff
@@ -546,7 +546,7 @@ public function optional($weight = 0.5, $default = null)
     {
         // old system based on 0.1 <= $weight <= 0.9
         // TODO: remove in v2
-        if ($weight > 0 && $weight < 1 && mt_rand() / mt_getrandmax() <= $weight) {
+        if ($weight > 0 && $weight <= 1 && mt_rand() / mt_getrandmax() <= $weight) {
```
pimjansen: Why do we want an equals here? If the weight is 1 it will always be the same value so not optional?

pimjansen:
@Seb-C bump

PR #1658, opened 2019-02-08
Bug Fix en_SG mobile number generation

ziming:
Singapore mobile number only starts with 8 or 9. Not 7. The 2nd digit is never a 9 as well

See: http://en.wikipedia.org/wiki/Telephone_numbers_in_Singapore#Numbering_plan

Hope it helps!

```
8zxx xxxx - Mobile and Data Services

9yxx xxxx - Mobile, Data Services and Pager (until May 2012)

Which also means:

8[1-8]xx xxxx - Mobile and Data Services

9[0-8]xx xxxx - Mobile, Data Services and Pager (until May 2012)

x denotes 0 to 9
y denotes 0 to 8 only
z denotes 1 to 8 only
```

### DIFF
--- a/src/Faker/Provider/en_SG/PhoneNumber.php
+++ b/src/Faker/Provider/en_SG/PhoneNumber.php
@@ -100,8 +100,8 @@ class PhoneNumber extends \Faker\Provider\PhoneNumber
         return static::randomElement(static::$zeroToEight);
     }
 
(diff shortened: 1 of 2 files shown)


ziming:
Hmm. How do I add the bug label?

pimjansen:
Thanks for your contribution!

PR #1663, opened 2019-02-15
fr_FR Address : Adding new Administrative Region

jbaptisteq:
**fix :** protected static $streetPrefix, remove double 'rue' prefix

**feature :** adding new administrative region, private static $newRegions, public static function newRegion()

**docs :** adding newRegion function in readme.md


jbaptisteq:
Array change but something goes wrong out of the update done.
I will check that.

PR #1564, opened 2018-08-22
fixed zh_CN realtext

ppiian:
Fixed zh_CN realtext.

PR #1666, opened 2019-02-24
fixed zh_CN realtext

ycgambo:
see: [**fixed zh_CN realtext** #1564](https://github.com/fzaninotto/Faker/pull/1564)

I don't write a test because this is a copy of [zh_TW/Text](https://github.com/fzaninotto/Faker/blob/master/src/Faker/Provider/zh_TW/Text.php) and we have the same [syntax rules](https://github.com/fzaninotto/Faker/blob/master/test/Faker/Provider/zh_TW/TextTest.php).


fzaninotto:
If it's the same content as zh_CN, then just extend that class instead of copying the body.

ycgambo:
I'm not sure if it's appropriate to extend another inherited Text

fzaninotto:
As it's the same text, please don't repeat the `$baseText ` property.

ycgambo:
No. They looks alike but they are actually not the same text if you look again.

PR #1668, opened 2019-03-01
Add 'Faker\Provider\en_US\Company::catchPhrase' to ru_RU locale

axklim:
Added test, set of words took with en_US locale

all tests passed

image


fzaninotto:
Thanks!

PR #1669, opened 2019-03-02
Add zh_TW city prefix and city suffix

Mombuyish:
In `city()` method having `$county.$city` but sometimes we need prefix `台北市` and suffix `中山區`.

### DIFF
--- a/src/Faker/Provider/zh_TW/Address.php
+++ b/src/Faker/Provider/zh_TW/Address.php
@@ -404,12 +404,12 @@ class Address extends \Faker\Provider\Address
 
     public static function cityPrefix()
     {
-        return '';
+        return static::randomElement(array_keys(static::$city));
     }
 
     public static function citySuffix()
     {
-        return '';
+        return static::randomElement(call_user_func_array('array_merge', array_values(static::$city)));
     }
 
     public static function secondaryAddress()
--- a/test/Faker/Provider/zh_TW/AddressTest.php
+++ b/test/Faker/Provider/zh_TW/AddressTest.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Faker\Test\Provider\zh_TW;
+
+use PHPUnit\Framework\TestCase;
+
+class AddressTest extends TestCase
+{
+
+    private $addressClass;
+
+    public function setUp()
+    {
+        $this->addressClass = new \ReflectionClass('Faker\Provider\zh_TW\Address');
+    }
+
+    protected function getMethod($name)
+    {
+        $method = $this->addressClass->getMethod($name);
+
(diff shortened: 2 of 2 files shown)


fzaninotto:
I can't judge if this is correct. @puckwang @tzhuan can you please give your insight?

tzhuan:
I understand @Mombuyish 's need but IMHO `台北市` is not a `cityPrefix` and `中山區` is not a `citySuffix`. Actually `中山區` is a district so you may create a new method `district()` and separate the city and district parts.

Mombuyish:
@tzhuan Yes, Should I change the PR? Maybe I need to create `county()` and `district()` instead of before `cityPrefix()` and `citySuffix()` ?

Any suggestions welcome.

puckwang:
I agree @tzhuan 's comment. 
Maybe you only need to create `district()`, And pass city name to function to get random district.

Also, I think cityPrefix() and citySuffix() two function can remove.

Mombuyish:
@puckwang Thanks for the suggestion, I will hand on it for later.
I agree the district method passing the city name to generate random district, but sometimes I need randomly county and district, maybe do I need to create `county` and `district` methods?

* Remove `cityPrefix` and `citySuffix` methods.
* Add county method random name.
* Add district method if passing county name throw belongs the array value, otherwise randomly.

puckwang:
Maybe you can change `city()`, Only return randomly county(直轄市 & 縣市). What do you think? @Mombuyish

Mombuyish:
@puckwang that's great so do I think.

* Remove `cityPrefix ` and `citySuffix ` methods.
* Change `city` implementation, generating the city name without `district` randomly.
* Add district method if passing county name throw belongs the array value, otherwise randomly.

I think it will keep understanding clearly.

Mombuyish:
@tzhuan @puckwang The works are done.
Please review the codes, if you have any suggestions let me know, thanks!

tzhuan:
`call_user_func_array('array_merge', array_values(static::$city))` might be very expensive if the number of cities/districts is very large. If I were you I would prefer to use two `randomElement()` calls to select a distinct:
```php
public function district($city = null)
{
  if (!isset(static::$city[$city])) {
    $city = static::randomElement(array_keys(static::$city));
  }
  return static::randomElement(static::$city[$city]);
}
```
While performance is not the key for this kind of library so your implementation is also okay, by the way.

Mombuyish:
@tzhuan @puckwang Thanks for suggestions, it's done for works.

tzhuan:
Now it looks good to me. :)

PS. you may consider improving the title of this PR to reflect your work more correctly and precisely.

PR #1671, opened 2019-03-05
update polish titles method that breaks the code

zkisly-toucan:
There was a call in static function to non-static methods that caused crashes. It is now fixed along with minor PHPdoc update


fzaninotto:
Can you paste a way to reproduce the crash this PR is supposed to fix? Even better: write a unit test about it.

pimjansen:
@zkisly-toucan bump

pimjansen:
I am going to close this PR since it is stale. Feel free to reopen it if needed later on.
Thanks for contributing!

PR #1673, opened 2019-03-15
Updated readme.md: added a missing backslash

aubryfr:
Just added a missing backslash in the populator example, to ensure consistency with the previous instruction.


fzaninotto:
Thanks!

PR #1676, opened 2019-03-20
Make en_CA Phone Provider Extend the en_US One

darrylhein:
Everything else applies to Canada.

It may even be possible to remove the Canadian class, or remove everything in it to avoid a BC.

### DIFF
--- a/src/Faker/Provider/en_CA/PhoneNumber.php
+++ b/src/Faker/Provider/en_CA/PhoneNumber.php
@@ -2,7 +2,7 @@
 
 namespace Faker\Provider\en_CA;
 
-class PhoneNumber extends \Faker\Provider\PhoneNumber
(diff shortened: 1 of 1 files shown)


src/Faker/Provider/en_CA/PhoneNumber.php
```diff
@@ -2,7 +2,7 @@
 
 namespace Faker\Provider\en_CA;
 
-class PhoneNumber extends \Faker\Provider\PhoneNumber
+class PhoneNumber extends \Faker\Provider\en_CA\PhoneNumber
```
marchanlon: I think there is a typo in your proposed commit (causing it to fail CI). Did you mean?

```suggestion
class PhoneNumber extends \Faker\Provider\en_US\PhoneNumber
```

darrylhein:
@marchanlon ops! thanks!

fzaninotto:
Thanks!

PR #1677, opened 2019-03-23
name list extended

ikidnapmyself:


fzaninotto:
Sorry, as your PR doesn't link to a source with an MIT-compatible licence, I can't accept that change in Faker, which uses a MIT Licence.

ikidnapmyself:
I understand your concern. I will update my code asap.

PR #1678, opened 2019-03-24
PHPDoc for `passthrough` function on Generator

tabakhase:
added in #1493 - fixup the missing PHPDoc

from currently broken

to inspection working after


fzaninotto:
Thanks!

PR #1682, opened 2019-03-31
Add Undeclared Faker Instance to PhoneNumber Test in en_NG

thearsalan:
Added declaration for the Faker Generator instance in the PhoneNumber test


fzaninotto:
Thanks!

PR #1683, opened 2019-04-01
Add loremflickr.com Provider to the thrid parties

xvladxtremal:
I have written a loremflickr.com provider, since lorempixel works slow or gets down.


fzaninotto:
thanks!

PR #1686, opened 2019-04-14
Added realText capability for fr_CA

bobanum:
Text from Project Gutenberg EBook
La chasse galerie, by Honoré Beaugrand.
Thanks


fzaninotto:
Thanks!

PR #1696, opened 2019-05-21
Calling a non static method as static not allowed.

wilco-1985:
Title method is not static so it raises a php strict warning


pimjansen:
Thanks for contributing!

PR #1698, opened 2019-05-22
added Payment->bank() method for it_IT provider

xfudox:


src/Faker/Provider/it_IT/Payment.php
```diff
@@ -16,4 +16,58 @@ public static function bankAccountNumber($prefix = '', $countryCode = 'IT', $len
     {
         return static::iban($countryCode, $prefix, $length);
     }
+
+    /**
+     * Source:
+     * @see http://www.comuni-italiani.it/banche/principali.html
```
localheinz: Perhaps we should use data from https://en.wikipedia.org/wiki/List_of_banks_in_Italy instead?

fzaninotto: indeed, the page linked by @xfudox has a copyright notice, we can't reuse the content in Faker, which uses an MIT license.

fzaninotto:
@localheinz @pimjansen no need to tag each other as reviewers on every PR, this brings too much notification noise

src/Faker/Provider/it_IT/Payment.php
xfudox: ok, i'll correct this asap

src/Faker/Provider/it_IT/Payment.php
```diff
@@ -16,4 +16,45 @@ public static function bankAccountNumber($prefix = '', $countryCode = 'IT', $len
     {
         return static::iban($countryCode, $prefix, $length);
     }
+
+    /**
+     * Source:
+     * @see https://en.wikipedia.org/wiki/List_of_banks_in_Italy
+     */
+    protected static $banks = array(
+        'UniCredit',
```
fzaninotto: Could you sort this list alphabetically and group bank names by capital letter? This will help tracking future additions / deletions.

xfudox:
bumping this up

PR #1699, opened 2019-05-22
added Payment->bank() method for en_US provider

xfudox:


xfudox:
Hello? Anyone gave a look to this?

src/Faker/Provider/en_US/Payment.php
```diff
@@ -29,10 +29,126 @@ public function bankRoutingNumber()
 
     public static function calculateRoutingNumberChecksum($routing)
     {
-        return (
```
localheinz: Can you revert this unrelated change here, please?

xfudox: no problem, i'll do asap

pimjansen: @xfudox bump

pimjansen:
Going to remove this from 1.9.0 since it is still open

xfudox:
bumping this up

pimjansen:
Since it was removed from 1.9 i dont see this happening anytime soon.

It can be part of 2.0 but this will be a new major which will remove all the locales from core. So please keep this in min

PR #1700, opened 2019-05-22
Added missing "red" color to $safeColorNames

xfudox:


xfudox:
noticed i made the branch not from master but from another branch, and it included a non-pertinent commit

PR #1701, opened 2019-05-22
Added 'red' to $safeColorNames

xfudox:


fzaninotto:
@localheinz @pimjansen This is the kind of change that I usually refuse. The reason is that many people use Faker for unit tests (with a fixed seed). Whenever we change a dataset, these unit tests break. We should only do it for valid reasons. Adding one color (or one name, or one city) isn't a good enough reason. 

If this dataset is incomplete, then we need a PR with a complete dataset, that doesn't risk being updated in the near future. I summarize it by saying "a change in a dataset should bring a large enough added value".

@xfudox I'm reverting your merge for the reasons details above. Thanks for your contribution anyway.

pimjansen:
Ok fair enough 👌🏻

PR #1702, opened 2019-05-22
Added two Color methods to be more web-related explicitly

xfudox:
Added two new methods to Color provider:
 - Color::htmlColorName()
 - Color::x11ColorName()
Those are just colorName() and safeColorName() renamed but with those new names it is more clear the relation with web colors.

### DIFF
--- a/src/Faker/Provider/Color.php
+++ b/src/Faker/Provider/Color.php
@@ -113,4 +113,12 @@ class Color extends Base
     {
         return static::randomElement(static::$allColorNames);
     }
+
+    /**
+     * @example 'olive'
+     */
+    public static function htmlColorName()
+    {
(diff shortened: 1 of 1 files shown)


src/Faker/Provider/Color.php
```diff
@@ -113,4 +113,20 @@ public static function colorName()
     {
         return static::randomElement(static::$allColorNames);
     }
+
+    /**
+     * @example 'olive'
+     */
+    public static function htmlColorName()
```
localheinz: Not sure about this method, it effectively introduce an alias vor `safeColorName()` - is that intended?

xfudox: yeah, the idea was to have two methods that clearly express relation with web colors

xfudox:
bumping this up

PR #1709, opened 2019-06-06
Added top 500 en_GB Surnames

pet1330:
Currently, there are only 99 surnames for en_GB which can be quite limiting when generating unique surnames, this PR just adds another 401.

Source: https://surnames.behindthename.com/top/lists/england-wales/1991

Although the source is from 1991 it seems okay to me. The old link also appears to be broken.

All of the 99 names originally in the list are also in the top 500, so no names have been lost.

### DIFF
--- a/src/Faker/Provider/en_GB/Person.php
+++ b/src/Faker/Provider/en_GB/Person.php
@@ -67,27 +67,30 @@ class Person extends \Faker\Provider\Person
     );
 
     /**
-     * @link http://surname.sofeminine.co.uk/w/surnames/most-common-surnames-in-great-britain.html
+     * @link https://surnames.behindthename.com/top/lists/england-wales/1991
      */
     protected static $lastName = array(
-        'Adams', 'Allen', 'Anderson',
-        'Bailey', 'Baker', 'Bell', 'Bennett', 'Brown', 'Butler',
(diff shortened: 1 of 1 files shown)


localheinz:
@pet1330 

Not sure if it's possible for us to use these.

I have checked https://surnames.behindthename.com/info/copyright and reached out to the maintainer of the website by mail, but unless he confirms we can use this under the MIT license, we probably will not be able to merge this in.

pimjansen:
What i could see is that a lot of references from commercial books and wikis are used which would mean it will not work i assume.

src/Faker/Provider/en_GB/Person.php
```diff
@@ -67,27 +67,30 @@ class Person extends \Faker\Provider\Person
     );
 
     /**
-     * @link http://surname.sofeminine.co.uk/w/surnames/most-common-surnames-in-great-britain.html
+     * @link https://surnames.behindthename.com/top/lists/england-wales/1991
```
fzaninotto: Unfortunately that source is copyrighted. We can't use it in an MIT project like Faker. Please find an open-source list, or close the PR.

pet1330:
Sorry, I did not think to check their licence. I will have a look for other similar sources and update this PR next week.

@localheinz can you let me know if you do hear back from the maintainers of the website?

Thanks,
Pete

PR #928, opened 2016-06-03
Added missing grayscale option/flag to Image provider

sebastianvilla:
LoremPixel offers a grayscale option which was already available in the `imageUrl` method in the Image provider; however, this option was missing in the`image` method.

PR #1714, opened 2019-06-13
Added missing grayscale flag in image provider

tswestendorp:
Contains the same change as #928, but includes README change as well

Closes #928


localheinz:
Duplicate of #928.

PR #1719, opened 2019-06-24
Switch image provider to Unsplash

### LINKED ISSUE #1712: Lorem Pixel SSL error
The `Image` provider uses Lorem Pixel with the domain `https://lorempixel.com/`, it would seem that the SSL for that site is invalid.

So, when you call the method it just returns `false`

Would it be better to update the URL or use a different provider?

### LINKED ISSUE #1715: Faker Image not working
http://lorempixel.com/ website is down. so no images is being downloaded or grab.

tswestendorp:
Lorempixel's service has been having issues lately, SSL expired some days ago (see #1712) and at time or writing it's completely down (again). 

Fixes #1712
Fixes #1715

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -261,11 +261,11 @@ Methods accepting a `$timezone` argument default to `date_default_timezone_get()
 
 ### `Faker\Provider\Image`
 
-    // Image generation provided by LoremPixel (http://lorempixel.com/)
-    imageUrl($width = 640, $height = 480) // 'http://lorempixel.com/640/480/'
-    imageUrl($width, $height, 'cats')     // 'http://lorempixel.com/800/600/cats/'
-    imageUrl($width, $height, 'cats', true, 'Faker') // 'http://lorempixel.com/800/400/cats/Faker'
-    imageUrl($width, $height, 'cats', true, 'Faker', true) // 'http://lorempixel.com/grey/800/400/cats/Faker/' Monochrome image
+    // Image generation provided by LoremFlickr (https://loremflickr.com/)
+    imageUrl($width = 640, $height = 480) // 'https://loremflickr.com/640/480/'
+    imageUrl($width, $height, 'cats')     // 'https://loremflickr.com/800/600/cats/all'
+    imageUrl($width, $height, 'cats', true, 'Faker') // 'https://loremflickr.com/800/400/cats,Faker/all'
+    imageUrl($width, $height, 'cats', true, 'Faker', true) // 'https://loremflickr.com/grey/800/400/grayscale,cats,Faker/all' Monochrome image
     image($dir = '/tmp', $width = 640, $height = 480) // '/tmp/13b73edae8443990be1aa8f1a483bc27.jpg'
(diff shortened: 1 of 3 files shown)


readme.md
```diff
@@ -261,11 +261,11 @@ Methods accepting a `$timezone` argument default to `date_default_timezone_get()
...
-    // Image generation provided by LoremPixel (http://lorempixel.com/)
-    imageUrl($width = 640, $height = 480) // 'http://lorempixel.com/640/480/'
-    imageUrl($width, $height, 'cats')     // 'http://lorempixel.com/800/600/cats/'
-    imageUrl($width, $height, 'cats', true, 'Faker') // 'http://lorempixel.com/800/400/cats/Faker'
-    imageUrl($width, $height, 'cats', true, 'Faker', true) // 'http://lorempixel.com/grey/800/400/cats/Faker/' Monochrome image
+    // Image generation provided by Unsplash (https://source.unsplash.com/)
+    imageUrl($width = 640, $height = 480) // 'https://source.unsplash.com/640x480/'
+    imageUrl($width, $height, 'cats')     // 'https://source.unsplash.com/800x600/?cats'
+    imageUrl($width, $height, 'cats', true, 'Faker') // 'https://source.unsplash.com/800x400/?cats,Faker'
+    imageUrl($width, $height, 'cats', true, 'Faker', true) // 'https://source.unsplash.com/800x400/?grayscale,cats,Faker' Monochrome image
```
Jamesking56: This doesn't always produce greyscale images in my testing.

test/Faker/Provider/ImageTest.php
```diff
@@ -7,24 +7,40 @@
 
 class ImageTest extends TestCase
 {
+    protected $host = 'source.unsplash.com';
```
Jamesking56: Why not a constant?

readme.md
tswestendorp: I know, Unsplash doesn't support specific grayscale images, this is the best it gets... other option would be to remove the `$gray` flag

test/Faker/Provider/ImageTest.php
tswestendorp: No specific reason actually

src/Faker/Provider/Image.php
```diff
@@ -30,25 +30,24 @@ class Image extends Base
      */
     public static function imageUrl($width = 640, $height = 480, $category = null, $randomize = true, $word = null, $gray = false)
     {
-        $baseUrl = "https://lorempixel.com/";
-        $url = "{$width}/{$height}/";
+        $baseUrl = 'https://source.unsplash.com/';
```
Jamesking56: This could also be a constant

pimjansen:
I wouldnt be a fan of just replacing the service. If unsplash is having problems we are going to switch back again?

fzaninotto:
I've already stated multiple times that I think the Image provider should be moved out of the core, because it depends on a third party service that can fail. Replacing it with another third party provider won't change that fact. 

But removing the Image provider is a breaking change, so we can only do that in a major release.

As for your Unsplash provider, I suggest that you publish it as a standalone package. we'll reference it in the README.

PR #1722, opened 2019-07-01
fix(issue-1721): add phpdoc method and properties annotations for php…

### LINKED ISSUE #1721: Undefined method and properties phpstan
Hello,

For information:

Call to an undefined method Faker\Generator::word().
Access to an undefined property Faker\Generator::$secondaryAddress.
Access to an undefined property Faker\Generator::$state.

lugus:
Fix #1721


pimjansen:
Thanks for the fix!

PR #1723, opened 2019-07-01
This function will generate a random number of cnh

mayckol:
- CNH's valid throughout the national territory


mayckol:
Hey guys, I'm need this function in my project... 
Please review as soon as possible because other companies in my group need the resource
Thank's

PR #1726, opened 2019-07-04
Fixed Greek Vat number function (grafm)

seldimi:
This pull request fixes the issue with the Greek Vat number generator

### DIFF
--- a/src/Faker/Provider/el_GR/Company.php
+++ b/src/Faker/Provider/el_GR/Company.php
@@ -73,6 +73,14 @@ class Company extends \Faker\Provider\Company
     }
 
     /**
+     * @example '111111111'
+     */
+    public static function grafm()
+    {
+        return static::numerify(static::randomElement(static::$grafm));
+    }
+
+    /**
      * Return The profession of a company
      *
      * @example 'Δημοσιογράφος'


src/Faker/Provider/el_GR/Company.php
```diff
@@ -72,6 +72,14 @@ public static function doy()
         return static::randomElement(static::$doy);
     }
 
+    /**
+     * @example '111111111'
+     */
+    public static function grafm()
```
localheinz: Should 

- this method be called `vat()` 
- this method be located in `Faker\Provider\el_GR\Payment`

?

localheinz:
Since a `vat()` formatter has not existed before for the `el_GR` locale, I have labeled this as an `enhancement`/

src/Faker/Provider/el_GR/Company.php
gmponos: Greek guy here 

Note that `vat` means the value added to a product.. Title is misleading as a Greek guy here this happens very often to me as well.

 `afm` though is the Tax Identification number of each Greek company/citizen

Back to our subject, well.. since two lines above there is `doy` which is the greeklish version of `Tax office` you could have `grafm` here as well which is the greeklish version of Tax Identification Number

The more accurate indeed would be to be named TIN and return in it's international form `EL123456789` but if you have a PHP application which does not read it in it's international form and reads it as `123456789` generating fake data for that app with the `tin` function will not help you..

Concluding I suggest to stay as is and maybe have later an enhancement adding tin functions to this package

pimjansen:
@gmponos could you add a test for this?

pimjansen:
@seldimi could you add a test for this as well?

seldimi:
> @seldimi could you add a test for this as well? If you make it on time we can add it in the 1.9.0 release this month

I dont know how to make tests.. any example?

pimjansen:
@seldimi have a look at the phpunit docs and other tests in the repo. Since i believe this will still take a while i will remove this from the 1.9.0 release

seldimi:
I really didn't see any tests in GR, so i couldn't make similar.
But function is really straight forward

PR #1727, opened 2019-07-05
Mussum Ipsum replaces lorem ipsum for brazilian provider

juliofiuza:
This is a known random text generator for brazilian developers based on:
http://mussumipsum.com/


pimjansen:
Since it is about random "text like content" i don't really see a benefit of adding this in the core at this point compared to the Lorum provider there already is.

@fzaninotto what is your view?

fzaninotto:
The Lorem formatter is never translated. Check the `realText` formatter for localized random text.

Issue #1471, opened 2018-04-20
Doctrine entity population for large amount of data hits memory limit


buffcode:
Currently the Doctrine populator (probably also the other populators) holds all generated entities in a variable while executing. When generating real life test data (in terms of object amount) we hit memory limits quite fast.

One possible workaround would be to replace the entity with a more lightweight Reference/proxy object.

Another workaround:
Using DoctrineFixturesBundle one could use the `--append` flag in order to create multiple batches of data. But splitting the generation into multiple fixtures won't help as the associations won't pick up data from previous batches. So I cannot run `--append` 100 times for 1.000 products resulting in 1.000.000 entities and then another run for just orders containing random (existing) products.

Any ideas?

tomzx:
How much RAM are we talking about? Could you provide an example snippet of code of your use case?

pimjansen:
@buffcode the memory limit hit due the generation or due to object size? I expect the last bit which is a bit how `Doctrine` works. We could make a choice to flush on specific batch sizes instead of the total size

Issue #1125, opened 2016-12-21
Allowed memory exhausted.


mikeshow:
Here what I get when I use realText() in a loop:

PHP Fatal error:  Allowed memory size of 268435456 bytes exhausted (tried to allocate 4096 bytes) in C:\...\faker\fzaninotto\faker\src\Faker\Provider\Text.php on line 91

fzaninotto:
What's your locale? How many iterations have you done when the memory is exhausted?

mikeshow:
Local fr_FR and two successful and then error.

mikeshow:
I tried to place the call into a function hope the memory would be freed upon return, but changed nothing.

curry684:
You could try invoking [`gc_collect_cycles`](http://php.net/manual/en/function.gc-collect-cycles.php) at the start of the function to force cleanup of any pending cycles left after the previous invocation.

Not a solution but a workaround of course.

mikeshow:
Thank you very much, apparently that do the trick.

curry684:
@fzaninotto perhaps if the text generator is that memory hungry this should be part of the provider wrapper?

cristiangrama:
Got same issue here in `vendor/fzaninotto/faker/src/Faker/Provider/Base.php`.

mohamednizar:
I fixed this issue temporarily by extending the php memory size.

sw4n:
I'm sorry, but why it's closed? Is it fixed?

mikeshow:
Hi,
I closed it because I got an answer which fixes my problem. Now I don't know if @fzaninotto has or is about to fix it.

sw4n:
Okey, got it. Increasing memory helped me too. But it's interesting, am I wrong with using it or it's a bug connected with memory leaks?

OrkhanAlikhanov:
I think it creates all models, then saves. Because when it fails here is no data added there. Or it uses transaction.
Dividing model creation into parts works:
```php
$count = \App\User::count();
for ($i = 0; $i < 10; $i++) {
    $this->command->info("Adding $count answers");
    factory(\App\UserAnswer::class, $count)->create();
    gc_collect_cycles();
    $this->command->info("Current answers: " . ($count * ($i + 1)));
}
```

prazian:
Increasing memory doesn't help me if I run unit test more than 3-4 times, so I had to manually add gc_collect_cycles() before each call to text generating method.

Is there any solution for this issue? :)

[...]

PR #1005, opened 2016-08-25
Implement singleton for Faker\Factory class

mathielo:
Ran into memory issues when seeding the project database. Several instances of Faker were being created in different loops for different locales, and turns out that `realText()` loads up a huge text chunk to the memory *per instance*. 

Therefore implemented a simple singleton solution to make it possible reuse previously created instances of Faker. A simple test shows the huge difference between reusing instances or not:

```php
// New instances, NO singleton
for ($i = 0; $i < 20; $i++) {
    $faker = Faker\Factory::create();
    $faker->realText(30, 5); // Memory eater
    $mem = memory_get_peak_usage(true) / 1024 / 1024;
    echo "Loop: {$i} - Mem: {$mem}Mb\n";
}
```

> Loop: 0 - Mem: 24Mb
Loop: 1 - Mem: 38Mb
Loop: 2 - Mem: 50Mb
Loop: 3 - Mem: 64Mb
Loop: 4 - Mem: 78Mb
Loop: 5 - Mem: 92Mb
Loop: 6 - Mem: 106Mb
Loop: 7 - Mem: 120Mb
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 4096 bytes) in /Users/mathielo/projects/marketplace-api/vendor/fzaninotto/faker/src/Faker/Provider/Text.php on line 121

Now reusing instances:

```php
for ($i = 0; $i < 20; $i++) {
//    $faker = Faker\Factory::create();
    $faker = Faker\Factory::load();
    $faker->realText(30, 5); // Memory eater
    $mem = memory_get_peak_usage(true) / 1024 / 1024;
    echo "Loop: {$i} - Mem: {$mem}Mb\n";
}
```

> Loop: 0 - Mem: 24Mb
Loop: 1 - Mem: 24Mb
Loop: 2 - Mem: 24Mb
Loop: 3 - Mem: 24Mb
Loop: 4 - Mem: 24Mb
Loop: 5 - Mem: 24Mb
Loop: 6 - Mem: 24Mb
Loop: 7 - Mem: 24Mb
Loop: 8 - Mem: 24Mb
Loop: 9 - Mem: 24Mb
Loop: 10 - Mem: 24Mb
Loop: 11 - Mem: 24Mb
Loop: 12 - Mem: 24Mb
Loop: 13 - Mem: 24Mb
Loop: 14 - Mem: 24Mb
Loop: 15 - Mem: 24Mb
Loop: 16 - Mem: 24Mb
[...]

Issue #921, opened 2016-05-23
Unique slug faker has really high memory usage


shopblocks:
```php
ini_set('memory_limit', -1);

function dd()
{
    array_map(function ($x) {
        dump($x);
    }, func_get_args());

    die( 1 );
}

function human_filesize($bytes, $dec = 2)
{
    $size   = array('B', 'kB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB');
    $factor = floor((strlen($bytes) - 1) / 3);

    return sprintf("%.{$dec}f", $bytes / pow(1024, $factor)) . @$size[$factor];
}

$mem_before = memory_get_peak_usage(true);

$faker = Faker\Factory::create();

$result = collect(range(1,20))->map(function() use ($faker) {
    return [
        'name' => $faker->unique()->sentence(3),
        'slug' => $faker->unique()->optional()->slug,
    ];
});

$mem_after = memory_get_peak_usage(true);

dd($result, human_filesize($mem_after - $mem_before));
```

Going for just the `name` the memory usage comes out as "0.00B"  and takes `0.018624067306519  seconds` which is odd, and i'm not sure how else to measure this. But the `slug` one being added in, memory goes up to `1.01GB` and takes `3.6474552154541 seconds`.

Removing the `unique` on the `slug` takes it down to `0.011013031005859 seconds` and `0.00B`

fzaninotto:
Surprising results. I guess this is because the slugs are long. Maybe we should store cryptographic hashes instead of full content.

fzaninotto:
what's `collect()`?

fzaninotto:
OK, I get the problem. Each of these statements are fats and consume low memory:

```php
$faker->unique()->sentence(3);
$faker->unique()->slug;
$faker->optional()->slug;
```

What takes a large amount of memory is the combination of `unique()` and `optional()` in that order, which somehow makes sense. I don't know what you expect exaclty, but swapping the order of the two modifiers solves the issue:

```php
// slow as hell, doesn't make sense
$faker->optional()->unique()->slug;
// naturally fast
$faker->unique()->optional()->slug;
```

shopblocks:
collect is github.com/tightenco/collect.

I'll give a better sample code to show the example. 

Also, you are saying that `unique()->optional()` is fast, but it is not. Unless you meant `optional()->unique()`

```php
<?php require_once "../vendor/autoload.php";

function human_filesize($bytes, $dec = 2)
{
[...]

PR #1730, opened 2019-07-10
workaround for memory leak

BePsvPT:
Current, Faker has mutual reference between Generator and Providers. These mutual references will result in memory leak[0]. The only way to free those memory is calling gc_collect_cycles[1].

This pull request uses a workaround to prevent memory endlessly growing, however, memory leak is still exist. To solve the problem completely, it will need to refactor the whole project.

[0] https://www.php.net/manual/en/features.gc.refcounting-basics.php#features.gc.cleanup-problems
[1] https://php.net/manual/en/function.gc-collect-cycles.php

related issues: #1471 #1125 #921
related pull requests: #1005
—

Code to reproduce the bug.

```php
<?php

require __DIR__.'/vendor/autoload.php';

echo sprintf('%.2f MB'.PHP_EOL, memory_get_usage() / 1024 / 1024);

for ($i = 0; $i < 10; ++$i) {
    Faker\Factory::create('zh_TW')->realText(150);

    echo sprintf('%.2f MB'.PHP_EOL, memory_get_usage() / 1024 / 1024);
}
```

master branch output

```text
0.44 MB
18.22 MB
34.54 MB
50.86 MB
67.17 MB
83.49 MB
99.81 MB
116.13 MB
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes)
```

this workaround output

```text
0.44 MB
18.22 MB
18.22 MB
18.22 MB
18.22 MB
18.22 MB
18.22 MB
18.22 MB
18.22 MB
18.22 MB
18.22 MB
```

### DIFF
--- a/src/Faker/Generator.php
+++ b/src/Faker/Generator.php
@@ -193,17 +193,26 @@ namespace Faker;
  */
 class Generator
 {
-    protected $providers = array();
-    protected $formatters = array();
+    protected static $providers = array();
+    protected static $formatters = array();
+
+    public function __construct()
+    {
+        foreach (self::$providers as &$provider) {
+            unset($provider);
+        }
+
+        self::$providers = self::$formatters = array();
+    }
 
     public function addProvider($provider)
     {
-        array_unshift($this->providers, $provider);
+        array_unshift(self::$providers, $provider);
     }
 
     public function getProviders()
     {
-        return $this->providers;
+        return self::$providers;
     }
 
     public function seed($seed = null)
@@ -231,16 +240,16 @@ class Generator
      */
     public function getFormatter($formatter)
     {
-        if (isset($this->formatters[$formatter])) {
-            return $this->formatters[$formatter];
+        if (isset(self::$formatters[$formatter])) {
+            return self::$formatters[$formatter];
         }
-        foreach ($this->providers as $provider) {
-            if (method_exists($provider, $formatter)) {
-                $this->formatters[$formatter] = array($provider, $formatter);
 
-                return $this->formatters[$formatter];
+        foreach (self::$providers as $provider) {
+            if (method_exists($provider, $formatter)) {
+                return self::$formatters[$formatter] = array($provider, $formatter);
             }
         }
+
         throw new \InvalidArgumentException(sprintf('Unknown formatter "%s"', $formatter));
     }
 


localheinz:
@BePsvPT 

Thank you for your pull request!

What do you think about adding a failing test first, perhaps using your code example:

```php
<?php

namespace Faker\Test;

use Faker\Factory;
use PHPUnit\Framework\TestCase;

final class FactoryTest extends TestCase
{
    public function testCreatingGeneratorAndUsingFormatterMultipleTimesDoesNotIncreaseMemoryConsumptionSignificantly()
    {
        for ($i = 0; $i < 10; ++$i) {
            Factory::create('zh_TW')->realText(150);

            $memoryUsageInMegaBytes[] = round(memory_get_usage() / 1024 / 1024);
        }

        self::assertEquals(
            array_slice($memoryUsageInMegaBytes, 0, 1),
            array_unique($memoryUsageInMegaBytes),
            'Failed asserting that the memory usage does not increase significantly when instantiating multiple generators und using the same formatter.'
        );
    }
}
```

src/Faker/Generator.php
```diff
@@ -193,17 +193,26 @@
  */
 class Generator
 {
-    protected $providers = array();
-    protected $formatters = array();
+    protected static $providers = array();
+    protected static $formatters = array();
```
localheinz: On the other hand, I'm not sure if we should introduce state here.

Personally, in projects where I use `fzaninotto/faker`, I have introduced a [test helper trait](https://github.com/localheinz/test-util/blob/master/src/Helper.php#L34-L47) which handles this. Maybe it would be indeed something that should be solved in user land code, as pointed out by @fzaninotto in https://github.com/fzaninotto/Faker/pull/1005#issuecomment-242667630?

fzaninotto:
Thanks for your report. Using ` Faker\Factory::create()` in a loop does, indeed, consume a lot of memory, but that's expected. The docs advocate the reuse of the same Faker instance for multiple calls. In my opinion, your example is a corner case, and the resultant memory leak should be handled in userland. So I'm closing thois issue as we won't fix it in the core.

bytestream:
@fzaninotto comparing PHP 7.3 vs 7.4 on a new installation of `laravel/laravel` with only the below changes I'm seeing the same issue:

```
$ php7.3 vendor/bin/phpunit --repeat 100
Time: 3.74 seconds, Memory: 380.00 MB

$ php7.4 vendor/bin/phpunit --repeat 100
Time: 13.53 seconds, Memory: 744.00 MB
```

If I add `gc_collect_cycles()` to `tearDown()` then the PHP 7.4 is about the same (2MB higher).

I can provide a simple repo with the below changes if you want to replicate.

```php
<?php

namespace App;

use Illuminate\Database\Eloquent\Model;

class Message extends Model
{
    protected $fillable = [ 'text' ];
}
```

```php
<?php

use Faker\Generator as Faker;

$factory->define(\App\Message::class, function (Faker $faker) {
    return [
        'text' => $faker->realText(600),
    ];
});
```

```php
    public function testFaker()
    {
        $x = factory(\App\Message::class)->make();

        $this->assertNotNull($x);
    }
```

deleugpn:
@fzaninotto 

>  The docs advocate the reuse of the same Faker instance for multiple calls.

The problem with the provided solution is that it completely breaks `$faker->unique()` use. By caching the instance of `$faker` as it was done by Laravel 6.18.12, if we now run a large enough test suite that relies on `$faker->unique()`, Faker will never be able to generate enough unique data and will always throw OverflowException.

PR #1736, opened 2019-07-23
sk_SK fix surenames

bbaronSVK:
correct male and female surename


pimjansen:
@bbaronSVK what is wrong with the existing lastname? And why did we add a single name? We are mostly no fan of modifying the resultsets without a proper use case

JanGalek:
@bbaronSVK because `Podhradská` is woman lastname ;)

pimjansen:
@JanGalek is it possible to review the pending comment?

JanGalek:
@pimjansen Pending review for @localheinz, not for me, I approved it.
@fzaninotto I'll create new PR for sort names ;)

PR #1738, opened 2019-08-02
Add  new Sport provider

troccoli:
I am in need of generating data related to sports: sport names, teams and venues for example. At the moment I use a lot of `$faker->company` for the names but that doesn't sound right. So I thought I contribute to the project and add a new provider for sport data.

**This is a Work In Progress still**

This is initial work. If there is interest for such provider I would add team and venue and maybe a some more translations.

**Please let me know in the comments if you would like this new provider**

### DIFF
--- a/src/Faker/Provider/en_GB/Sport.php
+++ b/src/Faker/Provider/en_GB/Sport.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Faker\Provider\en_GB;
+
+class Sport extends \Faker\Provider\Sport
+{
+    protected static $sports = array(
+        'American Football',
+        'Baseball',
+        'Basketball',
+        'Beach Volleyball',
+        'Bowling',
+        'Boxing',
+        'Canoeing',
+        'Cricket',
+        'Cricket',
+        'Curling',
+        'Cycling',
+        'Diving',
+        'Fishing',
+        'Football',
+        'Golf',
+        'Gymnastic',
+        'Handball',
+        'Hockey',
+        'Ice Hockey',
+        'Ice Skating',
+        'Judo',
+        'Karate',
+        'Kayaking',
+        'Lacrosse',
+        'Rollerblading',
+        'Rowing',
+        'Rugby League',
+        'Rugby Union',
+        'Skateboarding',
+        'Skating',
+        'Skiing',
+        'Snowboarding',
+        'Softball',
+        'Swimming',
+        'Taekwondo',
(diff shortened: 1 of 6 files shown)


src/Faker/Provider/en_GB/Sport.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace Faker\Provider\en_GB;
+
+class Sport extends \Faker\Provider\Sport
+{
+    protected static $sports = array(
```
OskarStark: Could you sort them alphabetically?

src/Faker/Provider/en_US/Sport.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace Faker\Provider\en_US;
+
+class Sport extends \Faker\Provider\Sport
+{
+    protected static $sports = array(
```
OskarStark: Same here

src/Faker/Provider/it_IT/Sport.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Faker\Provider\it_IT;
+
+class Sport extends \Faker\Provider\Sport
+{
+    protected static $sports = array(
+        'Golf',
```
localheinz: As @OskarStark suggested, can you sort alphabetically by value here?

OskarStark: Couldn't this be fixed through the whole codebase by a CS fixer rule?

localheinz: I think http://plugins.jetbrains.com/plugin/5919-lines-sorter will work best here.

localheinz (approved):
👍

localheinz (commented):
Perhaps add a section to `readme.md`?

troccoli:
@localheinz @OskarStark 

I have now ordered the list. But this is just the initial work, I spent like an hour on this, so it's not refined at all.

I just want to gauge the appetite for such faker. I now I need it, but I don't want to write my own package if instead it can be incorporated in the main package. I'd rather contribute here than write my own version.

troccoli:
> Perhaps add a section to `readme.md`?

@localheinz where do you suggest I create the readme? Or did you mean update the existing on, aka "the docs"?

troccoli:
> Perhaps add a section to `readme.md`?

I will

PR #1746, opened 2019-08-12
Updated implode and join statements to follow the argument order from the documentation.

japalekhin:
The documented argument order for the `implode` function is `implode($glue, $pieces);`. Although the function accepts any order for "historical reasons".
https://www.php.net/manual/en/function.implode.php

But in PHP 7.4.0beta2 (which I am using to build a Laravel application with) seems to have deprecated this support and the arguments must now be entered in the documented order.

Running `php artisan migrate --seed` encounters this deprecation warning and fails altogether. So I decided to contribute in a change to all `implode` calls.

**Note:** `join` is just an alias of `implode`.


localheinz:
@japalekhin 

Since the build is failing here and we have #1748 - would you mind if we closed this as duplicate?

japalekhin:
Not at all. I guess I forked at the time the build was failing. Glad this has been resolved.

localheinz:
Thank you, @japalekhin!

Issue #1745, opened 2019-08-10
Female last name differ from male last name in some countries. Bug.


tonykor:
Hello,
the bug is that if you want to generate 'First Last' name only for female, you will get a strange combination - a female first name with male last name, this happens with most slavic language groups 'ru, cs, lv, etc.', as male and female forms of the surname is differs a bit in this languages.

Example:

`$faker->firstName('female')." ".$faker->lastName`

can return 
"Филипов Ольга" (ru) must be "Филипов**а** Ольга"
"Novak Jana" (cs) must be "Novak**ová** Jana"

You can said to use a `$faker->name('female')` method, but for example in Russian it generates a full name which is "First Middle Last" which is more official form, and not used widely especially among peoples under 50 y.o. and in the internet.

Regards,
Tony

aanfarhan:
This seems to be not a bug, but not yet covered on Russian Locale.
```php
/**
     * {@link} http://ru.wikipedia.org/wiki/%D0%A1%D0%BF%D0%B8%D1%81%D0%BE%D0%BA_%D0%BE%D0%B1%D1%89%D0%B5%D1%80%D1%83%D1%81%D1%81%D0%BA%D0%B8%D1%85_%D1%84%D0%B0%D0%BC%D0%B8%D0%BB%D0%B8%D0%B9
     **/
    protected static $lastName = array(
        'Смирнов', 'Иванов', 'Кузнецов', 'Соколов', 'Попов', 'Лебедев', 'Козлов',
        'Новиков', 'Морозов', 'Петров', 'Волков', 'Соловьёв', 'Васильев', 'Зайцев',
        'Павлов', 'Семёнов', 'Голубев', 'Виноградов', 'Богданов', 'Воробьёв',
        'Фёдоров', 'Михайлов', 'Беляев', 'Тарасов', 'Белов', 'Комаров', 'Орлов',
        'Киселёв', 'Макаров', 'Андреев', 'Ковалёв', 'Ильин', 'Гусев', 'Титов',
        'Кузьмин', 'Кудрявцев', 'Баранов', 'Куликов', 'Алексеев', 'Степанов',
        'Яковлев', 'Сорокин', 'Сергеев', 'Романов', 'Захаров', 'Борисов', 'Королёв',
        'Герасимов', 'Пономарёв', 'Григорьев', 'Лазарев', 'Медведев', 'Ершов',
[...]

PR #1747, opened 2019-08-15
Add lastName gender specific on ru_RU locale

aanfarhan:
Add` lastName` method on **Person Provider** ru_RU locale with gender params.

Related to #1745.

### DIFF
--- a/src/Faker/Provider/ru_RU/Person.php
+++ b/src/Faker/Provider/ru_RU/Person.php
@@ -105,6 +105,8 @@ class Person extends \Faker\Provider\Person
         'Меркушев', 'Лыткин', 'Туров',
     );
 
+    protected static $lastNameSuffix = array('a', '');
+
     /**
      * Return male middle name
      *
@@ -166,11 +168,13 @@ class Person extends \Faker\Provider\Person
     public function lastName($gender = null)
     {
         $lastName = static::randomElement(static::$lastName);
-        
+
         if (static::GENDER_FEMALE === $gender) {
             return $lastName . 'a';
+        } elseif (static::GENDER_MALE === $gender) {
+            return $lastName;
         }
-        
-        return $lastName;
+
+        return $lastName . static::randomElement(static::$lastNameSuffix);
     }
 }
--- a/src/Faker/Provider/ru_RU/Person.php
+++ b/src/Faker/Provider/ru_RU/Person.php
@@ -161,7 +161,7 @@ class Person extends \Faker\Provider\Person
      * @access public
      * @param string|null $gender A gender of the last name should be generated
      *     for. If the argument is skipped a random gender will be used.
-     * @return string Middle name
+     * @return string Last name
      */
     public function lastName($gender = null)
     {
--- a/src/Faker/Provider/ru_RU/Person.php
+++ b/src/Faker/Provider/ru_RU/Person.php
@@ -172,6 +172,5 @@ class Person extends \Faker\Provider\Person
         }
         
         return $lastName;
-            static::randomElement(static::$lastName) . 'a' : static::randomElement(static::$lastName);
     }
 }
--- a/src/Faker/Provider/ru_RU/Person.php
+++ b/src/Faker/Provider/ru_RU/Person.php
@@ -18,6 +18,14 @@ class Person extends \Faker\Provider\Person
         '{{lastName}}а {{firstNameFemale}} {{middleNameFemale}}',
     );
 
+    protected static $lastNameFormat = array(
+        '{{lastName}}',
+    );
+
+    protected static $lastNameFemaleFormat = array(
+        '{{lastName}}a',
+    );
+
     /**
      * {@link} http://ru.wikipedia.org/wiki/%D0%A0%D1%83%D1%81%D1%81%D0%BA%D0%BE%D0%B5_%D0%BB%D0%B8%D1%87%D0%BD%D0%BE%D0%B5_%D0%B8%D0%BC%D1%8F
      * {@link} http://masterrussian.com/aa031701a.shtml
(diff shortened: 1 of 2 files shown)


antonkomarev:
Thank you! Just faced same issue and wanted to add a pull request.

src/Faker/Provider/ru_RU/Person.php
```diff
@@ -154,4 +154,18 @@ public function middleName($gender = null)
...
+     * Return last name for the specified gender.
+     *
+     * @access public
+     * @param string|null $gender A gender of the last name should be generated
+     *     for. If the argument is skipped a random gender will be used.
+     * @return string Last name
+     */
+    public function lastName($gender = null)
+    {
+        return ($gender === static::GENDER_FEMALE) ?
```
localheinz: Instead of using a ternary statement here and repeating the method invocation to randomly select a last name, what do you think about introducing a temporary variable and then using a condition and early returns?

```suggestion
        $lastName = static::randomElement(static::$lastName);
        
        if (static::GENDER_FEMALE === $gender) {
            return $lastName . 'a';
        }
        
        return $lastName;
```

Also, would you mind adding two tests? Otherwise we’ll follow up!

pimjansen:
@aanfarhan @antonkomarev my russian is not so good but is this valid for all male names that they have a female equivalent with an appended "a"? 
Code overall looks good so 👌🏻

src/Faker/Provider/ru_RU/Person.php
```diff
@@ -154,4 +154,23 @@ public function middleName($gender = null)
...
+     * @access public
+     * @param string|null $gender A gender of the last name should be generated
+     *     for. If the argument is skipped a random gender will be used.
+     * @return string Last name
+     */
+    public function lastName($gender = null)
+    {
+        $lastName = static::randomElement(static::$lastName);
+        
+        if (static::GENDER_FEMALE === $gender) {
```
localheinz: Maybe we should add a link here for explanation?

For example, https://en.m.wikipedia.org/wiki/Eastern_Slavic_naming_customs#Grammar. 

/cc @pimjansen

src/Faker/Provider/ru_RU/Person.php
```diff
@@ -154,4 +154,23 @@ public function middleName($gender = null)
             static::GENDER_FEMALE,
         )));
     }
+
+    /**
+     * Return last name for the specified gender.
+     *
+     * @access public
```
localheinz: Not sure if using the `@access` annotation here is something we really need - it duplicates the `public` visibility keyword that is already present. Can we remove it?

antonkomarev:
@pimjansen not all the female last names ends with `-a`, but most part or them and all the male last names presented in Faker matches this rule.

src/Faker/Provider/ru_RU/Person.php
pimjansen: Agree, im not a fan of "only in our usecases" but it will do for no i guess

src/Faker/Provider/ru_RU/Person.php
```diff
@@ -154,4 +154,23 @@ public function middleName($gender = null)
...
+     * @param string|null $gender A gender of the last name should be generated
+     *     for. If the argument is skipped a random gender will be used.
+     * @return string Last name
+     */
+    public function lastName($gender = null)
```
fzaninotto: Do I understand correctly or is the generated `lastName` always Male by default? Shouldn't it return a random gender?

localheinz: Good point, missed that one.

@aanfarhan 

Can you adjust, please?

aanfarhan: Alright, I'll add some adjustment to the code.

test/Faker/Provider/ru_RU/PersonTest.php
```diff
@@ -0,0 +1,32 @@
...
+use PHPUnit\Framework\TestCase;
+
+class PersonTest extends TestCase
+{
+    /**
+     * @var Generator
+     */
+    private $faker;
+
+    public function setUp()
```
localheinz: Visibility should be `protected` here, but we can - with the right rules - automatically fix.

src/Faker/Provider/ru_RU/Person.php
```diff
@@ -154,4 +156,25 @@ public function middleName($gender = null)
...
+    {
+        $lastName = static::randomElement(static::$lastName);
+
+        if (static::GENDER_FEMALE === $gender) {
+            return $lastName . 'a';
+        } elseif (static::GENDER_MALE === $gender) {
+            return $lastName;
+        }
+
+        return $lastName . static::randomElement(static::$lastNameSuffix);
```
localheinz: What do you think about

```suggestion
        return static::lastName(static::randomElement(array(
            static::GENDER_FEMALE,
            static::GENDER_MALE,
        ));
```

?

Then we can also do away with the newly introduced `protected` property.

src/Faker/Provider/ru_RU/Person.php
```diff
@@ -154,4 +154,28 @@ public function middleName($gender = null)
...
+    {
+        $lastName = static::randomElement(static::$lastName);
+
+        if (static::GENDER_FEMALE === $gender) {
+            return $lastName . 'a';
+        } elseif (static::GENDER_MALE === $gender) {
+            return $lastName;
+        }
+
+        return static::lastName(static::randomElement(array(
```
fzaninotto: this won't work: `Person::lastName()` isn't a static function. Also, I'm bot a big fan of using recursion here - I think it's overkill. Can you revert you last commit? Or perhaps you meant `Person::$lastName`?

fzaninotto:
Thanks!

PR #1748, opened 2019-08-15
PHP 7.4 compatibility

pimjansen:
For PHP 7.4 the implode methods will break:

`implode(): Passing glue string after array is deprecated. Swap the parameters`

I also updated travis to match 7.3. However there is no 7.4 env available in Travis yet as far as i know

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ php:
     - 7.0
     - 7.1
     - 7.2
+    - 7.3
     - nightly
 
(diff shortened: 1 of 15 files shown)


NFarrington:
PHP 7.4 is available in Travis as `7.4snapshot` at the moment.

pimjansen:
@NFarrington then the nightly would still work fine as well untill 7.4 is there as stable

PR #1750, opened 2019-08-20
Update to documentation relative to optional()

### LINKED ISSUE #1749: optional() documentation states you can pass 0 to 1 but it's wrong
I hit something today that the documentation did not state properly. If you are using the popular Faker library, which is a great library, then is a good chance you tried to use the optional() function.

Basically, optional takes a decimal value to generate a NULL or valid value for the provider that you use. The documentation states you can pass 0 to 1 but this isn’t true:

// optional() accepts a weight argument to specify the probability of receiving the default value.
// 0 will always return the default value; 1 will always return the provider. Default weight is 0.5 (50% chance).
$faker->optional($weight = 0.1)->randomDigit; // 90% chance of NULL
$faker->optional($weight = 0.9)->randomDigit; // 10% chance of NULL
In fact, if you load up to code for vendor/fzaninotto/faker/src/Faker/Provider/Base.php, you can see that there is a TODO that states this part must be removed in v2 and the code is:

if ($weight > 0 && $weight < 1 && mt_rand() / mt_getrandmax() <= $weight) {
     return $this->generator;
}
Problem is, if you pass 1 as the weight, it skips that part and takes the 1 to be 1% because the next case is 0 to 100…

I’ll open a bug report to fix the documentation but until then, take it for granted that you should multiply 1 by 100 if you wanted to pass a weight of 1.

That said, i'll post a documentation PR to update this!

No need to do anything.

crazycodr:
Fixes #1749.

The documentation in the readme.md doesn't state the complex nature of `optional()` with the two possible value sets. I've added documentation regarding this!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -342,58 +342,16 @@ for ($i = 0; $i < 10; $i++) {
 }
 print_r($values); // [1, 4, null, 9, 5, null, null, 4, 6, null]
 
-// optional() accepts a weight argument to specify the probability of receiving the default value.
-// --------------------------------------------------------
-// 2 value sets are possible
-// --------------------------------------------------------
-// First set is between 0 and 1 (Fractional representation)
-// --------------------------------------------------------
-// 0 will always return the default value
-// Anything else BELOW 1 will be converted to a percentage
-// 
-// --------------------------------------------------------
-// Second set is between 1 and 100 (Percentage representation)
-// --------------------------------------------------------
-// Value MUST be an integer
-// 0 will always return the default value
-// 1 will return the provider's value 1% of the time
-// 100 will return the provider's value 100% of the time
-// --------------------------------------------------------
-// Default weight is 0.5 (50% chance).
-
-// First set examples
-$faker->optional($weight = 0.05)->randomDigit; // 95% chance of NULL, 5% chance of provider
-$faker->optional($weight = 0.1)->randomDigit; // 90% chance of NULL, 10% chance of provider
-$faker->optional($weight = 0.9)->randomDigit; // 10% chance of NULL, 90% chance of provider
-
-// Second set examples
+// optional() accepts a weight argument (Between 0 and 100) to specify the probability of receiving the default value.
 $faker->optional($weight = 1)->randomDigit; // 1% chance of NULL, 99% chance of provider
 $faker->optional($weight = 10)->randomDigit; // 90% chance of NULL, 10% chance of provider
 $faker->optional($weight = 100)->randomDigit; // 0% chance of NULL, 100% chance of provider
 
-// PITFALLS
-// --------------------------------------------------------
-// Get a half a percent (0.5%) chance
-$faker->optional($weight = 0.5)->randomDigit; // This is not 0.5%, this is 50% because it falls within first set
-$faker->optional($weight = 0.005)->randomDigit; // This is a 99.5% chance of NULL or 0.5% chance of provider
-// --------------------------------------------------------
-// Get 100% chance using a true/false or 0/1 approach
-$weight = $faker->randomElement([0, 1]);
-$faker->optional($weight)->randomDigit; // This will yield randomDigit 0% or 1% of chances, not 0% or 100%
-$faker->optional($weight * 100)->randomDigit; // This will yield randomDigit 0% or 100% of chances
-// --------------------------------------------------------
-// Get 4.5% chance
-$faker->optional($weight = 4.5)->randomDigit; // This will always yield default/null value, fractional support is between 0 and 0.99~
-$faker->optional($chance * 0.045)->randomDigit; // This will yield default/null 95.5% of time or the provider's value 4.5% of the time
+// Deprecated: You can use a fractional between 0 and 0.99~ to express the weight, this feature will be removed in v2.0.0
+// Note that the weight must be a number between 0 and 100 and must be integral (No fractions)
 
 // optional() accepts a default argument to specify the default value to return.
 // Defaults to NULL.
-
-// First set examples
-$faker->optional($weight = 0.5, $default = false)->randomDigit; // 50% chance of FALSE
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -343,15 +343,60 @@ for ($i = 0; $i < 10; $i++) {
...
+$faker->optional($chance * 0.045)->randomDigit; // This will yield default/null 95.5% of time or the provider's value 4.5% of the time
 
 // optional() accepts a default argument to specify the default value to return.
 // Defaults to NULL.
+
+// First set examples
 $faker->optional($weight = 0.5, $default = false)->randomDigit; // 50% chance of FALSE
 $faker->optional($weight = 0.9, $default = 'abc')->word; // 10% chance of 'abc'
 
+// Second set examples
```
localheinz: Can you add a note that this behaviour will be removed in `2.0`?

See https://github.com/fzaninotto/Faker/blob/3f74f49dd7c53af959b3708a64d9232db271f985/src/Faker/Provider/Base.php#L547-L551.

readme.md
```diff
@@ -343,15 +343,60 @@ for ($i = 0; $i < 10; $i++) {
...
+// PITFALLS
+// --------------------------------------------------------
+// Get a half a percent (0.5%) chance
+$faker->optional($weight = 0.5)->randomDigit; // This is not 0.5%, this is 50% because it falls within first set
+$faker->optional($weight = 0.005)->randomDigit; // This is a 99.5% chance of NULL or 0.5% chance of provider
+// --------------------------------------------------------
+// Get 100% chance using a true/false or 0/1 approach
+$weight = $faker->randomElement([0, 1]);
+$faker->optional($weight)->randomDigit; // This will yield randomDigit 0% or 1% of chances, not 0% or 100%
+$faker->optional($weight * 100)->randomDigit; // This will yield randomDigit 0% or 100% of chances
```
localheinz: Perhaps we should add a note that this behaviour will be removed in `2.0`?

See https://github.com/fzaninotto/Faker/blob/3f74f49dd7c53af959b3708a64d9232db271f985/src/Faker/Provider/Base.php#L547-L551.

readme.md
```diff
@@ -343,15 +343,60 @@ for ($i = 0; $i < 10; $i++) {
...
+// --------------------------------------------------------
+// 2 value sets are possible
+// --------------------------------------------------------
+// First set is between 0 and 1 (Fractional representation)
+// --------------------------------------------------------
+// 0 will always return the default value
+// Anything else BELOW 1 will be converted to a percentage
+// 
+// --------------------------------------------------------
+// Second set is between 1 and 100 (Percentage representation)
```
localheinz: Perhaps we should add a note that this behaviour will be removed in `2.0`?

See https://github.com/fzaninotto/Faker/blob/3f74f49dd7c53af959b3708a64d9232db271f985/src/Faker/Provider/Base.php#L547-L551.

localheinz (commented):
This looks good to me.

However, since the behaviour using fractional values between `0` and `1` is to be removed for `2.0`, I'm not sure whether it makes sense to document here. If so, then we should probably add a note that this behaviour will be removed for `2.0`.

On the other hand, with [Designing for People Who Have Better Things To Do With Their Lives](https://www.joelonsoftware.com/2000/04/26/designing-for-people-who-have-better-things-to-do-with-their-lives/) in mind, perhaps it's best to leave out the documentation for the old behaviour.

readme.md
fzaninotto: I disagree. This kind of mention should only appear in the documentation in an Upgrade guide for 2.0. No need to add mental overhead to readers about something that doesn't concern them.

fzaninotto (changes requested):
We should not document the deprecated version at all. Having two versions leaves a choice to users which they don't know how to make. Please simply replace the outdated version with the new one.

crazycodr (commented):
Resolved, please review again

readme.md
```diff
@@ -342,15 +342,18 @@ for ($i = 0; $i < 10; $i++) {
 }
 print_r($values); // [1, 4, null, 9, 5, null, null, 4, 6, null]
 
-// optional() accepts a weight argument to specify the probability of receiving the default value.
-// 0 will always return the default value; 1 will always return the provider. Default weight is 0.5 (50% chance).
-$faker->optional($weight = 0.1)->randomDigit; // 90% chance of NULL
-$faker->optional($weight = 0.9)->randomDigit; // 10% chance of NULL
+// optional() accepts a weight argument (Between 0 and 100) to specify the probability of receiving the default value.
```
fzaninotto:
```suggestion
// optional() accepts a weight argument (between 0 and 100) to specify the probability of receiving the default value.
```

readme.md
```diff
@@ -342,15 +342,18 @@ for ($i = 0; $i < 10; $i++) {
...
+$faker->optional($weight = 1)->randomDigit; // 1% chance of NULL, 99% chance of provider
+$faker->optional($weight = 10)->randomDigit; // 90% chance of NULL, 10% chance of provider
```
fzaninotto: the comments are either wrong or confusing

readme.md
```diff
@@ -342,15 +342,18 @@ for ($i = 0; $i < 10; $i++) {
...
+$faker->optional($weight = 100)->randomDigit; // 0% chance of NULL, 100% chance of provider
+
+// Deprecated: You can use a fractional between 0 and 0.99~ to express the weight, this feature will be removed in v2.0.0
```
fzaninotto: don't document it at all (remove that line)

fzaninotto (changes requested):

readme.md
fzaninotto: still misleading. Please read your 3 examples, they don't make sense.

readme.md
fzaninotto: you can keep the code about integers, but remove the part about the old syntax

fzaninotto (approved):

localheinz (approved):
👍

PR #1751, opened 2019-08-22
Fix Iranian phone number prefix

### LINKED ISSUE #1706: fa_IR not worked for phone number
Hi
when I create the object with the fa_IR locale the generated phone number is not from IR:

My code:
```
$faker = Factory::create('fa_IR'); 
echo $faker->phoneNumber; # result: 02187852819 | 02123963771
```
but when used the `Faker\Provider\fa_IR\PhoneNumber` it is currect:
```
echo PhoneNumber::mobileNumber();  # result: 09379448873 | 09176712858
````

### PR DESCRIPTION
**Pull request by ghost**
- user can't generate the correct phone number prefix when doing it with
the Generator

- it was revealed that Generator was using the wrong fa_ir locale format

- this is to fix #1706

### DIFF
--- a/src/Faker/Provider/fa_IR/PhoneNumber.php
+++ b/src/Faker/Provider/fa_IR/PhoneNumber.php
@@ -7,7 +7,6 @@ class PhoneNumber extends \Faker\Provider\PhoneNumber
     /**
      * @link https://fa.wikipedia.org/wiki/%D8%B4%D9%85%D8%A7%D8%B1%D9%87%E2%80%8C%D9%87%D8%A7%DB%8C_%D8%AA%D9%84%D9%81%D9%86_%D8%AF%D8%B1_%D8%A7%DB%8C%D8%B1%D8%A7%D9%86#.D8.AA.D9.84.D9.81.D9.86.E2.80.8C.D9.87.D8.A7.DB.8C_.D9.87.D9.85.D8.B1.D8.A7.D9.87
      */
-
     protected static $formats = array(
         '0910#######',//mci
         '0911#######',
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by CaptFurball** [comment]
Re-creating this pull request as the previous forked repository was deleted

File: src/Faker/Provider/fa_IR/PhoneNumber.php [hunk 1]
```diff
@@ -7,12 +7,8 @@ class PhoneNumber extends \Faker\Provider\PhoneNumber
     /**
      * @link https://fa.wikipedia.org/wiki/%D8%B4%D9%85%D8%A7%D8%B1%D9%87%E2%80%8C%D9%87%D8%A7%DB%8C_%D8%AA%D9%84%D9%81%D9%86_%D8%AF%D8%B1_%D8%A7%DB%8C%D8%B1%D8%A7%D9%86#.D8.AA.D9.84.D9.81.D9.86.E2.80.8C.D9.87.D8.A7.DB.8C_.D9.87.D9.85.D8.B1.D8.A7.D9.87
      */
+
```
localheinz: Can you remove this extra empty line, please?

src/Faker/Provider/fa_IR/PhoneNumber.php
```diff
@@ -39,8 +35,9 @@ class PhoneNumber extends \Faker\Provider\PhoneNumber
         '0937#######',
         '0937#######',
     );
+
     public static function mobileNumber()
     {
-        return static::numerify(static::randomElement(static::$mobileNumberPrefixes));
+        return static::numerify(static::randomElement(static::$formats));
```
localheinz: Does this change really make sense?

Also see 

* https://en.wikipedia.org/wiki/Telephone_numbers_in_Iran#Landlines
* https://en.wikipedia.org/wiki/Telephone_numbers_in_Iran#Mobile_phones

localheinz (commented):
Thank you for your contribution, @CaptFurball!

I'm not sure about the changes, can you take a look at the comments, please?

src/Faker/Provider/fa_IR/PhoneNumber.php
CaptFurball: Okay removed

src/Faker/Provider/fa_IR/PhoneNumber.php
CaptFurball: Ah I see..

021, 026 and 031 are landline numbers from when called from within Iran. So logically the $formats number prefixes are equally valid. 

However, the "mobileNumber()" method are using "$formats" which contain only landline numbers.

Moving forward, I think what is right is to create another function call "landlineNumber()" to generate landline numbers, both with or without country code. What do you think? Should I go with this approach?

TBH, I'm pretty new here, for now I am just thinking what make sense to me. If my approach wrong, please correct me so I can learn from the mistake.

Hope to hear from you soon.

ghost (commented):
> hanges, can you take a look at the comments, please?

Hi, I have made another comment. I am not too sure how I should approach this, I have a suggestion, in the comments, will you take a look?

PR #1756, opened 2019-08-23
Enhancement: Clean up .gitattributes

localheinz:
This PR

* [x] sorts entries in `.gitattributes` by kind (directories first, then files) and then alphabetically
* [x] removes a non-applicable entry
* [x] indents


localheinz:

PR #1757, opened 2019-08-23
Fix: No need to update composer itself

localheinz:
This PR

* [x] removes an unnecessary invocation of `composer self-update`

💁‍♂ It's already updated, see for example

* https://travis-ci.org/fzaninotto/Faker/jobs/575886873#L129
* https://travis-ci.org/fzaninotto/Faker/jobs/575886873#L151


localheinz:

PR #1758, opened 2019-08-23
Enhancement: Disable Xdebug as early as possible

localheinz:
This PR

* [x] disables Xdebug as early as possible on Travis

💁‍♂ We currently don't need it, and this potentially speeds up builds. For reference, see https://docs.travis-ci.com/user/languages/php/#disabling-preinstalled-php-extensions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,9 @@ cache:
   directories:
     - $HOME/.composer/cache
 
+before_install:
+  - phpenv config-rm xdebug.ini || true
+
 before_script:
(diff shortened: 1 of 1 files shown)


localheinz:
@pimjansen 

Unless I've missed something, it appears we are currently not collecting code coverage. We can easily keep Xdebug enabled for a single build (rather than all builds, unless we have different execution paths for different PHP versions) when we decide to collect and display it (perhaps using something like https://codecov.io).

What do you think?

localheinz:

PR #1759, opened 2019-08-24
Fix: Allow nightly builds to fail

localheinz:
This PR

* [x] allows `nightly` builds to fail

Somewhat related to #1758, as 3rd-party extensions (such as Xdebug) are not installed on nightly builds (see https://docs.travis-ci.com/user/languages/php/#nightly-builds).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,11 @@ php:
     - 7.1
     - 7.2
     - 7.3
-    - 7.4snapshot
+    - nightly
+
+matrix:
+    allow_failures:
+        - nightly
 
 cache:
   directories:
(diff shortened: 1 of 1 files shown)


pimjansen:
@localheinz i think we should keep the nightly in there as well for any future changes isnt it? Since 7.5 and 8.0 will also be there at some point. Further we could also test for the next patch releases on php on the nightly builds as well

localheinz:
@pimjansen 

I updated the PR to allow `nightly` builds to fail instead - what do you think?

pimjansen:
@localheinz yeah its better indeed i think. We can then easily add 7.4 in november when its released where now the 7.4 snapshot and the nightly wouldnt be that much difference i suppose

localheinz:

PR #1761, opened 2019-08-24
Enhancement: Introduce build and help targets

localheinz:
This PR

* [x] declares all phony targets at the start of the `Makefile`
* [x] keeps other targets sorted by name
* [x] introduces a `build` target which depends on `fix` and `test`
* [x] introduces a `help` target which displays a list of targets with descriptions (when they have been provided as comments)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,5 @@
 .PHONY: fix sniff test
 
-vendor/autoload.php:
-	composer install --no-interaction --prefer-dist
-
 fix: vendor/autoload.php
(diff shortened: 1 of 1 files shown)


fzaninotto:
I'd rather have `make` display a lost of all available commands. See https://github.com/marmelab/react-admin/blob/810e2e4f76190caa13c3d812056eff2b946007b9/Makefile#L1-L4 for an example.

localheinz:
@fzaninotto 

That's a great idea - I have adjusted this PR accordingly!

PR #1762, opened 2019-08-24
Enhancement: Assert that imageUrl can be used with gray flag

localheinz:
This PR

* [x] asserts that `Image::imageUrl()` can be used with `$gray` flag and produces different results depending on its value

Somewhat related to #1714.


fzaninotto:
Nice!

localheinz:

PR #1765, opened 2019-08-25
Enhancement: Add .editorconfig

localheinz:
This PR

* [x] adds a basic `.editorconfig` configuration
* [x] consistently indents `.travis.yml` with 2 spaces

💁‍♂ For reference, see https://editorconfig.org.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+root = true
+
+[*]
+charset = utf-8
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.yml]
+indent_size = 2
+
+[Makefile]
+indent_style = tab
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,5 @@
 /test/              export-ignore
+/.editorconfig      export-ignore
 /.gitattributes     export-ignore
 /.gitignore         export-ignore
 /.travis.yml        export-ignore
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,25 +3,25 @@ language: php
 dist: precise
 
 php:
-    - 5.3
-    - 5.4
-    - 5.5
-    - 5.6
-    - 7.0
-    - 7.1
-    - 7.2
-    - 7.3
-    - nightly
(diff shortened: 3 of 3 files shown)


.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+
+[*]
+charset = utf-8
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.yml]
+indent_size = 2
```
localheinz: Note that [`.travis.yml`](https://github.com/fzaninotto/Faker/blob/1273b5dd71b42cd7e0d9a8551cb0a5675ebc117c/.travis.yml) is currently indented with 2 and 4 spaces. Personally, I'd prefer to use 2 spaces for indentation of `yml` files, as otherwise indentation and wrapping of nested objects is a bit weird.

I'd follow up (or push into this branch if you prefer) an adjusted `.travis.yml` with consistent indentation.

pimjansen: Agree, lets reindent it

fzaninotto:
I'm -1 on this one. We can't assume anything about the editor the Faker developers will use.

pimjansen:
@fzaninotto i disagree though. The editorconfig does not force anything initially but it helps syncing the project on developer side to match the project expectations.

The PSR standard will force spaces instead of tabs and an indent of 4. With the config we can prevent these unneeded mistakes

localheinz:
@fzaninotto 

As you can see on https://editorconfig.org/#download, this is supported by quite a range of editors and IDEs. That is, there’s no harm done when merging this in and someone uses an editor or an IDE that does not support it. 

Other than that I agree that editor-specific configuration (for example, rules in `.gitignore` and similar) should be kept out of the code base.

fzaninotto:
The list you're linking to shows that this isn't a standard yet: the most popular IDEs (PHPStorm, TextMate, Atom, SublimeText, VSCode) require a plugin for this to work. 

As it's not a standard, I still believe it's IDE specific, and therefore should not go into our code.

localheinz:
@fzaninotto 

No problem!

PR #1545, opened 2018-08-11
fix austrian postal codes starting with invalid 0

mpge:
Temp fix for https://github.com/fzaninotto/Faker/issues/1544 - should probably implement more strict values in the future

PR #1766, opened 2019-08-25
Enhancement: Assert that postcode generated by Austrian Address provider matches format

localheinz:
This PR

* [x] asserts that the `postcode` formatter of the Austrian `Address` provider returns a `string` matching the expected format

Follows #1545.


localheinz:

Issue #1523, opened 2018-07-10
Literal parentheses not outputting with regexify


marcost5m:
Doing something like this in a model factory in Laravel; I can't get the parentheses to output:

`$faker->unique()->regexify('\(\d{3}\) \d{3}-\d{4}')`

ppelgrims:
That might be solved with #1515

icanhazstring:
Sadly, even with #1515 this problem is not solved.
The parentheses are still not printed.

icanhazstring:
@localheinz @pimjansen using this code above opens a new issue.
Seems like the old `array_replace(['(', ')', '', ...])` did solve a some misconfigured regexify calls.

For example:
`Provider/fr_FR/PhoneNumber.php:117`

The regex here is totally screwed:
```
$regex = '([012]{1}\d{1}|(9[1-357-9])( \d{2}){3}';
                         ^- this parentheses is too much here, since the test matches it without it
```

As mentioned, the old code did "fix" it behind the scenes.

icanhazstring:
@fzaninotto think we can close this one as well.

PR #1768, opened 2019-08-25
Fix: Literal parentheses not outputting with regexify

icanhazstring:
Should solve #1523 

This improves the grouped regex matching for regexify.
There is now the possibility for the input regex to contain escaped
parentheses which will stay in the final result.

### DIFF
--- a/src/Faker/Provider/fr_FR/PhoneNumber.php
+++ b/src/Faker/Provider/fr_FR/PhoneNumber.php
@@ -116,7 +116,7 @@ class PhoneNumber extends \Faker\Provider\PhoneNumber
      */
     public function phoneNumber08WithSeparator()
     {
-        $regex = '([012]{1}\d{1}|(9[1-357-9])( \d{2}){3}';
+        $regex = '([012]{1}\d{1}|9[1-357-9])( \d{2}){3}';
         return $this->regexify($regex);
     }
 
--- a/test/Faker/Provider/BaseTest.php
+++ b/test/Faker/Provider/BaseTest.php
@@ -633,7 +633,7 @@ class BaseTest extends TestCase
                 'input' => '/\w/'
             ),
             'replace meta sequence with ascii' => array(
-                'expected' => '/[[:alnum:]]/',
+                'expected' => '/[0-9]|[a-z]|[~!"#$%&\'()*+,-.\/:;<=>?@\[\]\\\\^_`]/i',
                 'input' => '/./'
             ),
             'superfluous backslashes should be removed' => array(
--- a/test/Faker/Provider/BaseTest.php
+++ b/test/Faker/Provider/BaseTest.php
@@ -2,6 +2,7 @@
 
 namespace Faker\Test\Provider;
 
+use Faker\Provider\Base;
 use Faker\Provider\Base as BaseProvider;
 use PHPUnit\Framework\TestCase;
 use Traversable;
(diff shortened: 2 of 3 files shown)


src/Faker/Provider/fr_FR/PhoneNumber.php
```diff
@@ -116,7 +116,7 @@ public function phoneNumber08()
      */
     public function phoneNumber08WithSeparator()
     {
-        $regex = '([012]{1}\d{1}|(9[1-357-9])( \d{2}){3}';
+        $regex = '([012]{1}\d{1}|9[1-357-9])( \d{2}){3}';
```
icanhazstring: This needed to be changed since the test on this method tested with another regex.
The previous `array_replace(['(', ')'], '', ...)` solved this problem behind the scenes.

https://github.com/fzaninotto/Faker/blob/master/test/Faker/Provider/fr_FR/PhoneNumberTest.php#L56

test/Faker/Provider/BaseTest.php
```diff
@@ -565,6 +566,31 @@ public function testRandomElements()
...
+     *
+     * @param $expectedRegex
+     * @param $input
+     */
+    public function testRegexify($expectedRegex, $input)
+    {
+        $this->assertRegExp($expectedRegex, BaseProvider::regexify($input));
+    }
+
+    public function testRegexifyDataProvider()
```
icanhazstring: I only covered the new behavior.
Since the `regexify` is super complex, we should think about splitting the single steps of regexify into seperate methods of some sort. This way we can provide more tests for each step in there.

test/Faker/Provider/BaseTest.php
```diff
@@ -565,6 +566,31 @@ public function testRandomElements()
         $this->assertCount(3, $allowDuplicates);
         $this->assertContainsOnly('string', $allowDuplicates);
     }
+
+    /**
+     * @dataProvider testRegexifyDataProvider
+     *
+     * @param $expectedRegex
```
localheinz: @icanhazstring 

What do you think about adding types for the `@param` annotations?

test/Faker/Provider/BaseTest.php
localheinz: Perhaps we should add some tests for the existing behavior before we merge this in, what do you think?

test/Faker/Provider/BaseTest.php
icanhazstring: Sure, will add them

test/Faker/Provider/BaseTest.php
icanhazstring: Will try and figure some out according to the comments in the function itself. But these will merely scratch a good coverage.

icanhazstring: Alright, added as much tests as possible. Problem is the complexity of the method itself. But this should cover a good portion of what is happening inside. Now I hope we can refactor this method with more test coverage to not break things ;)

fzaninotto:
Thanks for the patch. 

I'm honestly not very enthusiastic about complexifying the regex as you did. The code will be harder to read and maintain, for a short benefit - the `regexify` formatter will always be very limited. Also, I'm concerned about the performance impact, because `regexify` is already one of the slowest Faker formatters. 

So I'd rather keep the `regexify` formatter with its limitations (already documented), and encourage you to publish your own regex formatter with a much larger coverage of the regular expression grammar - which, I think, is a project by itself.

icanhazstring:
@localheinz there are also numerous other open PR concerning `regexify`.
#1453, #1624, and #1515

The all focus around the same changes I did.
How to we solve this?

icanhazstring:
Thanks for clarification. Should we then close the linked PRs as well including their issues?

I would create a new issue, with the information provided, to link to those issues for keeping a small feature list. So we can then easily plan the new regex feature.

What do you think?

fzaninotto:
@icanhazstring you're right, I closed the 3 PRs you linked. 

As for creating an issue with the remaining issues with regexify, let me reiterate: Faker won't implement a more complex regexify formatter, it's up to the community to publish it as a third party package (not in the Faker code).

icanhazstring:
> it's up to the community to publish it as a third party package
Alright, got it. Then I put it on my TODO ;)

PR #1770, opened 2019-08-26
Enhancement: Consistently use composer package name for link content

localheinz:
This PR

* [x] consistently uses the `composer` package name as link content for links to 3rd party libraries 

💁‍♂ I'd also suggest 

- consistently liking to GitHub repositories (rather than in at least one example, linking to https://packagist.org)
- perhaps removing abandoned packages
- splitting the list (perhaps into sections _Framework Integrations_ and _Providers_)
- keeping the list(s) sorted by package name

We could also start adding GitHub issue and pull request templates, and add a template for adding a library that gives suggestions, what do you think?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1688,33 +1688,33 @@ echo $faker->VAT; //23456789
 ## Third-Party Libraries Extending/Based On Faker
 
 * Symfony2 bundles:
-  * [BazingaFakerBundle](https://github.com/willdurand/BazingaFakerBundle): Put the awesome Faker library into the Symfony2 DIC and populate your database with fake data.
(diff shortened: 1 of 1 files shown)


pimjansen:
@localheinz agree with all the bullets. I think the templates could really help. Also the template how to report a bug since we miss a lot of code snippets and such

localheinz:
@pimjansen 

Personally, I'd prefer to keep the scope of pull requests limited to one change - what do you think?

pimjansen:
> @pimjansen
> 
> Personally, I'd prefer to keep the scope of pull requests limited to one change - what do you think?

Totally agree, was just wondering since you commented it here. So this PR further looks ok!

localheinz:

PR #1772, opened 2019-08-26
Fix: Name of package

localheinz:
This PR

* [x] fixes the name of a package

Follows #1770.

🤦‍♂ Sorry!


localheinz:

PR #1773, opened 2019-08-26
Fix: Split item into two items

localheinz:
This PR

* [x] splits an item referencing two packages into two items

Follows #1770.


fzaninotto:
-1 for this one. Reading about two packages with exactly the same description will only bring confusion.

PR #1775, opened 2019-08-26
Fix: Remove duplicate values

localheinz:
This PR

* [x] removes duplicate values from properties in providers
* [x] fixes tests failing because the data set has been modified

💁‍♂ I hacked around a bit with [`rector/rector`](https://github.com/rectorphp/rector), might push something later into [`localheinz/faker-rector`](http://github.com/localheinz/faker-rector).


src/Faker/Provider/Internet.php
```diff
@@ -5,7 +5,7 @@
 class Internet extends Base
 {
     protected static $freeEmailDomain = array('gmail.com', 'yahoo.com', 'hotmail.com');
-    protected static $tld = array('com', 'com', 'com', 'com', 'com', 'com', 'biz', 'info', 'net', 'org');
```
localheinz: Not even sure why there are duplicates - is there some reason I'm unaware of, @fzaninotto?

OskarStark: Just thinking loud... could it be, that this helps bringing some things up more often by the faker? I mean, if you fake a domain it’s 40/50/60% a toplevel domain with .com? That would make sense to me, same for name formats which are more common in some situations/countries

src/Faker/Provider/en_ZA/Address.php
```diff
@@ -5,10 +5,10 @@
 class Address extends \Faker\Provider\en_US\Address
 {
     protected static $cityPrefix = array('North', 'East', 'West', 'South', 'New', 'Lake', 'Port');
-    protected static $citySuffix = array('fontein','town', 'ton', 'land', 'ville', 'berg', 'burgh', 'borough', 'bury', 'view', 'port', 'mouth', 'stad', 'furt', 'chester', 'mouth', 'fort', 'haven', 'side', 'shire');
+    protected static $citySuffix = array('fontein','town', 'ton', 'land', 'ville', 'berg', 'burgh', 'borough', 'bury', 'view', 'port', 'mouth', 'stad', 'furt', 'chester', 'fort', 'haven', 'side', 'shire');
```
OskarStark: Will you run PHP CS fixer later? Short array syntax + space after comma?

src/Faker/Provider/Internet.php
fzaninotto: indeed, the repetition is intended

fzaninotto:
This is a misunderstanding of how Faker works. Repeating the same data or pattern increases the likeliness that Faker chooses it. It's a way to push certain data or patterns on purpose, to better match the real world.

Removing the duplication will not only remove that matching, it will also break unit tests on third-party projects based on Faker data for no valid reason. 

We should avoid modifying the datasets as much as possible.

src/Faker/Provider/en_ZA/Address.php
localheinz: Can't do short array syntax yet as we still support PHP 5.3, see https://github.com/fzaninotto/Faker/blob/768354c630e9f3644fb1cb3c6d6bef0a1b45d7e9/composer.json#L13.

localheinz:
@fzaninotto 

Thank you for the explanation, I indeed was unaware!

src/Faker/Provider/en_ZA/Address.php
OskarStark: But its outdated and should be dropped in a minor version

PR #1776, opened 2019-08-26
Add hslColor to ColorProvider

### LINKED ISSUE #1720: Generate HSL color
I've checked the docs, and scoured this repo, but I can't seem to find the command for generating HSL colors.

This commit looks like it adds tests for it: https://github.com/stympy/faker/pull/310, but having tried `$faker->hslcolor`, `$faker->hslacolor`, `$faker->hslColor`, `$faker->hslCssColor`
I can't seem to work it out?

This file doesn't seem to have a command for it: So I'm guessing it's not available: https://github.com/fzaninotto/Faker/blob/eef05943cd5f9b00edf81daac1f1ef28b16942d6/src/Faker/Provider/Color.php

Would this be of use https://gist.github.com/bedeabza/10463089 ?

icanhazstring:
This will add `hslColor` and `hslColorAsArray` for `ColorProvider`

Closes #1720

### DIFF
--- a/src/Faker/Provider/Color.php
+++ b/src/Faker/Provider/Color.php
@@ -113,4 +113,26 @@ class Color extends Base
     {
         return static::randomElement(static::$allColorNames);
     }
+
+    /**
+     * @example '340,50,20'
+     */
+    public static function hslColor()
+    {
+        return sprintf(
+            '%s,%s,%s',
+            static::numberBetween(0, 360),
+            static::numberBetween(0, 100),
+            static::numberBetween(0, 100)
+        );
+    }
+
+    public static function hslColorAsArray()
+    {
+        return array(
+            static::numberBetween(0, 360),
+            static::numberBetween(0, 100),
+            static::numberBetween(0, 100)
+        );
+    }
 }
--- a/src/Faker/Provider/Color.php
+++ b/src/Faker/Provider/Color.php
@@ -116,6 +116,7 @@ class Color extends Base
 
     /**
      * @example '340,50,20'
+     * @return string
      */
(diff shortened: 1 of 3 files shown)


src/Faker/Provider/Color.php
```diff
@@ -113,4 +113,26 @@ public static function colorName()
     {
         return static::randomElement(static::$allColorNames);
     }
+
+    /**
+     * @example '340,50,20'
```
localheinz: What do you think about adding a `@return` annotation with `array` as type here?

src/Faker/Provider/Color.php
```diff
@@ -113,4 +113,26 @@ public static function colorName()
...
+    {
+        return sprintf(
+            '%s,%s,%s',
+            static::numberBetween(0, 360),
+            static::numberBetween(0, 100),
+            static::numberBetween(0, 100)
+        );
+    }
+
+    public static function hslColorAsArray()
```
localheinz: What do you think about adding a `@return` annotation with `array` as type here?

icanhazstring:
> This looks good to me!
> 
> One more thing though before I can approve: Can you update [`readme.md`](https://github.com/fzaninotto/Faker/blob/master/readme.md#fakerprovidercolor), please?

Sure thing, totally forgot about that. 👍

src/Faker/Provider/Color.php
icanhazstring: Its more like `string`. But I added it :)

localheinz:
Thank you, @icanhazstring!

Issue #1616, opened 2018-11-12
Deprecated  $schema->column in CakePHP


gus0074:
Hello @fzaninotto, sorry I'm a newbie, CakePHP throw an error loop. Fixed it changing line 45 in:
cakephp\your-project\vendor\fzaninotto\faker\src\Faker\ORM\CakePHP\ColumnTypeGuesser.php
from:
$columnData = $schema->column($column);
to:
$columnData = $schema->getColumn($column);

icanhazstring:
@pimjansen @localheinz could be closed then with #1777 merged

PR #1777, opened 2019-08-26
FIx: Avoid deprecation warning in newer CakePHP version

icanhazstring:
This will avoid deprecation warnings by accessing `getColumn` if the method exists. Otherwise access the old `column` method.

Solve #1616


icanhazstring:
Thanks :)

PR #1779, opened 2019-08-27
Enhancement: Add issue template for bug report

localheinz:
This PR

* [x] provides a GitHub issue template for a bug report

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,0 +1,35 @@
+---
+name: Bug
+about: Something does not work as expected
+----
+
+### Summary
+
+<!-- provide a summary here -->
+
+### Versions
+
+<!-- Please provide the versions of PHP as well as `fzaninotto/faker` on which the issue has been observed -->
+
+|                     | Version |
+|:--------------------|:--------|
+| PHP                 | x.y.z   |
+| `fzaninotto/faker`  | x.y.z   |
+
+### Self-enclosed code snippet for reproduction
+
+```php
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Affected System
+
+#### `php -v`
+
+```txt
+<!-- please replace this with the output of running `php -v` --> 
+```
+
+#### `composer show | grep 'fzaninotto/faker'`
```
pimjansen: I would just add the description on whats needed here:
`#### Faker version`

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,40 @@
+---
+name: Bug
+about: Something does not work as expected
+----
+
+### Summary
+
+<!-- provide a summary here -->
```
pimjansen: Somehow i dont see this in the markdown preview, is that just a `github` thing or do we need to push it in a code block?

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Affected System
+
```
pimjansen: Maybe we should explain them atleast what is expected here like in the other blocks as well

.github/ISSUE_TEMPLATE/bug.md
localheinz: It's rendered as an HTML comment!

PR #1780, opened 2019-08-27
Added some more information for autoloading in the readme

pimjansen:
Added some more info about autoloading since there are some questions as issues on this. Since 99% of the people will install the package via composer anyways i think its more easy

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -51,14 +51,29 @@ composer require fzaninotto/faker
 
 ## Basic Usage
 
+### Autoloading
+
+Faker supports both `PSR-0` as `PSR-4` autoloaders. 
+```php
+<?php
+# When installed via composer
+require_once 'vendor/autoload.php';
+```
+
+You can also load `Fakers` shipped `PSR-0` autoloader
+```php
+<?php
+# Load Fakers own autoloader
+require_once '/path/to/Faker/src/autoload.php';
+```
+
+*alternatively, you can use any another PSR-4 compliant autoloader*
+
+### Create fake data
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -51,14 +51,30 @@ composer require fzaninotto/faker
...
 require_once '/path/to/Faker/src/autoload.php';
-// alternatively, use another PSR-4 compliant autoloader
+```
 
+*alternatively, you can use any another PSR-4 compliant autoloader*
+
+### Create fake data
+Use `Faker\Factory::create()` to create and initialize a faker generator, which can generate data by accessing properties named after the type of data you want.
+
+###
```
fzaninotto: please remove this line

pimjansen: Ah nice find, missed it indeed

readme.md
```diff
@@ -51,14 +51,29 @@ composer require fzaninotto/faker
 
 ## Basic Usage
 
-Use `Faker\Factory::create()` to create and initialize a faker generator, which can generate data by accessing properties named after the type of data you want.
+### Autoloading
+
+Faker supports both `PSR-0` as `PSR-4` autoloaders.
```
localheinz: Not sure if users care about the kind of autoloader we use internally.

Perhaps we should point out two options:

- when installed via `composer`, require `vendor/autoload.php`
- otherwise, require `/path/to/Faker/src/autoload.php`

What do you think?

pimjansen: I do agree but this is a bit "as is" and i think we should not just wipe it. The focus now is on composer autoload and otherwise the custom one which is psr0.

But i would not be suprised if there are enough people around that just load it manually instead and that do not use composer yet.

localheinz:
Thank you, @pimjansen!

PR #1781, opened 2019-08-27
Added batch inserts for doctrine orm populate

pimjansen:
Added a counter a batch size to insert in batches instead of all at once on which doctrine orm will block itself due the fact it holds all in memory.

Works fine and it reduces the memory consumption a lot. The smaller the batches the less memory it will consume but will be less fast as well.

Only problem that still remains is the return of the PKs inserted though. I don't think there is a real solution for that when you want to keep track of so much data. Did a test with 100k records for this though meaning it holds 100k PKs in an array.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -409,6 +409,9 @@ Faker provides adapters for Object-Relational and Object-Document Mappers (curre
 
 To populate entities, create a new populator class (using a generator instance as parameter), then list the class and number of all the entities that must be generated. To launch the actual data population, call the `execute()` method.
 
+Note that some of the `populators` could require additional parameters. As example the `doctrine` populator has an option to specify
+its batchSize on how often it will flush the UnitOfWork to the database.
+
 Here is an example showing how to populate 5 `Author` and 10 `Book` objects:
 
 ```php
--- a/readme.md
+++ b/readme.md
@@ -452,6 +452,8 @@ print_r($insertedPKs);
 // )
 ```
 
+**Note:** Due to the fact that `Faker` returns all the primary keys inserted, the memory consumption will go up drastically when you do batch inserts due to the big list of data.
+
 In the previous example, the `Book` and `Author` models share a relationship. Since `Author` entities are populated first, Faker is smart enough to relate the populated `Book` entities to one of the populated `Author` entities.
 
 Lastly, if you want to execute an arbitrary function on an entity before insertion, use the fourth argument of the `addEntity()` method:
--- a/src/Faker/ORM/Doctrine/Populator.php
+++ b/src/Faker/ORM/Doctrine/Populator.php
@@ -3,6 +3,7 @@
 namespace Faker\ORM\Doctrine;
 
 use Doctrine\Common\Persistence\ObjectManager;
+use Faker\Generator;
 
 /**
  * Service class for populating a database using the Doctrine ORM or ODM.
@@ -10,20 +11,35 @@ use Doctrine\Common\Persistence\ObjectManager;
  */
 class Populator
 {
+    /** @var int  */
+    protected $batchSize;
+
+    /** @var Generator  */
     protected $generator;
+
+    /** @var ObjectManager|null  */
     protected $manager;
+
+    /** @var array  */
     protected $entities = array();
+
+    /** @var array  */
     protected $quantities = array();
+
+    /** @var array  */
     protected $generateId = array();
 
     /**
-     * @param \Faker\Generator $generator
+     * Populator constructor.
+     * @param Generator $generator
      * @param ObjectManager|null $manager
+     * @param int $batchSize
      */
-    public function __construct(\Faker\Generator $generator, ObjectManager $manager = null)
+    public function __construct(Generator $generator, ObjectManager $manager = null, $batchSize = 1000)
     {
         $this->generator = $generator;
         $this->manager = $manager;
+        $this->batchSize = $batchSize;
     }
 
     /**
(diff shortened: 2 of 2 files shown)


fzaninotto:
Well, the PK drawback is serious and it should be documented.

pimjansen:
Added some docs of the negative impact on bigger amounts of data

readme.md
```diff
@@ -452,6 +452,9 @@ print_r($insertedPKs);
 // )
 ```
 
+**Note:** due to the fact that `Faker` returns all the PKs inserted, the memory consumption will go up drasticly when you
```
localheinz: What do you think about modifying to this:

```suggestion
**Note:** Due to the fact that `Faker` returns all the primary keys inserted, the memory consumption will go up drastically when you 
```

Also see https://www.merriam-webster.com/dictionary/drastically.

src/Faker/ORM/Doctrine/Populator.php
```diff
@@ -55,6 +68,9 @@ public function addEntity($entity, $number, $customColumnFormatters = array(), $
     /**
      * Populate the database using all the Entity classes previously added.
      *
+     * Please note that large amounts of data will result in more memory usage since the the Populator will return
+     * all newly created PKs after executing
```
localheinz: What do you think about this?

```suggestion
     * all newly created primary keys after executing.
```

src/Faker/ORM/Doctrine/Populator.php
```diff
@@ -68,13 +84,24 @@ public function execute($entityManager = null)
             throw new \InvalidArgumentException("No entity manager passed to Doctrine Populator.");
         }
 
+        $insertCount = 0;
```
localheinz: Perhaps rename to `$entityCount`?

The entities will only have been inserted once `$entityManager->flush()` has been called. As long as that has not happened, the variable name is a bit misleading.

readme.md
```diff
@@ -452,6 +452,8 @@ print_r($insertedPKs);
 // )
 ```
 
+**Note:** Due to the fact that `Faker` returns all the primary keys inserted, the memory consumption will go up drastically when you
```
localheinz: ❓ 

Looks like we are missing something here!

😉

src/Faker/ORM/Doctrine/Populator.php
```diff
@@ -10,20 +10,33 @@
...
+
+    /** @var array  */
     protected $generateId = array();
 
     /**
      * @param \Faker\Generator $generator
      * @param ObjectManager|null $manager
      */
-    public function __construct(\Faker\Generator $generator, ObjectManager $manager = null)
+    public function __construct(\Faker\Generator $generator, ObjectManager $manager = null, $batchSize = 1000)
```
localheinz: Should we add a `@param` annotation for the `$batchSize` parameter?

src/Faker/ORM/Doctrine/Populator.php
```diff
@@ -68,13 +84,24 @@ public function execute($entityManager = null)
             throw new \InvalidArgumentException("No entity manager passed to Doctrine Populator.");
         }
 
+        $batchCountIterateCount = 0;
```
localheinz: Maybe we can even inline the variable `$batchCountIterateCount` and replace it with a query instead?

```diff
diff --git a/src/Faker/ORM/Doctrine/Populator.php b/src/Faker/ORM/Doctrine/Populator.php
index 692486f..9087c63 100644
--- a/src/Faker/ORM/Doctrine/Populator.php
+++ b/src/Faker/ORM/Doctrine/Populator.php
@@ -84,22 +84,24 @@ class Populator
             throw new \InvalidArgumentException("No entity manager passed to Doctrine Populator.");
         }

-        $batchCountIterateCount = 0;
         $insertedEntities = array();
+
         foreach ($this->quantities as $class => $number) {
             $generateId = $this->generateId[$class];
+
             for ($i=0; $i < $number; $i++) {
-                $batchCountIterateCount++;
                 $insertedEntities[$class][]= $this->entities[$class]->execute(
                     $entityManager,
                     $insertedEntities,
                     $generateId
                 );
-                if ($batchCountIterateCount % $this->batchSize === 0) {
+
+                if (count($insertedEntities) % $this->batchSize === 0) {
                     $entityManager->flush();
                     $entityManager->clear($class);
                 }
             }
+
             $entityManager->flush();
             $entityManager->clear($class);
         }
```

readme.md
fzaninotto: Right, please fix this sentence

src/Faker/ORM/Doctrine/Populator.php
fzaninotto: :+1:

pimjansen:
All comments have been handled and applied to the code

fzaninotto:
@pimjansen please don't force push. Just add commits to the PR so that we can see what you did in reaction to the comments. I don't mind having several commits in a PR.

Also, you haven't documented the `batchSize` anywhere. Considering this is a (slight) breking change, I suggest you mention it in the documentation.

jaapromijn:
Upgrading from 1.8 to 1.9, this commit seems to have broken our setup:
https://github.com/fzaninotto/Faker/commit/c49cd5438655043b487f4d986961746ae58edb3a

image

Using the following code:
```
        $users = $manager->getRepository(User::class)->findAll();

        $populator->addEntity(Status::class, 50, [
            'user' => function () use ($generator, $users) {
                $key = array_rand($users);

                return $users[$key];
            },
        ]);
```
This addition seems to be the problem:
https://github.com/fzaninotto/Faker/commit/c49cd5438655043b487f4d986961746ae58edb3a#diff-d38baf8c465e27d0cc5334cf5d8475aaR105

aless673:
Lost 2 hours to understand what was happening. Same here, 1.9 is now broken for using faker with doctrine relations...

PR #1782, opened 2019-08-27
Feature/composer lock

pimjansen:
As mentioned by the docs the lock file should be added to the repo in order to ensure that all packages stay equal during its lifetime:

https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
 vendor
-composer.lock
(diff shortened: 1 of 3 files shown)


pimjansen:
This is due the fact that there will be references more versions which the lock cant handle

PR #1529, opened 2018-07-17
Improve Dutch company names

koole:
Previously generated company names only consisted of one or two surnames and a suffix. This resulted in unrealistic and sometimes extremely long company names like "de la Fontaine und d’Harnoncourt Unverzagt Mohan EESV" and "Martena van Burmania Vegilin van Claerbergen".

The new faker generates realistic company names like "Fietsenmaker Zijlemans" and "Elektronicakampioen BV".

PR #1783, opened 2019-08-28
Enhancement: Slightly clean up Company provider

localheinz:
This PR

* [x] slightly cleans up the `Company` provider

Follows #1529.

💁‍♂ Probably best reviewed commit by commit.

### DIFF
--- a/src/Faker/Provider/nl_NL/Company.php
+++ b/src/Faker/Provider/nl_NL/Company.php
@@ -82,7 +82,7 @@ class Company extends \Faker\Provider\Company
                 break;
         }
 
-        if (static::numberBetween(0, 1)) {
+        if (0 !== static::numberBetween(0, 1)) {
             $companyName = $companyName . ' ' . static::randomElement(static::$companySuffix);
         }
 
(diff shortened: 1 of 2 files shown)


src/Faker/Provider/nl_NL/Company.php
```diff
@@ -36,47 +36,51 @@ class Company extends \Faker\Provider\Company
...
-                $companyName = static::randomElement(static::$product) . ' ' . static::randomElement(static::$type);
-                break;
-            case 1:
-                $companyName = static::randomElement(static::$product) . strtolower(static::randomElement(static::$type));
-                break;
-            case 2:
-                $companyName = static::randomElement(static::$store) . ' ' . $lastName;
-                break;
+
+        if (0 === $determinator) {
```
fzaninotto: what's wrong with switch / case?

localheinz: It's probably a matter of personal taste - personally I avoid to use it because

- it usually requires a lot more code
- it uses weak comparisons

But if you prefer, I can revert this change!

fzaninotto: Yes, I personally have nothing against switch/case, so please revert just that part

localheinz: Adjusted!

Issue #1626, opened 2018-11-29
Add "anyDateInFuture" and "anyDateInPast" functionality


tatitati:
It would be cool to have something like

$faker->anyDateInTheFuture()
$faker->anyDateInThePast()

icanhazstring:
@tatitati this is currenty possible using `DateTime::dateTimeBetween` which you can use the following:

```
$anyDateInTheFuture = DateTime::dateTimeBetween('now', '+100 year);
$anyDateInThePast = DateTime::dateTimeBetween('-100 year', 'now');
```

The two functions you suggested would be merely an alias. So I don't know about this one.
If, however, you would randomize the `min` or `max` using `randomDigit`, then this would make sense.

@localheinz @pimjansen what do you guys think?

PR #1786, opened 2019-09-02
Reword/Feature: Better naming of dateTime relative methods

icanhazstring:
This PR will address #1165 where we want to address a better naming for dateTime relative methods.

`dateTimeThisMonth` (and sorts) seems misleading as it pretends to return a value from *the current month* which is does not. With these changes we want to clarify by which method the user can expect the correct return value.

### Open discussion
This PR should also be used as an open discussion if we even should add the new methods, or simply clarify the naming of the old ones. This is why this PR does not yet contain the implementation of the new methods. If we agree upon adding the new methods I will add the implementation.

### The full feature list

Assuming the current date is `2019-09-02`. Here is a full list of methods and what you can expect from them:
- `dateTimePastCentury` datetime between `1919-09-02` and `2019-09-02`
- `dateTimePastDecade` datetime between `2009-09-02` and `2019-09-02`
- `dateTimePastYear` datetime between `2018-09-02` and `2019-09-02`
- `dateTimePastMonth` datetime between `2019-08-02` and `2019-09-02`
- `dateTimeLastCentury` datetime between `1900-01-01` and `1999-12-31`
- `dateTimeLastDecade` datetime between `2000-01-01` and `2009-12-31`
- `dateTimeLastYear` datetime between `2018-01-01` and `2018-12-31`
- `dateTimeLastMonth` datetime between `2019-08-01` and `2018-08-31`
- `dateTimeCurrentCentury` datetime between `2000-01-01` and `2019-12-31` (with `$now` as maximum)
- `dateTimeCurrentDecade` datetime between `2010-01-01` and `2019-12-31` (with `$now` as maximum)
- `dateTimeCurrentYear` datetime between `2019-01-01` and `2019-12-31` (with `$now` as maximum)
- `dateTimeCurrentMonth` datetime between `2019-09-01` and `2019-09-30` (with `$now` as maximum)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -187,10 +187,18 @@ Each of the generator properties (like `name`, `address`, and `lorem`) are calle
     time($format = 'H:i:s', $max = 'now') // '20:49:42'
     dateTimeBetween($startDate = '-30 years', $endDate = 'now', $timezone = null) // DateTime('2003-03-15 02:00:49', 'Africa/Lagos')
     dateTimeInInterval($startDate = '-30 years', $interval = '+ 5 days', $timezone = null) // DateTime('2003-03-15 02:00:49', 'Antartica/Vostok')
-    dateTimeThisCentury($max = 'now', $timezone = null)     // DateTime('1915-05-30 19:28:21', 'UTC')
-    dateTimeThisDecade($max = 'now', $timezone = null)      // DateTime('2007-05-29 22:30:48', 'Europe/Paris')
-    dateTimeThisYear($max = 'now', $timezone = null)        // DateTime('2011-02-27 20:52:14', 'Africa/Lagos')
-    dateTimeThisMonth($max = 'now', $timezone = null)       // DateTime('2011-10-23 13:46:23', 'Antarctica/Vostok')
+    dateTimePastCentury($max = 'now', $timezone = null)     // DateTime('1915-05-30 19:28:21', 'UTC')
+    dateTimePastDecade($max = 'now', $timezone = null)      // DateTime('2007-05-29 22:30:48', 'Europe/Paris')
+    dateTimePastYear($max = 'now', $timezone = null)        // DateTime('2011-02-27 20:52:14', 'Africa/Lagos')
+    dateTimePastMonth($max = 'now', $timezone = null)       // DateTime('2011-10-23 13:46:23', 'Antarctica/Vostok')
+    dateTimeCurrentCentury($max = 'now', $timezone = null)  // DateTime('2000-02-21 12:14:23', 'UTC')
+    dateTimeCurrentDecade($max = 'now', $timezone = null)   // DateTime('2010-03-02 12:14:23', 'Europe/Paris')
+    dateTimeCurrentYear($max = 'now', $timezone = null)     // DateTime('2019-05-30 12:14:23', 'Africa/Lagos')
+    dateTimeCurrentMonth($max = 'now', $timezone = null)    // DateTime('2019-09-01 12:14:23', 'Antarctica/Vostok')
+    dateTimeLastCentury($max = 'now', $timezone = null)     // DateTime('1900-02-21 12:14:23', 'UTC')
+    dateTimeLastDecade($max = 'now', $timezone = null)      // DateTime('2000-03-02 12:14:23', 'Europe/Paris')
+    dateTimeLastYear($max = 'now', $timezone = null)        // DateTime('2018-05-30 12:14:23', 'Africa/Lagos')
+    dateTimeLastMonth($max = 'now', $timezone = null)       // DateTime('2019-08-01 12:14:23', 'Antarctica/Vostok')
     amPm($max = 'now')                    // 'pm'
     dayOfMonth($max = 'now')              // '04'
     dayOfWeek($max = 'now')               // 'Friday'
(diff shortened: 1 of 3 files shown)


readme.md
```diff
@@ -199,8 +207,26 @@ Each of the generator properties (like `name`, `address`, and `lorem`) are calle
...
-
+    
 Methods accepting a `$timezone` argument default to `date_default_timezone_get()`. You can pass a custom timezone string to each method, or define a custom timezone for all time methods at once using `$faker::setDefaultTimezone($timezone)`.
+    
+#### Naming of relative dateTime methods
+In current state we have `dateTimeThisXXX` which is a misleading name. One would assume that this methods
+will return a calendar accurate datetime but rather it will return a random datetime in the "past" month (`$now - 1 month`).
+
+Assuming the current date is `2019-09-02`. Here is a full list of methods and what you can expect from them:
+- `dateTimePastCentury` datetime between `1919-09-02` and `2019-09-02`
```
localheinz: Not sure I have a misunderstanding, but personally, I would expect this method to return a value that corresponds to the past century, that is, the 20th century - so any date between `1900-01-01` and `1900-12-31`.
```

File: src/Faker/Provider/DateTime.php [hunk 2]
```diff
@@ -177,41 +177,105 @@ public static function dateTimeInInterval($date = '-30 years', $interval = '+5 d
...
      * @example DateTime('1964-04-04 11:02:02')
      * @return \DateTime
+     *
+     * @deprecated use dateTimePastCentury - The naming of this method is misleading since "this century" does not
+     *             point to the past century but rather to a random between 100 years ago and $now.
      */
     public static function dateTimeThisCentury($max = 'now', $timezone = null)
     {
-        return static::dateTimeBetween('-100 year', $max, $timezone);
+        return static::dateTImePastCentury($max, $timezone);
```
localheinz:
```suggestion
        return static::dateTimePastCentury($max, $timezone);
```

readme.md
icanhazstring: You are right. Will change it.

icanhazstring: Now that I thought about it again. This is the current bahavior. `PastCentury` simply goes back 100 years. So the stated value there is correct.

If you want 1900 to 2000 yiu need to use `LastCentury`.

src/Faker/Provider/DateTime.php
```diff
@@ -177,41 +177,105 @@ public static function dateTimeInInterval($date = '-30 years', $interval = '+5 d
...
+
+    /**
+     * Generate a random datetime from up to a hundred years ago
+     *
+     * @param \DateTime|int|string $max maximum timestamp used as random end limit, default to "now"
+     * @param string|null $timezone time zone in which the date time should be set, default to DateTime::$defaultTimezone, if set, otherwise the result of `date_default_timezone_get`
+     * @example DateTime('1964-04-04 11:02:02')
+     * @return \DateTime
+     */
+    public static function dateTImePastCentury($max = 'now', $timezone = null)
```
pimjansen: dateTimePastCentury

pimjansen (changes requested):

pimjansen:
@icanhazstring could you review the last bits so we can include it in the release?

icanhazstring:
Sure thing. What about the "new" methods, as I didn't implement them yet. Should we add them in a later release. Or with this one?

icanhazstring:
@pimjansen meaning these ones where I currently only added the README entries:

```
dateTimeCurrentCentury($max = 'now', $timezone = null)  // DateTime('2000-02-21 12:14:23', 'UTC')
    dateTimeCurrentDecade($max = 'now', $timezone = null)   // DateTime('2010-03-02 12:14:23', 'Europe/Paris')
    dateTimeCurrentYear($max = 'now', $timezone = null)     // DateTime('2019-05-30 12:14:23', 'Africa/Lagos')
    dateTimeCurrentMonth($max = 'now', $timezone = null)    // DateTime('2019-09-01 12:14:23', 'Antarctica/Vostok')
    dateTimeLastCentury($max = 'now', $timezone = null)     // DateTime('1900-02-21 12:14:23', 'UTC')
    dateTimeLastDecade($max = 'now', $timezone = null)      // DateTime('2000-03-02 12:14:23', 'Europe/Paris')
    dateTimeLastYear($max = 'now', $timezone = null)        // DateTime('2018-05-30 12:14:23', 'Africa/Lagos')
    dateTimeLastMonth($max = 'now', $timezone = null)       // DateTime('2019-08-01 12:14:23', 'Antarctica/Vostok')
```

pimjansen:
@icanhazstring i think we maybe should skip this and push this to 2.0 instead. Can you update the readme accordingly for the changes you did? After that we can merge

icanhazstring:
@pimjansen sure. Will remove the new ones, and move them into another PR targeting for 2.0 then.

pimjansen (approved):

pimjansen:
Thanks! Ready for review

fzaninotto (changes requested):
This is a breaking change and can only land in 2.0. As for the new names, I'm not sure they are clearer than the previous ones... `dateTimePastYear` can be understood as a date between Jan 1st and Dec 31st last year, or as a date in the last 365 days. Also, I don't think it's Faker's job to generate all kinds of specialized date methods. `dateTimeBetween`, used in conjunction with a good Date library, should be enough.

icanhazstring:
@fzaninotto i don't see were this is a breaking change?
I don't remove old methods, I simply mark the as deprecated an added a new one instead. So the old are just a proxy to the new ones.

fzaninotto:
You're right, it's not a breaking change. But I don't agree with the change anyway, for reasons I already explained above.

icanhazstring:
If so we could close this PR and all related issues with the dateTime methods.

#1165 
#1626 
#599 

As these have a use case where they have another understanding of the current methods, or a usecase where they want a new method to be implemented. As you already mentioned that Faker should not simply provide all the methods it self.

Would it be useful (as I already mentioned) to provide a seperate faker-package (targeting 2.0?) to provide such methods? This way the faker core will stay small and not bloated with methods that might have a small usage on the whole user base.

PR #1788, opened 2019-09-10
Add mobileNumber() method for nb_NO

carestad:
Added `$faker->mobileNumber` for Norwegian mobile phone numbers.

Included tests for the method, and added docs to the readme.md about this. Tests are currently passing.


localheinz:
Thank you, @carestad and @pimjansen!

PR #1789, opened 2019-09-16
typo

enumag:


src/Faker/Provider/Uuid.php
```diff
@@ -10,7 +10,7 @@ class Uuid extends Base
      */
     public static function uuid()
     {
-        // fix for compatibility with 32bit architecture; seed range restricted to 62bit
+        // fix for compatibility with 32bit architecture; seed range restricted to 32bit
```
fzaninotto: I don't understand this "seed range restricted to" comment. But concatenating two 32 bits random numbers does sound like 64 bits of randomness, not 32.

enumag: As I understand it, it means that each `mt_rand` call is limited to 32 bits. Previously PHP_MAX_INT was used which produced 32 bit randomness on 32bit systems and 64 bit randomnes on 64bit systems. Two 32bit calls are used instead to produce 64bit result.

fzaninotto: Can you please update the comment to make it more clear?

enumag:
```suggestion
        // fix for compatibility with 32bit architecture; each mt_rand call is restricted to 32bit
        // two such calls will cause 64bits of randomness regardless of architecture
```

fzaninotto:
Thanks!

PR #1790, opened 2019-09-17
Floor is not always returning an int

pimjansen:
Floor is not always returning an int:
https://www.php.net/floor


localheinz:

PR #1791, opened 2019-09-17
Facebook doesn't allow names with 3 or more consecutive identical characters

ktzouno:
Duplicate of https://github.com/fzaninotto/Faker/pull/1433 , because in the meantime I had deleted my fork.

### DIFF
--- a/test/Faker/Provider/PersonTest.php
+++ b/test/Faker/Provider/PersonTest.php
@@ -2,9 +2,11 @@
 
 namespace Faker\Test\Provider;
 
-use Faker\Provider\Person;
 use Faker\Generator;
+use Faker\Provider\Person;
 use PHPUnit\Framework\TestCase;
+use ReflectionException;
+use ReflectionProperty;
 
 class PersonTest extends TestCase
 {
@@ -83,5 +85,43 @@ class PersonTest extends TestCase
         $this->assertContains($faker->name('foobar'), array('John Doe', 'Jane Doe'));
         $this->assertContains($faker->name('male'), array('John Doe'));
         $this->assertContains($faker->name('female'), array('Jane Doe'));
+    }
+
+    public function localeProviderForPerson()
+    {
+        $locales = array();
+        foreach (glob(__DIR__ . '/../../../src/Faker/Provider/*/Person.php') as $localizedPerson) {
+            preg_match('#/([a-zA-Z_]+)/Person\.php#', $localizedPerson, $matches);
+            $locales[] = array($matches[1]);
+        }
+        return $locales;
+    }
+    /**
+     * @dataProvider localeProviderForPerson
+     *
+     * @param string $locale
+     * @throws ReflectionException
+     */
+    public function testFacebookCompliance($locale)
+    {
+        $personClassName = sprintf('Faker\Provider\%s\Person', $locale);
+        foreach (array('firstNameMale', 'firstNameFemale', 'lastName') as $propertyName) {
+            $propertyReflection = new ReflectionProperty($personClassName, $propertyName);
+            $propertyReflection->setAccessible(true);
+            /** @var string[] $names */
+            $names = $propertyReflection->getValue();
+            foreach ($names as $name) {
+                $this->assertNotRegExp(
+                    '/(.)\\1{2,}/',
+                    $name,
+                    sprintf(
+                        'Property %s::$%s contains name with 3+ consecutive identical characters: "%s"',
+                        $personClassName,
+                        $propertyName,
+                        $name
+                    )
+                );
+            }
+        }
     }
 }


test/Faker/Provider/PersonTest.php
```diff
@@ -84,4 +86,42 @@ public function testNameReturnsFirstNameAndLastName()
         $this->assertContains($faker->name('male'), array('John Doe'));
         $this->assertContains($faker->name('female'), array('Jane Doe'));
     }
+
+    public function localeProviderForPerson()
+    {
+        $locales = array();
+        foreach (glob(__DIR__ . '/../../../src/Faker/Provider/*/Person.php') as $localizedPerson) {
```
pimjansen: Im not happy with these kind of tests to parse php files. @localheinz what do you think?

pimjansen: @ktzouno could you update those tests and create something that does not scan the actual code but just consumes it?

pimjansen:
@ktzouno i added this PR to our 1.9.0 milestone. Hope we can make it on time to resolve the open topics.

Thanks

test/Faker/Provider/PersonTest.php
ktzouno: @pimjansen I see similar implementations using glob() in `InternetTest::localeDataProvider()`, `PaymentTest::localeDataProvider()` and `ProviderOverrideTest::localeDataProvider()`. Is your point still about the phpunit-provider or the test itself?

pimjansen: Yeah but its not really nice to do it. So please update accordingly. Would be great if we can make it asap. Else we have to remove it from the 1.9 release

ktzouno: Is it better if I get the list of locales from `Faker\Provider\Miscellaneous::$localeData`? Of course not all of them will have a `Person` provider.

pimjansen: Well what i would so is remove the name `Facebook` from the code since it has nothing to do with Facebook in general, it is just an use case. Further throw an exception when a name with 3 consecutive characters pops up. In the test just feed the method where you expect if the exception is thrown once it happens.

ktzouno: > Well what i would so is remove the name Facebook from the code since it has nothing to do with Facebook in general, it is just an use case.

Sounds fine.

> In the test just feed the method where you expect if the exception is thrown once it happens.

I don't get this part. (What I'd like to test is that no `firstNameMale`, `firstNameFemale`, `lastName` entries in all locales' `Person` Providers contain 3+ consecutive-identical characters).

test/Faker/Provider/PersonTest.php
```diff
@@ -2,9 +2,11 @@
 
 namespace Faker\Test\Provider;
 
-use Faker\Provider\Person;
 use Faker\Generator;
+use Faker\Provider\Person;
 use PHPUnit\Framework\TestCase;
+use ReflectionException;
+use ReflectionProperty;
```
localheinz: Personally, I'd prefer not to import classes from the root namespace, i.e, from the Standard PHP Library (see https://www.php.net/manual/en/book.spl.php).

Perhaps we can reference them relative to the root namespace instead?

test/Faker/Provider/PersonTest.php
localheinz: @ktzouno @pimjansen 

Perhaps something like this makes sense instead?

```php
<?php

namespace Faker\Test\Provider;

use Faker\Generator;
use Faker\Provider\Person;
use PHPUnit\Framework\TestCase;

abstract class AbstractPersonTestCase extends TestCase
{
    /**
     * @dataProvider facebookCompliantPropertyNameProvider
     *
     * @param string $propertyName
     */
    final public function testMagicPropertyProvidesFacebookCompliantName($propertyName)
    {
        $provider = static::fakerWithAddedPersonProvider();

        self::assertValueProvidedByMagicPropertyIsFacebookCompliantName($propertyName, $provider->{$propertyName});

    }

    final public function facebookCompliantPropertyNameProvider()
    {
        return array(
            'firstNameMale' => array('firstNameMale'),
            'firstNameFemale' => array('firstNameFemale'),
            'lastName' => array('lastName'),
        );
    }

    /**
     * @param string $propertyName
     * @param string $value
     */
    final protected static function assertValueProvidedByMagicPropertyIsFacebookCompliantName($propertyName, $value)
    {
        self::assertInternalType('string', $propertyName);
        self::assertInternalType('string', $value);

        self::assertNotRegExp('/(.)\\1{2,}/', $propertyName, sprintf(
            'Failed asserting that the "%s" provider returned a name that is compliant with Facebook, as "%s" has more than 3 consecutive identical characters.',
            $propertyName,
            $value
        ));
    }

    /**
     * @throws \RuntimeException
     *
     * @return Generator
     */
    final protected static function fakerWithAddedPersonProvider()
    {
        $className = self::personProviderClassName();

        if (!class_exists($className)) {
            throw new \RuntimeException(sprintf(
                'Class "%s", resolved from test class name "%s", does not exist',
                $className,
                get_called_class()
            ));
        }

        $faker = new Generator();

        $provider = new $className($faker);

        if (!$provider instanceof Person) {
            throw new \RuntimeException(sprintf(
                'Class "%s" does not extend "Faker\\Provider\\Person"',
                $className
            ));
        }

        $faker->addProvider($provider);

        return $faker;
    }

    /**
     * @return string
     */
    private static function personProviderClassName()
    {
        return preg_replace(
            '/Test$/',
            '',
            str_replace(
                'Faker\\Test\\Provider\\',
                'Faker\\Provider\\',
                get_called_class()
            )
        );
    }
}
```

The individual `PersonTest` classes could then extend from this `AbstractTestCase`.

A bit less magic, perhaps?

PR #1792, opened 2019-09-17
Improve en_US area code generation

shawnlindstrom:
The existing PhoneNumber provider does not fully implement the [NANP rules for area codes](https://www.nationalnanpa.com/area_codes/index.html). This can be problematic when running generated phone numbers against validation.

In concert with implementing the full NANP rules, this PR adds a new formatter: randomDigitNotIn(). The formatter is used here because the first and second digits of the area code cannot be 37 or 96. Since the formatter accepts an integer or array, it's convenient to pass [3, 7] or [9, 6] to get a digit exclusive of the passed digits.

This is a resubmission of https://github.com/fzaninotto/Faker/pull/1473 which I closed asmy rebase did not not go smoothly. Sorry.

### DIFF
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -63,6 +63,24 @@ class Base
     }
 
     /**
+     * Generates a random digit, which cannot be in $except
+     *
+     * @param array|int $except
+     * @return int
+     */
+    public static function randomDigitNotIn($except)
+    {
+        $digits = array(0, 1, 2, 3, 4, 5, 6, 7, 8, 9);
+        $pool = array_diff($digits, (array) $except);
+        if (!$pool) {
+            throw new \InvalidArgumentException('The argument is inclusive of the digits 0 to 9. A digit cannot be returned.');
+        }
+        $key = array_rand($pool);
+
+        return $pool[$key];
+    }
+
+    /**
      * Returns a random integer with 0 to $nbDigits digits.
      *
      * The maximum value returned is mt_getrandmax()
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -75,7 +75,7 @@ class Base
         if (!$pool) {
             throw new \InvalidArgumentException('The argument is inclusive of the digits 0 to 9. A digit cannot be returned.');
         }
-        $key = array_rand($pool);
+        $key = self::randomElement($pool);
 
         return $pool[$key];
     }
--- a/readme.md
+++ b/readme.md
@@ -119,6 +119,7 @@ Each of the generator properties (like `name`, `address`, and `lorem`) are calle
 
     randomDigit             // 7
     randomDigitNot(5)       // 0, 1, 2, 3, 4, 6, 7, 8, or 9
+    randomDigitNotIn(array(0, 2, 4, 6, 8)) // 1, 3, 5, 7 or 9
     randomDigitNotNull      // 5
     randomNumber($nbDigits = NULL, $strict = false) // 79907610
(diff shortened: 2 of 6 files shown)


shawnlindstrom:
@pimjansen Here is my resubmission. Sorry again for not being able to make the original work. Let me know if it needs anything else.

pimjansen:
@localheinz @fzaninotto i think this could be useful in the `randomDigitNot(x)` method right? We can easiliy make it backwards compatible to accept an array as well as an int and validate it there?

Or do you guys prefer 2 different methods for it?

shawnlindstrom:
> @localheinz @fzaninotto i think this could be useful in the `randomDigitNot(x)` method right? We can easiliy make it backwards compatible to accept an array as well as an int and validate it there?
> 
> Or do you guys prefer 2 different methods for it?

Hi Pim,

Actually, that's the approach I took in the original PR and asked the same question: https://github.com/fzaninotto/Faker/pull/1473#pullrequestreview-131574892

And, that's why this PR includes randomDigitNotIn().

pimjansen:
@shawnlindstrom understand. Lets wait for the other 2 on their ideas. I do agree that a list could be very useful so. Now it is just on the implementation

src/Faker/Provider/Base.php
```diff
@@ -62,6 +62,24 @@ public static function randomDigitNot($except)
...
+     * @return int
+     */
+    public static function randomDigitNotIn($except)
+    {
+        $digits = array(0, 1, 2, 3, 4, 5, 6, 7, 8, 9);
+        $pool = array_diff($digits, (array) $except);
+        if (!$pool) {
+            throw new \InvalidArgumentException('The argument is inclusive of the digits 0 to 9. A digit cannot be returned.');
+        }
+        $key = array_rand($pool);
```
fzaninotto: prior to php 7.1, this does not use a Mersenne Twister randomizer. Use the `randomElement` formatter to choose an element in the array.

fzaninotto (changes requested):

src/Faker/Provider/Base.php
shawnlindstrom: Change made as requested. The test failure is unrelated:
```
Faker\Test\Provider\ImageTest::testDownloadWithDefaults
getimagesize(): Read error!
```

shawnlindstrom:
Rebased and tests are passing. Please note this fixes #1837 and that there is an alternate PR started for the same issue open at #1838.

pimjansen:
We already released 1.9 last week so new features wont be added to the 1.x stream. We are now working on ideas of 2.0 so once that is clarified we can see what to do with this

pimjansen:
Please see #1851

anobjectn:
Please advise: if someone wants the fix for this issue - what are the current options? Is there a patch available? Is this added to 1.9 or some other version, or do I have to wait till 2.0 for proper fake US phone numbers?

shawnlindstrom:
It's dead @anobjectn. It didn't make it into 1.9 and they froze features until they figure out what's going into 2.0. I tried hard. Sorry.

anobjectn:
> It's dead @anobjectn. It didn't make it into 1.9 and they froze features until they figure out what's going into 2.0. I tried hard. Sorry.

No need for apology sir! Info was appreciated.  Thank you.

I could use a variation of the current phone number code but selecting from a given set of known valid area codes (e.g. 212,718,518) with the rest being the same (sometimes +1 country code, with and without extensions, different formatting). I may have to roll up my sleeve but more likely I'll just live with it for now (so if anyone wants to suggest something like that, share or dm). 

Thanks!

PR #1794, opened 2019-09-25
Add Gravatar provider

ottaviano:
👋 , 
this PR adds Gravatar Image generator:
https://fr.gravatar.com/site/implement/images


pimjansen:
What is the use-case for this one? Since there is already an image provider in the repo? So i do not really see any added benefit of adding another one

ottaviano:
you are right, an image provider already exists, but Gravatar generator can provide randomly generated pixeled avatars:

image

The actual Image provider calls `http://lorempixel.com` API which crops a big image for the requested size, for example: 
download

I think it'll be nice to have `identicon` or `retro` avatar generator too
image image

pimjansen:
@CodeCooler well i dont think this needs to be in the faker core to be honest. Since we provide fake data and there is already a way to generate a fake image there is no need.

However feel free to create your custom provider and let us link to your repo instead.

CodeCooler:
> @CodeCooler well i dont think this needs to be in the faker core to be honest.

I get that, and I don't really have a dog in the fight, but for what it's worth, I do think this meets a broad need that is not currently being addressed by the core faker providers:

Most of the time, I just want to generate profile avatars when seeding users for enterprise applications, and I find that more often than not, the current data source returns images that distract clients into conversations about the appearance of models, etc (case in point attached):

Buyer profile 2019-09-26 14-56-03

Sure, we could use a generic placeholder image, but this results in every profile being indistinguishable at first glance.

Generic/abstract avatars such as those provided here could be extremely handy in this use case.

pimjansen:
@CodeCooler understand. Feel free to make a custom provider and link it to our repo that you can still use it with Faker itself and it is not part of the Faker core.

ottaviano:
done https://github.com/ottaviano/faker-gravatar 
@pimjansen, should I make a PR to add a link to my repo? thanks

pimjansen:
Yes please 👌🏻

PR #1795, opened 2019-09-27
Update out-of-date currency codes

Brindster:
This pull request replaces the following obsolete currency codes with their new equivalents:

MRO -> MRU
STD -> STN
VEF -> VES

Data taken from wikipedia https://en.wikipedia.org/wiki/ISO_4217#Historical_currency_codes


localheinz:
Thank you, @Brindster!

PR #1798, opened 2019-09-30
Add elgentos/masquerade to third-party library list

erikhansen:


localheinz:
Thank you, @erikhansen!

PR #1799, opened 2019-10-02
Remove last names from the male first names array (et_EE)

Mihkel100:


pimjansen:
Can you please provide a source or anything and a reason why you want these names to be removed?

Mihkel100:
The removed first names are last names. You can find these in the lastNames array. 

The list of popular last names: https://www.stat.ee/public/apps/nimed/pere/TOP

localheinz:
Thank you, @Mihkel100!

PR #1806, opened 2019-10-06
Add all Iran's provinces land lines numbers

kingofnull:
Add all of Iran's provinces land lines number formats separated by province name.


fzaninotto:
thanks!

PR #1808, opened 2019-10-09
Add dateTimeImmutable functions

mcfedr:
Basically for every `dateTime` function there is a `dateTimeImmutable` equivalent.

- Also everywhere an argument accepted `DateTime` it now accepts `DateTimeInterface` so that either DateTime or DateTimeImmutable work. (in the code I check `$x instanceof \DateTime || $x instanceof \DateTimeInterface` so its backwards compatible with 5.3)


pimjansen:
Please see #1851

PR #1812, opened 2019-10-14
[Feature] Add Mobile Carrier faker

djunehor:
I've added a formatter, MobileCarrier. It has default values which are US mobile carriers. I also added an example for en_NG which returns random Nigerian mobile carrier. Tests have also been included which passed.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -32,6 +32,7 @@ Faker requires PHP >= 5.3.3.
 	- [Miscellaneous](#fakerprovidermiscellaneous)
 	- [Biased](#fakerproviderbiased)
 	- [Html Lorem](#fakerproviderhtmllorem)
+	- [Mobile Carrier](#fakerprovidercarrier)
 - [Modifiers](#modifiers)
 - [Localization](#localization)
 - [Populating Entities Using an ORM or an ODM](#populating-entities-using-an-orm-or-an-odm)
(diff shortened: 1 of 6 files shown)


pimjansen:
We recently closed the 1.9 release which means this is the last minor for Faker 1.9. In version 2.0 we will remove the locales from the core so dont expect this to be merged any time soon. I would like to refer to the issue tracker where we have an active discussion on the new 2.0 release

djunehor:
has work commenced on the 2.0? I saw the issue tagged 2.0 where you stated that locales should be removed from core and I agree. The MobileCarrier formatter can be left in the core, so each locale can extend it. Or what do you think?

pimjansen:
@djunehor we are now having an open discussion on the scope and ideas so feel free to comment on the issue and share your ideas there. We are still discussion on how to attack this where a locale is standalone enough but also that the faker core can hold all stuff together. Could be with some interfacing and so on but feel free to share your thoughts.

About the mobile carrier, i think this is typically something we should do in a locale since every country has its own set of carriers. The core in this case has not much meaning except maybe the fact that a locale can or should hold a method like that.

djunehor:
I think the MobileCarrier provider be left as non locale-specific provider like PhoneNumber provider, so that locales specific MobileCarrier class  just need to extend and provide data as it works with the PhoneNumber provider.

For example
en_NG/MobileCarrier will contain something like `protected static $networks = ['MTN Nigeria', 'Airtel', 'Globacom'];` which the MobileCarrier class uses to generate random mobile carrier when ne_NG is defined as locale.

pimjansen:
Please see #1851

PR #1815, opened 2019-10-18
Update .travis.yml

andreybolonin:


pimjansen:
No need for this yet since there is no stable php 7.4?

PR #1816, opened 2019-10-19
PHP 7.4 compat

trasher:


erikgaal:
PHP 7.4 compatibility was already added in #1748.

trasher:
Sounds like something has been missed; and there were no travis test for 7.4.

erikgaal:
Oop, good catch. Assumed that `nightly` would be the testing on `7.4`. Could you please elaborate bit further the required and already made changes in the original post?

trasher:
I'm sorry, I do no understand. Current master fails on PHP 7.4 because of brackets syntax accessing arrays; that is what the current PR fix.
Also, nightly on Travis is PHP 8.0.0 (see travis outputs).

localheinz:
@trasher 

Closing, apologies for the delay.

Also see #1748 and #1861.

PR #1817, opened 2019-10-20
Adding ability to generate a random normally  distributed number

shutchinson:
This adds the ability to generate a random number that is normally distributed around a mean and a standard deviation.

It is not a perfect solution but should give decent results for most fake data use cases.


pimjansen:
Please see #1851

PR #1819, opened 2019-10-25
Fix: Remove unnecessary class-level DocBlocks

localheinz:
This PR

* [x] removes unnecessary class-level DocBlocks


localheinz:

PR #1820, opened 2019-10-25
Fix: Mark test classes as final

localheinz:
This PR

* [x] marks test classes as `final`

💁‍♂ There's probably no reason to extend from them.


pimjansen:
Good for me to with 1.9.1 indeed

fzaninotto:
Thanks!

localheinz:

PR #1821, opened 2019-10-25
Fix: Reduce visibility of setUp() and tearDown()

localheinz:
This PR

* [x] reduces the visibility of `setUp()` and `tearDown()` from `public` to `protected`

💁‍♂ This is the visibility as declared on `PHPUnit\Framework\TestCase`, and there's probably no need to increase it.


localheinz:

PR #1822, opened 2019-10-25
Enhancement: Configure verbose output via phpunit.xml.dist

localheinz:
This PR

* [x] configures verbose output via `phpunit.xml.dist


pimjansen:
This simplifies indeed

fzaninotto:
Thanks!

localheinz:

PR #1827, opened 2019-10-25
Fix: Work around flaky test

localheinz:
This PR

* [x] attempts to work around flaky tests related to loading an image file from an external API


pimjansen:
@localheinz think we can drop this right since we skipped the test in the 1.9.0 release in total

PR #1828, opened 2019-10-26
Adding pl_PL language specific formatters to readme.

piotrgradzinski:
I've updated few language specific formatters in readme. It will be easier to find them without checking the source code.


pimjansen:
Please see #1851

PR #1831, opened 2019-10-31
Show supported version of UUID (version 3)

haakym:
Update readme to show supported version of UUID (version 3).

This got me recently where I assumed it was version 4 and I couldn't get the Faker generated UUID to validate. I wanted to ensure no-one else made the wrong assumption I made too by having it clearly in the readme!


pimjansen:
Please see #1851

PR #1832, opened 2019-11-04
En_GB local payment provider new methods

DWMcload:
Adding UK Bank Account and Sort Code generators to the local Payment Provider with tests.


pimjansen:
Please see #1851

PR #1838, opened 2019-11-07
Fix U.S. Area Code Generation (Fixes #1837)

### LINKED ISSUE #1837: Generates invalid US phone numbers
### Summary

When using the en_US locale, the phoneNumber method generates invalid numbers.  This makes validating Faker's output using e.g. Google's libphonenumber impossible.

This has the same ramifications as #1349, though not specifically for e164 format.

### Versions

|                     | Version |
|:--------------------|:--------|
| PHP                 | 7.3.11  |
| `fzaninotto/faker`  | 1.8.0   |

For example:

```php
$faker->phoneNumber === '(969) 476-8825 x74390';
```

This number is invalid, and trying to validate it as a US or CA phone number fails (can't start with 96).  The logic used to generate area codes in `\Faker\Provider\en_US\PhoneNumber::areaCode` is just too simplistic.  Here is the regex used by google's libphonenumber:

```regex
/(?:2(?:0[1-35-9]|1[02-9]|2[03-589]|3[149]|4[08]|5[1-46]|6[0279]|7[0269]|8[13])|3(?:0[1-57-9]|1[02-9]|2[0135]|3[0-24679]|4[167]|5[12]|6[014]|8[056])|4(?:0[124-9]|1[02-579]|2[3-5]|3[0245]|4[0235]|58|6[39]|7[0589]|8[04])|5(?:0[1-57-9]|1[0235-8]|20|3[0149]|4[01]|5[19]|6[1-47]|7[013-5]|8[056])|6(?:0[1-35-9]|1[024-9]|2[03689]|[34][016]|5[017]|6[0-279]|78|8[0-29])|7(?:0[1-46-8]|1[2-9]|2[04-7]|3[1247]|4[037]|5[47]|6[02359]|7[02-59]|8[156])|8(?:0[1-68]|1[02-8]|2[08]|3[0-28]|4[3578]|5[046-9]|6[02-5]|7[028])|9(?:0[1346-9]|1[02-9]|2[0589]|3[0146-8]|4[0179]|5[12469]|7[0-389]|8[04-69]))[2-9]\d{6}/
```

Unfortunately this is too complicated for regexify, but it could be converted into a PHP function easily enough.  This isn't likely to change often, but potentially more frequently than the current solution.

hackel:
Ensures phone numbers generated are valid.

Regular expressions extracted from the PHP port of Google's [libphonenumber](https://github.com/google/libphonenumber) via the PHP port:
https://github.com/giggsey/libphonenumber-for-php/blob/f3a55dcb6fead9d4fb291c6dadda0564e5ae8d3f/src/data/PhoneNumberMetadata_US.php#L29

### DIFF
--- a/src/Faker/Provider/en_US/PhoneNumber.php
+++ b/src/Faker/Provider/en_US/PhoneNumber.php
@@ -4,6 +4,17 @@ namespace Faker\Provider\en_US;
 
 class PhoneNumber extends \Faker\Provider\PhoneNumber
 {
+    protected static $areaCodeRegexes = [
+        2 => '(0[1-35-9]|1[02-9]|2[03-589]|3[149]|4[08]|5[1-46]|6[0279]|7[0269]|8[13])',
+        3 => '(0[1-57-9]|1[02-9]|2[0135]|3[0-24679]|4[167]|5[12]|6[014]|8[056])',
+        4 => '(0[124-9]|1[02-579]|2[3-5]|3[0245]|4[0235]|58|6[39]|7[0589]|8[04])',
+        5 => '(0[1-57-9]|1[0235-8]|20|3[0149]|4[01]|5[19]|6[1-47]|7[013-5]|8[056])',
+        6 => '(0[1-35-9]|1[024-9]|2[03689]|[34][016]|5[017]|6[0-279]|78|8[0-29])',
+        7 => '(0[1-46-8]|1[2-9]|2[04-7]|3[1247]|4[037]|5[47]|6[02359]|7[02-59]|8[156])',
+        8 => '(0[1-68]|1[02-8]|2[08]|3[0-28]|4[3578]|5[046-9]|6[02-5]|7[028])',
+        9 => '(0[1346-9]|1[02-9]|2[0589]|3[0146-8]|4[0179]|5[12469]|7[0-389]|8[04-69])',
+    ];
+
     /**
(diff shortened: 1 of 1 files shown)


hackel:
It doesn't look like the ImageTest is failing because of me: `getimagesize(): Read error!`

pimjansen:
> It doesn't look like the ImageTest is failing because of me: `getimagesize(): Read error!`

Correct, fix for this is already pushed in a PR for the 1.9.0 release.

This fix will solve #1837

pimjansen:
Could we maybe add some tests?

hackel:
Yeah, I wasn't sure on the best approach for testing this beyond what is already there.  I could check the output against the full regex, or I could pull in libphonenumber-php as a dev dependency and check it against that.

Actually that way we could easily test all of the different locales, though I suspect that could trigger even more failures.  Specifically, CA and other +1 countries should use similar logic to this.

pimjansen:
> Yeah, I wasn't sure on the best approach for testing this beyond what is already there. I could check the output against the full regex, or I could pull in libphonenumber-php as a dev dependency and check it against that.
> 
> Actually that way we could easily test all of the different locales, though I suspect that could trigger even more failures. Specifically, CA and other +1 countries should use similar logic to this.

Well we clearly have an issue with the 96 one which should not be there. So we can test that one right? Or even all the formats that can match

shawnlindstrom:
Please see #1792 which I submitted a while ago. It fixes #1837. I just rebased and tests are passing.

PR #1839, opened 2019-11-07
fr_FR e164PhoneNumber provider

Geekimo:
Updated fr_FR phoneNumber provider to make it compliant with fr_FR phone number scheme


pimjansen:
Please see #1851

PR #1844, opened 2019-11-12
Add named unique() "bags"

haggys22:
I want to seed two table columns independently from each other but with the same set of possible values, e.g. `[1, 2, 3, 4, 5, 6, 7, 8]`

My Laravel factory and seeder look like this:
```
// Factory
$factory->define(MyModel::class, function (Faker $faker) {
    return [
        'a' => $faker->unique()->randomElement([1, 2, 3, 4, 5, 6, 7, 8]),
        'b' => $faker->unique()->randomElement([1, 2, 3, 4, 5, 6, 7, 8])
    ];
});
// Seeder
public function run() {
    factory(MyModel::class, 4)->create();
}
```

Currently, the UniqueGenerator will save the existing values only by function name, e.g. `"randomElement"`. Therefore I can only seed up to 4 models and receive no `a` value as a `b` value.

**Solution:**
With this pull request I propose adding a `$bag` parameter to the `unique()` modifier to define independent sets of existing values. The given example factory could then be changed to

```
$factory->define(MyModel::class, function (Faker $faker) {
    return [
        'a' => $faker->unique('a')->randomElement([1, 2, 3, 4, 5, 6, 7, 8]),
        'b' => $faker->unique('b')->randomElement([1, 2, 3, 4, 5, 6, 7, 8])
    ];
});
```

This allows generation of up to 8 models with distinct values per column (`a` and `b`) but same values using the same formatter for different columns.

When used without the `$bag` parameter the behavior will not change.

### DIFF
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -15,9 +15,9 @@ class Base
     protected $generator;
 
     /**
-     * @var \Faker\UniqueGenerator
+     * @var \Faker\UniqueGenerator[]
      */
-    protected $unique;
+    protected $unique = [];
 
     /**
      * @param \Faker\Generator $generator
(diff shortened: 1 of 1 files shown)


pimjansen:
Please see #1851

PR #1847, opened 2019-11-14
Enhancement: Switch to using GitHub actions

localheinz:
This PR

* [x] switches to using GitHub actions


.travis.yml
```diff
@@ -1,24 +0,0 @@
-language: php
-
-dist: precise
-
-php:
-    - 7.2
-    - 7.3
-    - nightly
```
localheinz: Can't run on `nightly` - do we care?

pimjansen: No really if you ask me. Isnt there a 7.4 image available already since release is in 2 weeks from now.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,76 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v1.1.0
+
+      - name: "Install PHP with extensions"
+        uses: shivammathur/setup-php@1.5.1
+        with:
+          php-version: 7.1
```
pimjansen: Are we running phpcs only for 7.1 at this point? No idea if we could actually have a diff there based on a PHPversion. Dont really think so though

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,76 @@
+# https://help.github.com/en/categories/automating-your-workflow-with-github-actions
```
pimjansen: Let me just add a comment on rootlevel. Isnt it maybe wise to integrate all tooling already that we want to have implemented? Like PHPstan and maybe some comp. checks or will we add them later

localheinz: @pimjansen 

I'd prefer to add them later, one by one.

.github/workflows/continuous-integration.yml
localheinz: @pimjansen 

I think it suffices to run a coding standards fixer on the lowest supported version of PHP.

localheinz:
@fzaninotto 

Looks like Actions aren't enabled - at least they are not running and there are no complaints about invalid YML configuration either.

Can you double-check that GitHub Actions are enable?

Screen Shot 2019-11-23 at 12 12 33

Could also be that they are not activated because I do not have push privileges.

fzaninotto:
Yes, actions are enabled... Dunno what's wrong.

pimjansen:
I see that there is now something within actions, so maybe if we rebase this it should be fine i guess

localheinz:
@pimjansen 

Will do!

localheinz:
@fzaninotto @pimjansen 

Looking good!

.travis.yml
localheinz: Is available now, yes!

localheinz:
I need to collect code coverage here as well, give me a moment!

.github/workflows/continuous-integration.yml
```diff
@@ -12,11 +12,106 @@ on:
...
+
+      - name: "Create build directory"
+        run: mkdir -p build/logs
+
+      - name: "Collect code coverage with Xdebug and phpunit/phpunit"
+        run: vendor/bin/phpunit --coverage-clover=build/logs/clover.xml
+
+      - name: "Send code coverage report to Codecov.io"
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
localheinz: This token needs to be added via https://github.com/fzaninotto/faker/settings/secrets.

Also take note of

- https://github.com/codecov/codecov-bash/issues/224
- https://community.codecov.io/t/whitelist-github-action-servers-to-upload-without-a-token/491/8

fzaninotto:
Please avoid force pushing. The merger can squash the commits when merging the PR.

localheinz:
@fzaninotto @pimjansen 

I think this is good for another review!

localheinz:
@fzaninotto @pimjansen 

Please make sure to squash the commits on merge!

localheinz:

PR #1849, opened 2019-11-18
Update the list of Kazakhstan regions and cities.

crayxt:
As per this wikipedia page 
https://kk.wikipedia.org/wiki/%D2%9A%D0%B0%D0%B7%D0%B0%D2%9B%D1%81%D1%82%D0%B0%D0%BD_%D3%99%D0%BA%D1%96%D0%BC%D1%88%D1%96%D0%BB%D1%96%D0%BA_%D0%B1%D1%96%D1%80%D0%BB%D1%96%D0%BA%D1%82%D0%B5%D1%80%D1%96


pimjansen:
Please see #1851

crayxt:
@pimjansen I see. I would categorize this PR as bugfix then, since I fixed the outdated information.

pimjansen:
@crayxt i dont agree here since we are actually just changing a recordset. We generate dummy data meaning that outdated data is not directly wrong. It is maybe not for your needs though.

So i would like to ask to please join the discussion Faker 2 so we can get it going asap and also have a full new Kazakhstan locale at some point

PR #1843, opened 2019-11-12
Curly braces for arrays is deprecated in PHP 7.4

wimg:
Curly braces for arrays is deprecated in PHP 7.4 (detected by PHPCompatibility)

PR #1850, opened 2019-11-18
fix phplint warnings (7.4)

remicollet:
Array and string offset access syntax with curly braces is deprecated


remicollet:
Duplicate of #1843

Issue #1851, opened 2019-11-21
Enhancement freeze


pimjansen:
Hey all,

Since the release of `Faker 1.9.0` last week the team decided that we will freeze the minor versions for the 1.x stream. This means that no new functionality will be added for 1.9 and that there won't be a 1.10 in the future. For the 1.x stream only bug fixes will still be applied.

You may be wondering what to do next with my great ideas for `Faker`. We are currently in an open discussion on how the new structure of `Faker 2.0` should look like #1807. One of the topics where we are absolute sure about is that all the locales with be remove from the core.

So if you want to make your own locale provider in the future or anything like that, please hold your horses for now, join the discussion so we can get Faker 2.0 up and running as fast as we can

chris-doehring:
Hey @pimjansen,

I use this project frequently as a helper for testing scenarios and it helped me a lot. I understand the intention of moving forward to an improved new major version of Faker, but seeing the current date, the fixed date of the release of php 8 (26th of November) and the importance this project has in the php developer community, I would kindly ask you and the team to make an exception and release another 1.x minor version of Faker. The current php version constraint does not allow a usage under php 8, but I’m sure many projects are already in preparation of trying to migrate to php 8 or will definitely do it after the 26th of November.

I don’t think that it’s realistic to say, that Faker 2.x is close to a release or that it will be released before php 8, seeing the current state of progress on it. Would you and other team members mind considering my proposal in doing an exception and join the ongoing discussion on my pull request #2063?
The alternative would be, that a new official fork of this project will be created which will be maintained separately for years to come, but I don’t think that this in your and the communities’ interest.

PR #1854, opened 2019-11-23
Modify th_TH ssn() to conform to Thail national id

theminer3746:
Modify th_TH ssn() to conform to Thail national id and have correct checksum.


theminer3746:
Saw #1851

PR #1855, opened 2019-11-23
Add ar_EG locale support

AhmedRaafat14:
##### Here am adding support for the Arabic language that is used in Egypt.

* Tested these changes and it works as expected.
* Willing to add more tests in the future if possible.


pimjansen:
See #1851

PR #1858, opened 2019-11-26
Enhancement: Echo a greeting

localheinz:
This PR

* [x] echos a greeting using GitHub actions

Related to #1847.

💁‍♂ Let's see if this even works.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,0 +1,22 @@
+# https://help.github.com/en/categories/automating-your-workflow-with-github-actions
+
+on:
+  pull_request:
+  push:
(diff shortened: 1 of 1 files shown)


localheinz:
@fzaninotto 

Something is still off - we could expect a workflow to run here, but nothing happens.

fzaninotto:
I know, but the Actions settings seem right...

image

localheinz:
Hmm, weird.

Perhaps we just merge this in, as a test?

We can still remove it afterwards in case this doesn't work at all - less work than merging in #1847 and reverting afterwards.

pimjansen:
Lets indeed merge this as a test

PR #1861, opened 2019-12-01
Fix: Run build against PHP 7.4

localheinz:
This PR

* [x] runs builds against PHP 7.4 instead of nightly

Related to #1847.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ jobs:
     - php: 7.1
     - php: 7.2
     - php: 7.3
+    - php: 7.4
+      dist: bionic
       env: WITH_COVERAGE=true
     - php: nightly
 


.travis.yml
```diff
@@ -12,8 +12,8 @@ jobs:
     - php: 7.1
     - php: 7.2
     - php: 7.3
+    - php: 7.4
       env: WITH_COVERAGE=true
```
localheinz: I guess it's better to collect coverage on PHP 7.4 instead of PHP 7.3 - it's probably faster, too.

.travis.yml
```diff
@@ -12,28 +12,17 @@ jobs:
     - php: 7.1
     - php: 7.2
     - php: 7.3
```
pimjansen: Shouldnt we add 7.4 here now instead of nightly?

localheinz: Not available on Travis, that's why I added workflow for GitHub actions.

pimjansen: Is it time to wait before we do this or?

fzaninotto: I think we should merge GitHub actions first

pimjansen: Done

localheinz: 👍

.travis.yml
```diff
@@ -12,12 +12,8 @@ jobs:
...
+    - php: 7.4
       env: WITH_COVERAGE=true
-    - php: nightly
-
-matrix:
-    allow_failures:
-        - nightly
```
fzaninotto: I think we should still allow failures on nightly

localheinz: Adjusted!

localheinz:
@fzaninotto @pimjansen 

Build failed, let's see if it still does.

snapshotpl:
@localheinz you need to change to Xenial and Bionic for php 7.4

localheinz:
Thank you, @fzaninotto, @pimjansen, and @snapshotpl!

PR #1866, opened 2019-12-03
Fix: Aggregate badges in one place

localheinz:
This PR

* [x] aggregates build status and similar badges in a single place


pimjansen:
Thanks!

localheinz:

PR #1869, opened 2019-12-04
Generate valid ward numbers

sapkotamadan:
Ward number in Nepal start from 1 and goes maximum 33 (in new division) and maximum 36 (in old division) system.


remotemerge:
@pimjansen this one is pending much 🤔 can you make a review?

Issue #1876, opened 2019-12-18
German phone numbers invalid for libphonenumber


neeg:
### Summary

Faker generates german phone numbers like `00879 67680` that are not considered valid:
https://libphonenumber.appspot.com/phonenumberparser?number=00879%2067680&country=DE 

### Versions

|                     | Version |
|:--------------------|:--------|
| PHP                 | 7.3.10   |
| `fzaninotto/faker`  | 1.9.1   |

### Self-enclosed code snippet for reproduction

```php
$faker->phoneNumber()
```

### Expected output

German phone numbers never start with 00.
Only 0[1-9]xxx are possible

### Actual output

00879 67680

pimjansen:
@localheinz could you verify this?

minthemiddle:
@pimjansen Here is an official confirmation that leading zeros are not allowed:
https://www.bundesnetzagentur.de/DE/Sachgebiete/Telekommunikation/Unternehmen_Institutionen/Nummerierung/Rufnummern/ONRufnr/ONL%C3%A4nge_ONRufnr/L%C3%A4nge_ONRufnr_Basepage.html

PR #1877, opened 2019-12-18
Generate valid german phone numbers

neeg:
fix for #1876


pimjansen:
Could you please add a valid test for this scenario too?

PR #1889, opened 2020-01-13
Fix: Reduce visibility

localheinz:
This PR

* [x] reduces the visibility of `setUp()` and `tearDown()` from `public` to `protected`

💁‍♂ I ran

```
$ composer global require friendsofphp/php-cs-fixer
```

followed by

```
$ php-cs-fixer fix --allow-risky=yes --rules=php_unit_set_up_tear_down_visibility test
```

### DIFF
--- a/test/Faker/Provider/AddressTest.php
+++ b/test/Faker/Provider/AddressTest.php
@@ -10,7 +10,7 @@ class AddressTest extends TestCase
 {
     private $faker;
 
-    public function setUp(): void
+    protected function setUp(): void
     {
         $faker = new Generator();
         $faker->addProvider(new Address($faker));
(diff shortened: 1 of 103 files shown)


pimjansen:
Is there a reason why those are not private @localheinz ?

localheinz:
@pimjansen 

Yes, `protected` is the visibility declared in the parent class, `PHPUnit\Framework\TestCase`.

pimjansen:
@localheinz can we rebase this? Think we already updated it with your previous PR as well

localheinz:
@pimjansen 

Rebased!

pimjansen:
Thanks @localheinz

PR #1890, opened 2020-01-13
Enhancement: Require phpunit/phpunit:^8.5.2

localheinz:
This PR

* [x] requires `phpunit/phpunit:^8.5.2`
* [x] migrates tests
* [x] fixes risky tests

💁‍♂ Since we require PHP 7.2, we can also use `phpunit/phpunit:^8.5.2`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
-vendor
+.build/
+vendor/
 composer.lock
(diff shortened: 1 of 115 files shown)


pimjansen:
Looking good!

localheinz:
Thank you, @pimjansen!

PR #1891, opened 2020-01-13
Fix: Consistently quote strings in Yaml files

localheinz:
This PR

* [x] consistently quotes strings in Yaml files


pimjansen:
In YAML all is a string on default? I dont really see a valid case for doing all this tbh. Except for all the weird url encoded parts and such

PR #1892, opened 2020-01-13
Fix: Invoke composer directly

localheinz:
This PR

* [x] invokes `composer` directly


pimjansen:
nice one!

localheinz:
Thank you, @pimjansen!

PR #1893, opened 2020-01-13
Fix: Configure preferred installation method via composer.json

localheinz:
This PR

* [x] configures the preferred installation method via `composer.json` instead of specifying the corresponding option in `Makefile`


pimjansen:
Less errors indeed, nice catch

localheinz:
Thank you, @pimjansen!

PR #1895, opened 2020-01-13
Enhancement: Update squizlabs/php_codesniffer

localheinz:
This PR

* [x] updates `squizlabs/php_codesniffer`
* [x] runs `make fix`

💁‍♂ For reference, see https://github.com/squizlabs/PHP_CodeSniffer/compare/2.9.2...3.5.4


localheinz:
Thank you, @pimjansen!

PR #1896, opened 2020-01-13
Enhancement: Use XML configuration for squizlabs/php_codesniffer

localheinz:
This PR

* [x] uses an XML configuration file for `squizlabs/php_codesniffer`


pimjansen:
Thanks @localheinz

PR #1899, opened 2020-01-16
Change Image generation service to use https://placeholder.com/

### LINKED ISSUE #1884: Lorempixel is down : impossible to use $faker->image (again)
### Summary

LoremPixel.com is used by $faker->image to download files from the service.

The service is down, so we can't use $faker->image() at the moment.

We use it in the seeds of a Laravel application, which disables our preproduction environment.

A timeout detection MAY be enough for us, as it would allow the preproduction environment to run, even without images.

jonathanbossenger:
Fixes #1884 

In recent tests, even though lorempixel service is live, the faker image creation functionality takes over 120 seconds to create and return the image. Changing the service to use placeholder.com reduces that time to less than two seconds.

This PR changes the Image Provider to use the placeholder.com service. This is the most basic of implementations and unfortunately this means the loss of categories, but it will successfully create faker images more quickly than lorempixel

### DIFF
--- a/src/Faker/Generator.php
+++ b/src/Faker/Generator.php
@@ -181,8 +181,8 @@ namespace Faker;
  * @property string $fileExtension
  * @method string file($sourceDirectory = '/tmp', $targetDirectory = '/tmp', $fullPath = true)
  *
- * @method string imageUrl($width = 640, $height = 480, $category = null, $randomize = true, $word = null, $gray = false)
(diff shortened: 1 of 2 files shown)


jonathanbossenger:
as per @pimjansen this functionality is going to be removed from core in the 2.x release, so closing this PR

hasanalibalcioglu:
It is down again.

PR #1904, opened 2020-01-22
added tax support for de_DE

bjoernffm:
I added two new functions for company tax ids (VAT) and the local tax id for persons (TIN).


bjoernffm:
@fzaninotto The tests seem to run smoothly but composer seems to have a problem with the php-intl extension

pimjansen:
See #1851

PR #1906, opened 2020-01-23
Remove unsuitable job titles

jaapromijn:
Pooier means _pimp_, Prostitué and Prostituee mean _prostitute_ in Dutch.
While they are true job titles I wonder if they are suitable for faker data.

I guess you can merge this at your own discretion.


pimjansen:
Haha goed gespot!

I will mark this as enhancement though since it could break certain implementations. I will keep it in mind for the NL locale for 2.0

PR #1908, opened 2020-01-25
Added script to run tests using local PhpUnit

juliolmuller:
Just added a *Composer* script to the project  to run tests using project's PhpUnit dependency . A nice turnarround for those who have incompatible PhpUnit versions installed globally.

Just run `composer run test` in terminal.


pimjansen:
We use make for this, see the Makefile

PR #1911, opened 2020-02-02
Removed legacy autoloader

pimjansen:
I remove the PSR0 autoloader. Since we want to drop all the legacy support i don't really see any usecase for this anymore


localheinz:
Thank you, @pimjansen!

Issue #1913, opened 2020-02-07
Durable Determistic Faker Using seed()


kfriars:
This is less of a bug report, and more of a really useful feature request, so forgive me if I am putting this in the wrong place.

My issue is as follows:

1) I want deterministic output from faker
2) I have essential 3rd party dependencies
3) I want to be able to use seed on faker once, and ensure determinism whenever I use that instance

To give a little more context, I am using laravel and binding a Faker singleton to the service container, and setting seed on it. Through all of my many seeder classes I want the output to be consistent on every run.

Something is ruining the detemrinism of Faker even though seed is set, and it is impossible to debug mt_srand, so I have come up with a solution and will post a PR here in soon.

My solution is to provide a durable mode, where calling seed sets a private property on the Generator class called $seedNo. By adding a second argument to seed($seed =null, $durable = false) we can enforce a durable mode. Every time __call and __get are invoked it checks if durable is set and will call $this->seed($this->seedNo+1). This solutions maintains the "randomness" while also guaranteeing determinism regardless of other dependencies.

This solution also maintains backward compaitibility since it the default value of $durable enforces the old behavior.

Please let me know your thoughts I will fork and submit a PR and post it here. Thanks,

Kurt

kfriars:
Ok here it is PR#1914

https://github.com/fzaninotto/Faker/pull/1914

kfriars:
Found a better solution that doesn't require a PR.

thinkstylestudio:
What was that better solution.

PR #1914, opened 2020-02-07
durable-deterministic-mode Added durable mode to seed()

kfriars:
See https://github.com/fzaninotto/Faker/issues/1913


pimjansen:
As mentioned in the issue, better solution without code change found

thinkstylestudio:
What was that the better solution?

PR #1918, opened 2020-02-14
Remove unneeded countries

jamesfairhurst:
Local `en_GB` addresses shouldn't have other random countries.


pimjansen:
The provider can output a list of all possible countries and is as intended like this.

jamesfairhurst:
Ah ok I didn't know that, seems strange that you can generate a localized address with a country outside of the localization. How would you restrict it without hard-coding? Or is that outside the scope?

In my mind having the full list of countries in the base `Address` address and then limit per localization seems right.

pimjansen:
The method is there to generate random country names based on the selected locale.

If you want to have a single static one you already know what you want right?

jamesfairhurst:
Yeah true, think I was looking at https://github.com/fzaninotto/Faker/blob/master/src/Faker/Provider/en_HK/Address.php which restricts the countries so assumed it should work like that, thanks for getting back to me.

PR #1933, opened 2020-02-20
update french region and delete double data on street prefix

jbaptisteq:
- Update French Region with new administrative region
- Fix double term 'rue' in street prefix variable

### DIFF
--- a/src/Faker/Provider/fr_FR/Address.php
+++ b/src/Faker/Provider/fr_FR/Address.php
@@ -5,7 +5,7 @@ namespace Faker\Provider\fr_FR;
 class Address extends \Faker\Provider\Address
 {
     protected static $citySuffix = array('Ville', 'Bourg', '-les-Bains', '-sur-Mer', '-la-Forêt', 'boeuf', 'nec', 'dan');
-    protected static $streetPrefix = array('rue', 'rue', 'chemin', 'avenue', 'boulevard', 'place', 'impasse');
+    protected static $streetPrefix = array( 'rue', 'chemin', 'avenue', 'boulevard', 'place', 'impasse');
     protected static $cityFormats = array(
         '{{lastName}}',
         '{{lastName}}',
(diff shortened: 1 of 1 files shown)


src/Faker/Provider/fr_FR/Address.php
```diff
@@ -44,12 +44,11 @@ class Address extends \Faker\Provider\Address
     );
 
     private static $regions = array(
-        'Alsace', 'Aquitaine', 'Auvergne', 'Bourgogne', 'Bretagne', 'Centre', 'Champagne-Ardenne',
-        'Corse', 'Franche-Comté', 'Île-de-France', 'Languedoc-Roussillon', 'Limousin',
-        'Lorraine', 'Midi-Pyrénées', 'Nord-Pas-de-Calais', 'Basse-Normandie', 'Haute-Normandie',
-        'Pays-de-Loire', 'Picardie', 'Poitou-Charentes', "Provence-Alpes-Côte d'Azur", 'Rhone-Alpes',
-        'Guadeloupe', 'Martinique', 'Guyane', 'Réunion', 'Saint-Pierre-et-Miquelon', 'Mayotte',
-        'Saint-Barthélémy', 'Saint-Martin', 'Wallis-et-Futuna', 'Polynésie française', 'Nouvelle-Calédonie'
+        'Auvergne-Rhône-Alpes', 'Bourgogne-Franche-Comté', 'Bretagne', 'Centre-Val de Loire', 'Corse',
```
pimjansen: Why did we remove data here?

jbaptisteq: Hello,

This data are removed because they don't exist anymore. The 'region' administrative system has changed in 2016. I have replaced the old region by the current.

pimjansen: Ok however that means that i can't just merge it for the 1.9.2 release since it will be a breaking change.

If you look at the announcement we also have an enhancement freeze while development of 2.x is underway at this point which will be a big refactor.

jbaptisteq: Ok Thanks for the review.
I will wait for this new version to propose the modification again.

PR #1935, opened 2020-02-20
Fix: Add type and return type declarations to seed()

### LINKED ISSUE #1934: Document that seed argument gets casted to an int
### Summary

Seed method internally casts the value to an int.
I wasn't expecting it to be, so i used two strings without numbers as seed value.

### Versions

v1.9.1

### Self-enclosed code snippet for reproduction

```php
$faker = Factory::create();

$faker->seed('a');
$first = $faker->uuid;
$faker->seed('b');
$second = $faker->uuid;

assert($first !== $second); <- fails
```

### Solution

A simple phpdoc or a type hint would allow an IDE to warn the user about this.

localheinz:
This PR

* [x] adds type and return type declarations to `Faker\Generator::seed()`
* [x] cleans up underlying code

Fixes #1934.

### DIFF
--- a/src/Faker/Generator.php
+++ b/src/Faker/Generator.php
@@ -210,15 +210,15 @@ class Generator
         return $this->providers;
     }
 
-    public function seed($seed = null)
+    public function seed(int $seed = null): void
     {
         if ($seed === null) {
             mt_srand();
         } else {
             if (PHP_VERSION_ID < 70100) {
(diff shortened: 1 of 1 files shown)


localheinz:
Build failure is unrelated.

localheinz:
Closing in favor of #1894.

PR #1938, opened 2020-02-25
change https to http in Faker Image to better performance

antonio-aguirre:
was changed https to http in Faker Image to better performance. See the examples in lorempixel.com


pimjansen:
I don't see a reason why to do this. There is not that much performance overhead where the full web goes over ssl

antonio-aguirre:
yep, i did tests in my personal projects, some load in droplets and local, and without https, images load more faster. i'm not sure if servers are over ssl, 'cause the links examples are over http only.. but it's only my suggestion

PR #1939, opened 2020-02-26
Feature/pl pl license plate

scybulski:
A generator of Polish vehicle registration numbers.
It tries to follow [variety of possible registration numbers used in Poland](https://pl.wikipedia.org/wiki/Tablice_rejestracyjne_w_Polsce#Tablice_standardowe).
It also allows to generate numbers only for specific voivodeships or counties as well as for cars of services like Police, Tax Service or Counterintelligence Service.


scybulski:
Tests fail because of the issue in another provider. [The PR](https://github.com/fzaninotto/Faker/pull/1942) is ready.

Chomiciak (approved):
Good approach, it however doesn't cover the whole license plate system - there are plenty of exceptions, eg. custom plates or short plates for vehicles imported.

Regardless of that, it's a very useful addition.

Kraudia (approved):

PR #1942, opened 2020-02-26
ro_RO Person: fixed generating first and last day of year

scybulski:
I've noticed in Febuary tests in all PRs fail because of Romanian Person provider.
I found out it was because the first and last day of the year was generated incorrectly. 
This made the provider in one of the tests to generate non-existing date `1983-02-29` which was then compared against `1983-03-01`
PHP function `strtotime()` assumes that when no month name is given, the first day of current month in the specified year is generated.

The tests are likely to pass in the following months without this update.


mikk150:
@fzaninotto plz have this merged, so we all can have working PR-s

scybulski:
@fzaninotto, can we have this simple PR merged to avoid issues with tests in the future?

PR #1956, opened 2020-03-11
Project specific files

pimjansen:


.gitignore
```diff
@@ -1,3 +1,3 @@
-.build/
-vendor/
-composer.lock
+/.build
+/vendor
+/.idea
```
localheinz: I'd prefer to add this to a global configuration - otherwise we end up with a list here for every single IDE.

What do you think?

pimjansen: We could do but tbb, i dont think there are that many is it?

localheinz:
Thank you, @pimjansen!

PR #1957, opened 2020-03-11
Removed all the ORM/ODM integrations from the core

pimjansen:
As discussed i removed all the ORM/ODM integrations from the core


localheinz:
Thank you, @pimjansen!

PR #1958, opened 2020-03-11
Enhancement: Collect code coverage using pcov

localheinz:
This PR

* [x] uses `pcov` to collect code coverage instead of `Xdebug`

Follows #1890.

💁‍♂ It's a lot faster!


pimjansen:
great @localheinz

localheinz:
Thank you, @pimjansen!

PR #1962, opened 2020-03-15
Migrated Faker to match PSR12 instead of PSR2

pimjansen:
Ran an autofix over the project


localheinz:
Thank you, @pimjansen!

PR #1963, opened 2020-03-15
Fix: Update composer.lock

localheinz:
This PR

* [x] updates `composer.lock`

💁‍♂ Running

```
$ composer validate
```

on current `next` yields

```
The lock file is not up to date with the latest changes in composer.json, it is recommended that you run `composer update` or `composer update <package name>`.
./composer.json is valid
```


localheinz:
Thank you, @pimjansen!

PR #1897, opened 2020-01-13
Enhancement: Run a static code analysis and collect type coverage with vimeo/psalm

localheinz:
This PR

* [x] requires `vimeo/psalm`
* [x] runs a static code analysis and collects type coverage with `vimeo/psalm` on GitHub Actions
* [x] adds `Makefile` targets
* [x] requires `psalm/phpunit-plugin`
* [x] fixes a few fixable issues reported via static analysis
* [x] runs `make static-code-analysis-baseline` to generate the baseline for `vimeo/psalm`

PR #1964, opened 2020-03-15
Enhancement: Require error level 7

localheinz:
This PR

* [x] requires error level 7 for `vimeo/psalm`

Follows #1897.

💁‍♂ The idea is to raise the error level one by one and fix issues as we encounter them.


localheinz:
Thank you, @pimjansen!

PR #1966, opened 2020-03-15
Fix: Require test code to follow our coding standard as well

localheinz:
This PR

* [x] requires the test code to follow our coding standard as well
* [x] runs `make fix`
* [x] manually fixes issues that `phpcbf` cannot fix automatically

Follows #1896.


composer.json
```diff
@@ -37,7 +37,8 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "Faker\\Test\\": "test/Faker/"
+            "Faker\\Test\\": "test/Faker/",
+            "Faker\\Test\\Fixture\\": "test/Fixture/"
```
localheinz: I've introduced a new namespace mapping here - we might want to use PSR-4 mapping with a flattened directory structure for the tests at some point as well!

localheinz:
Thank you, @pimjansen!

PR #1969, opened 2020-03-15
Enhancement: Reference schema for phpcs configuration

localheinz:
This PR

* [x] references `phpcs.xsd` as installed with `composer` in `phpcs.xml`

Follows #1896.

💁‍♂ Helps with auto-completion!


localheinz:
Thank you, @pimjansen!

PR #1975, opened 2020-03-23
Update PhoneNumber.php

ngocvd:
Change code


pimjansen:
Do we have some reference docs on why this change is needed?

Issue #1873, opened 2019-12-12
French phone numbers invalid for libphonenumber


williamdes:
### Summary
Ref: https://github.com/giggsey/libphonenumber-for-php/issues/352

Faker generates french phone numbers like `+33 (0)4 06 21 24 41` that are not considered valid for https://libphonenumber.appspot.com/phonenumberparser?number=%2B33+%280%294+06+21+24+41&country=FR (https://github.com/giggsey/libphonenumber-for-php/issues/352#issuecomment-564572213)

### Versions

|                     | Version |
|:--------------------|:--------|
| PHP                 | 7.2.24   |
| `fzaninotto/faker`  | 1.9.1   |

### Self-enclosed code snippet for reproduction

```php
$faker->phoneNumber()
```

### Expected output

```txt
+33 4 06 21 24 41
```

### Actual output

```txt
+33 (0)4 06 21 24 41
```

If their an option to have libphonenumber/google valid phone numbers ?

williamdes:
@fzaninotto maybe Faker can not do nothing for this issue ?

williamdes:
@fzaninotto I added some links and pictures

fzaninotto:
Thanks. Indeed, the French PhoneNumber provider could be improved to generate only valid numbers. Would you like to work on a patch?

williamdes:
@fzaninotto If you have time you can patch the issue, I think I will have time in ~1-2 weeks

williamdes:
@fzaninotto if you want to fix it then go ahead, I am not having the time needed for this issue soon

jonatanbzk:
I checked this issue and fixed it changing this file :
**Faker/src/Faker/Provider/fr_FR/PhoneNumber.php**
![faker](https://user-images.githubusercontent.com/62499410/77735215-73065480-700a-11ea-9412-82a1b49edde5.png)

The result looks ok and the tests also passed.

PR #1978, opened 2020-03-27
Fix bug issue #1873 - French phone numbers invalid

jonatanbzk:
I checked this issue and fixed it changing this file :
Faker/src/Faker/Provider/fr_FR/PhoneNumber.php
Faker didn't generate numbers like +33 (0)4 06 21 24 41 anymore for french phone numbers.

https://github.com/fzaninotto/Faker/issues/1873

### DIFF
--- a/src/Faker/Provider/fr_FR/PhoneNumber.php
+++ b/src/Faker/Provider/fr_FR/PhoneNumber.php
@@ -7,16 +7,6 @@ class PhoneNumber extends \Faker\Provider\PhoneNumber
     // Phone numbers can't start by 00 in France
     // 01 is the most common prefix
     protected static $formats = array(
-        '+33 (0)1 ## ## ## ##',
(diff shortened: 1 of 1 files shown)


pimjansen:
Can we have some reference docs to verify the change? Why is it needed and so on?

jonatanbzk:
Hi,

Currently Faker generates French phone numbers like +33 (0)4 06 21 24 41 that are not considered valid.

[](https://en.wikipedia.org/wiki/Telephone_numbers_in_France)

'When calling France from abroad, the leading zero should be omitted: for example, to call a number in Southwest France, one would dial +33 5 xx xx xx xx.' 

For exemple, French phone number 01 23 45 67 89 from abroad should be +33 1 23 45 67 89 (or 0033 1 23 45 67 89), not +33 (0)1 23 45 67 89.

PR #1982, opened 2020-04-08
Update Internet.php

maitrepylos:
Add tld official Belgium


pimjansen:
I am not a fan of this since technically everything can be a TLD nowadays

pimjansen:
@localheinz what is your view on this?

PR #1986, opened 2020-04-24
Inn mask

tmaslov22:
Now you can use 12 digits of the Inn.


src/Faker/Provider/ru_RU/Company.php
```diff
@@ -96,7 +96,7 @@ public static function companyNameSuffix()
         return static::randomElement(static::$companyNameSuffixes);
     }
 
-    public static function inn($area_code = "")
+    public static function inn($area_code = "", $mask = '#######')
```
pwsdotru, line 99: Use mask as parameter is not good idea. I think better is use type of INN: for Person (12 digits) or Company (10 digits)

And then use type inside method for generate mask.

src/Faker/Calculator/Inn.php
```diff
@@ -15,8 +15,11 @@ class Inn
     public static function checksum($inn)
```
pwsdotru, line 15: There is possible add INN type (Person or Company) and then use it in cases inside method

src/Faker/Provider/ru_RU/Company.php
pwsdotru: Or move INN generator to common function in \Faker\Calculator\Inn and add inn to Person.php too

tmaslov22: > Or move INN generator to common function in \Faker\Calculator\Inn and add inn to Person.php too

Thank you for your comments. Using Inn base class and divide the INN for the company and user will be more object-oriented way.

PR #1991, opened 2020-04-30
Run unit tests without PHP memory limit

pwsdotru:
Changed command line for run phpunit

In default settings tests failed with Fatal Error.


pimjansen:
I dont see any need for this change where it runs fine? Also on our CI

pwsdotru:
Ok. Then for run tests users need change settings for PHP and up memory_limit for all PHP scripts.

Issue #1981, opened 2020-04-07
Invalid UK Postcode HR9 9AJ


SOMERSOFTAndrewCrook:
### Summary

The post code HR9 9AJ is nolonger valid. I suggest that it is replaced with HR9 5BW which is possibly near the same place.

See https://www.streetcheck.co.uk/postcode/hr99aj for more information about this post code and its history.

### Versions

All versions.

PR #1992, opened 2020-05-01
#1981: Fix UK Postcodes

pwsdotru:
Fix UK Postcodes as in issue https://github.com/fzaninotto/Faker/issues/1981


nusje2000 (approved):

pimjansen:
Can we have some refs where it indicates that it was wrong etc? Since we are removing stuff from the lib

pwsdotru:
> Can we have some refs where it indicates that it was wrong etc? Since we are removing stuff from the lib

Pull request was made by issue: https://github.com/fzaninotto/Faker/issues/1981

PR #1993, opened 2020-05-04
[Docs] randomElement accepts an array

reed-jones:


pimjansen (approved):

pimjansen:
Thanks for the update!

Issue #1862, opened 2019-12-02
Loaded entities in column formatter seem to be treated as new entities in 1.9.0


jaapromijn:
### Summary

When upgrading from 1.8.0 to 1.9.0, using a collection of existing entities in a column formatter results in an error.
It seems that the existing entities are now treated as new entities instead.

This seems to be caused by the addition in PR https://github.com/fzaninotto/Faker/pull/1781: https://github.com/fzaninotto/Faker/commit/c49cd5438655043b487f4d986961746ae58edb3a#diff-d38baf8c465e27d0cc5334cf5d8475aaR105

### Versions

|                     | Version |
|:--------------------|:--------|
| PHP                 | 7.3.10   |
| `fzaninotto/faker`  | 1.9.0  |

### Self-enclosed code snippet for reproduction

```php
        $users = $manager->getRepository(User::class)->findAll();

        $populator->addEntity(Status::class, 50, [
            'user' => function () use ($generator, $users) {
                $key = array_rand($users);

                return $users[$key];
            },
        ]);
```

### Expected output

```txt

```

### Actual output

```txt
  [Doctrine\DBAL\Driver\PDOException (23000)]
  SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'fabienne01@example.com' for key 'UNIQ_8D93D649F85E0677'
```

pimjansen:
@jaapromijn i dont think this commit breaks the unique constraint here. It executes an extra flush on Doctrine ORm when it hits the batchsize.

However i see you grab a random user from your existing ORM here in a callback without using a generator?

jaapromijn:
@pimjansen The ` $entityManager->clear($class);` breaks the code as it was. When removing this call in the code, the code runs again without error.

The doctrine documentation says the following about clear:
> When EntityManager#clear() is invoked, all entities that are currently managed by the EntityManager instance become detached.

Could you point me to an example of using the generator to return a random entity from the database?

pimjansen:
@jaapromijn correct however since all is flushed already, clearing and detaching it should not be a problem? Can you explain the use case, since looking at the code you are quering all users and are populating those again? Isnt it a bit normal that it gives duplicates in that case?

jaapromijn:
@pimjansen Thanks for your time. We are generating `Status` objects and assigning a (random) existing user (previously generated with the same populator) to those `Status` objects.

So for the `Status.user` field, a `User` object is returned from the collection.

[...]

PR #1995, opened 2020-05-09
Fix (bug #1862): Do not clear entity manager for doctrine orm populator

jlekowski:
Bug: https://github.com/fzaninotto/Faker/issues/1862
Clear specific entities is deprecated and will not be supported in Doctrine 3.0. (https://github.com/doctrine/orm/commit/9ccb8837e7f0bc605117024561c07e487a57eca0)
Bug introduced: c49cd5438655043b487f4d986961746ae58edb3a
Related issue: https://stackoverflow.com/questions/18215975/doctrine-a-new-entity-was-found-through-the-relationship

### DIFF
--- a/src/Faker/ORM/Doctrine/Populator.php
+++ b/src/Faker/ORM/Doctrine/Populator.php
@@ -98,11 +98,9 @@ class Populator
                 );
                 if (count($insertedEntities) % $this->batchSize === 0) {
                     $entityManager->flush();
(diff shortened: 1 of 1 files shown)


jlekowski:
I tested that on my project with different batch sizes, but I don't generate many records.

Alternatively the whole c49cd5438655043b487f4d986961746ae58edb3a (`batchSize` option) could be reverted.

PR #2000, opened 2020-05-14
Romanian Company provider and small .gitignore tweaks

lucian-vasile:
Updated .gitignore with some IDE-specific folders
Added Romanian Company provider


.gitignore
```diff
@@ -1,3 +1,11 @@
 .build/
 vendor
 composer.lock
+
```
pimjansen, line 4: Unrelated and not needed at all

pimjansen (changes requested):

.gitignore
lucian-vasile: Done

PR #2001, opened 2020-05-20
added feature to create a date time string between the two given dates in the given format

### LINKED ISSUE #1878: Add method to generate unique dates in a range
### Summary

There is currently no way to generate unique dates within a particular range.  `unique()->date()` only allows a maximum and always uses 1970-01-01 as the minimum date.  The dateTime* methods obviously return DateTimes, for which there could be up to 86,400 duplicate values for a single date rendering the unique modifier useless.

This limitation was already raised in #984, but it was closed even though the proposed solution cannot work with the unique modifier.

### Versions

|                     | Version |
|:--------------------|:--------|
| PHP                 | 7.4.0   |
| `fzaninotto/faker`  | 1.9.1   |

nusje2000:
Fixes #1878


nusje2000:
Closed due too  #1851

PR #2005, opened 2020-05-31
[fix] complies with psr-2

rymiyamoto:
I refactored some methods which were not psr-2 compliant.

Please review.


rymiyamoto:
:thinking: 
PR rebuild

PR #2023, opened 2020-07-08
add chinese bank account

elon-funs:
this my first time,i never do work for a open source project.i will do my best


Chomiciak (changes requested):
Could you please explain, why have you changed all the arrays in the whole lib? 😄 
_533 files changed_

PR #2026, opened 2020-07-09
Update Person.php

maris-liepins-ermac:
Use current year to search for century instead of random year.

### DIFF
--- a/src/Faker/Provider/kk_KZ/Person.php
+++ b/src/Faker/Provider/kk_KZ/Person.php
@@ -183,7 +183,7 @@ class Person extends \Faker\Provider\Person
      */
     private static function getCenturyByYear($year)
     {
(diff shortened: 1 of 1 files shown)


src/Faker/Provider/kk_KZ/Person.php
```diff
@@ -183,7 +183,7 @@ class Person extends \Faker\Provider\Person
      */
     private static function getCenturyByYear($year)
     {
-        if ($year >= 2000 && $year <= DateTime::year()) {
+        if ($year >= 2000 && $year <= date('Y')) {
```
loogi007, line 186: why dont delete `&& $year <= date('Y')` ?

jevgenijsblaus: To not run into a future I believe

PR #2031, opened 2020-07-16
Typo in function's name

carlos-mora:
There is a typo in the functions's name, it should be tollFreePhoneNumber instead of tollFreeNumber


pimjansen:
This will be a breaking change!

carlos-mora:
@pimjansen The breaking change was to accept the spanish module with a bad named function. The interface spects the function to be called **tollFreePhoneNumber** instead of **tollFreeNumber**, so in the current state the spanish language is broken.
Try it yourself: Use tollFreePhoneNumber in your current language, be sure it works fine, then change to es_ES language. 
Are you the reviewer to accept the push?

carlos-mora:
According to the docs, 
imagen
So the function should keep the name accross languages.

PR #2034, opened 2020-07-23
Added Uzbek language support

Erkinbek:
in latin alphabet


Erkinbek:
I'll fix errors and will be resend pull request

Issue #2027, opened 2020-07-13
[BUG] dateTimeThisYear method does not generate a properly date with current year


didix16:
### Summary

Hi, like title says, I'm using the method dateTimeThisYear() to generate a date from current year but, sometimes is generating a date from previous year!

### Versions

|                     | Version |
|:--------------------|:--------|
| PHP                 | 7.4.5   |
| `fzaninotto/faker`  | ^1.4   |

### Self-enclosed code snippet for reproduction

```php
for($i = 0; $i < 100; $i++) {
    $y = $faker->dateTimeThisYear()->format("Y");

    if ($date != "2020") {
        throw new \Exception("FAKER ERROR. Generated year: ". $y);
    }
}
```

### Expected output

```txt
It should not generate an Exception.
```

### Actual output

```txt
FAKER ERROR. Generated year: 2019
```

PR #2039, opened 2020-08-05
Fix dateTimeThisYear method

macaron:
Fix dateTimeThisYear method in issue #2027

I referred to this document.
https://www.php.net/manual/en/datetime.formats.relative.php#122278

Thank you.


Chomiciak (approved):
Great fix. It is quite interesting how did it happen that someone made such a mistake...

richardkeep (approved):
Good catch

pimjansen (approved):

pimjansen:
Thanks, nice find

PR #2048, opened 2020-09-03
Fix: Mark `$unique` property as nullable

andyg0808:
`$unique` is `null` until `unique()` is called; marking it as nullable should make that clear.

This came up for me because [Psalm](https://psalm.dev/) complains about subclasses of `Faker/Provider/Base` not initializing `$unique`.

### DIFF
--- a/src/Faker/Provider/Base.php
+++ b/src/Faker/Provider/Base.php
@@ -15,7 +15,7 @@ class Base
     protected $generator;
 
     /**
-     * @var \Faker\UniqueGenerator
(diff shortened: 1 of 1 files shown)


andyg0808:
It looks like CI failed due to some kind of transfer error on `composer install`. I don't seem to have permission to rerun CI; would someone else be able to do so?

pimjansen:
We have a psalm implementation discussion going so lets wait untill that is cleared up

Issue #97, opened 2012-12-18
Documentor & non-API public static methods


alexsegura:
I noticed `Documentor` throws a warning with french locale

<pre>
Warning: Missing argument 1 for Faker\Provider\fr_FR\Company::isCatchPhraseValid()
</pre>

The problem is that `Documentor` parses all the public methods,  and `isCatchPhraseValid` is public static, to make it testable. The warning can be easily removed by adding a default value to the parameter (should I send a pull request for that ??)

<pre>
public static function isCatchPhraseValid($catchPhrase = '')
</pre>

**BUT** `isCatchPhraseValid` is still returned as an API method by `Documentor`

I didn't really understand why there is a mix of instance & static public method inside Providers, what is the difference ? 

Maybe `isCatchPhraseValid` shouldn't be exposed as public or moved to another class ?

PR #2057, opened 2020-09-29
Changed latin "a" to cyrillic "а"

mingalevme:
Faker\Provider\ru_RU\Person uses suffix "a" to make a female last name from male, but letter is a latin "a" (ASCII #97) and not cyrillic "a" (ASCII #208), so the final last name is "some cyrillic symbols" + "latin a", commit fixes that, now female last name forms correctly: cyrillic male last name + cyrillic "a".

### DIFF
--- a/src/Faker/ORM/Doctrine/Populator.php
+++ b/src/Faker/ORM/Doctrine/Populator.php
@@ -98,11 +98,9 @@ class Populator
                 );
                 if (count($insertedEntities) % $this->batchSize === 0) {
                     $entityManager->flush();
-                    $entityManager->clear($class);
                 }
             }
             $entityManager->flush();
-            $entityManager->clear($class);
         }
 
         return $insertedEntities;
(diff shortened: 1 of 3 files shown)


sanmon1985 (approved):

src/Faker/ORM/Doctrine/Populator.php
```diff
@@ -98,11 +98,9 @@ public function execute($entityManager = null)
                 );
                 if (count($insertedEntities) % $this->batchSize === 0) {
                     $entityManager->flush();
-                    $entityManager->clear($class);
```
pimjansen, line 101: This is totally unrelated?

src/Faker/ORM/Doctrine/Populator.php
```diff
@@ -98,11 +98,9 @@ public function execute($entityManager = null)
...
                 }
             }
             $entityManager->flush();
-            $entityManager->clear($class);
```
pimjansen, line 105: Unrelated?

pimjansen (changes requested):
See my comments

src/Faker/ORM/Doctrine/Populator.php
mingalevme: This is from already merged commit (https://github.com/fzaninotto/Faker/commit/5337ce5261b393937bd7aaebb21020159756a152), just because I branched the repo before.

localheinz:
Closing in favour of #2056.

PR #2058, opened 2020-09-30
adding JCB card

mihaileu:
Adding JCB card type


mihaileu:
@pimjansen @fzaninotto still alive ?

Issue #2055, opened 2020-09-24
Doctrine\Common\Persistence\ is Deprecated in 5.0 Symfony


tebaly:
### Summary

Replace all Doctrine\**Common**\Persistence\ to Doctrine\Persistence\

### Versions

|                     | Version |
|:--------------------|:--------|
| PHP                 | x.y.z   |
| `fzaninotto/faker`  | x.y.z   |

### Self-enclosed code snippet for reproduction

```php
<!-- please replace this with a self-enclosed usage example -->
```

### Expected output

```txt
<!-- please replace this with the expected output of your self-enclosed example --> 
```

### Actual output

```txt
<!-- please replace this with the actual output of your self-enclosed example --> 
```

PR #2059, opened 2020-09-30
update use Doctrine\Common\Persistence to Doctrine\Persistence

etienneleba:
#2055


dunglas (approved):

Alymosul:
Can we get this merged?

pimjansen:
How does this work with older Doctrine versions?

dunglas:
No it doesn't work with oldest Doctrine versions. But those versions aren't supported anymore.

PR #2063, opened 2020-10-02
Enhancement: Add support for PHP 8.0

chris-doehring:
ThiHey there,

as you may know, php 8 will be released on the 26th of November 2020. Unfortunately, this package does not allow php 8, so I simply tried to change the composer version constraint to `>=5.3.3`, but then no unit tests would work under php 8, as only phpunit `^9.0` would allow php 8. Setting something like `^5.3.3 || ^7.0 || ^8.0` is no option either, as phpunit 8 introduced some backward incompatible changes which would require different tests for different phpunit and php versions.

So the logical conclusion for me was to set the same version constraint as in phpunit 9 (`>=7.3`). This also means, that a new major release of Faker (as of now `v2.0.0`) would be required, as it now drops support for older php versions. I know that's a hard step to take, and I know that there is indeed an ongoing planning and working going on for the "next Faker version", but as far as I can tell, these works are not going to be finished before php 8 has been released. And this package is included in many projects `require-dev` section. It is required for those projects to have a php 8 compatible Faker version in order to upgrade to php 8. This is the reason why I created this pull request for the master branch, instead of the "next" release branch.

I was able to get everything running for php 7.3, 7.4 and 8.0. What’s your thought on this?


GrahamCampbell:
Laravel needs PHP 7.2 support to stay around. You should be able to support PHPUnit 8.5 in parallel with 9.x.

GrahamCampbell:
Let me know if you need a hand. Also, dropping old PHP is not a BC break, and could be done as 1.10.0.

chris-doehring:
Thanks for your feedback, @GrahamCampbell! I changed it locally and it looks good, but I have one problem to solve: there are a lot of regex assertions in the tests and until now `assertRegExp` has been used, which also exists in phpunit `^9.0` but throws deprecations as warnings as it has been replaced by `assertMatchesRegularExpression`, but this method does not exist in phpunit `^8.0`. I will try to create a custom TestCase class which is overloading the original phpunit class where I will solve that compatibility issue. I will leave feedback here as soon as I am done.

chris-doehring:
I was able to change it as described and I was even able to support php 7.1 as well, but I can't get lower than this. Regarding what the new version will be: I guess that’s up to the maintainer to decide. I just hope that there will be a php 8 compatible Faker version soon.

GrahamCampbell:
In the event of the maintainer being long-term unavailable, I think it's worth considering making an official fork of this package to allow the package to continue into 2021 and beyond.

pimjansen:
What is the usecase of dropping old php versions? I do not think it is needed right?

chris-doehring:
It's not really a feature or my intend to drop the support for php `>=5.3.3 <7.1` but a necessity in order to use phpunit `>=8` and therefor php 8, as those `setUp` and `tearDown` methods now need to return `void`, which is a php 7.1 feature. Otherwise the most tests are not compatible. In order to keep the php versions `>=5.3.3 <7.1` supported, many tests need to be rewritten in order to use different approaches so that it's compatible to those older phpunit versions. But to be honest: I have no clue how to do that without creating some messy tests.

chris-doehring:
Maybe I have an idea how to do it, but again, thats usually what one shouldn't do when creating unit tests: I will try to rename all setUp/tearDown methods to something different and make my new TestCase magically call them in the original setUp/tearDown. It won't look nice, but that way we could support php `>=5.3.3`

GrahamCampbell:
I don't think dropping old PHP should be an issue. People can still use an older version of faker.

pimjansen:
> I don't think dropping old PHP should be an issue. People can still use an older version of faker.

Dont think @fzaninotto agrees on that statement though

chris-doehring:
@pimjansen That's a shame though. The code base regarding testing will be a mess. I found a way how to support all php versions, but I have to create some backward compatible custom methods in my base test class. I'm in progress of implementing that right now.

pimjansen:
Yeah lets wait for his reply in that case since i agree that it should be php 8 compatible once its released

ollieread:
Any news on this? Anything I can do to help speed things up?

chris-doehring:
We are waiting for feedback of @fzaninotto on this. @GrahamCampbell was so kind and wrote him a mail about the future of this project. So we either get a response or the other possibility still remains: Creating a new forked project with new maintainers.

ollieread:
> We are waiting for feedback of @fzaninotto on this. @GrahamCampbell was so kind and wrote him a mail about the future of this project. So we either get a response or the other possibility still remains: Creating a new forked project with new maintainers.

Ah fair enough. Should such an effort be required id be happy to help maintain/contribute.

GrahamCampbell:
FYI, this was the content of the email:

image

dunglas:
Regarding the issue with PHPUnit versions, you can use the Symfony PHPUnit Bridge with handle automatically compatibility with most PHPUnit versions (using black magic). Basically, replace `phpunit/phpunit` by `symfony/phpunit-bridge` in `composer.json` and update the CI to run `vendor/bin/simple-phpunit` instead of `phpunit` and this should do the trick.

We've done it recently to support PHP 8 in API Platform and the Negotiation library and it worked like a charm. The tool provide a polyfill for new PHPUnit features, so you can use the new assertions even with old PHPunit versions, no need for backward compatibility hacks in the tests.

dunglas (approved):

chris-doehring:
That was a great hint @dunglas! Thanks a lot! It's indeed black magic, but really nice as it saves a lot of time  😄

GrahamCampbell:
I see no reason to change this PR. PHP 5 is totally dead now, and we don't need to make any effort to support it. A minimum version of PHP 7.1, going forward, is totally fine,

.gitignore
```diff
@@ -1,3 +1,4 @@
+/.idea
```
GrahamCampbell, line 1: Don't add this.

composer.json
```diff
@@ -18,7 +18,7 @@
     },
     "require-dev": {
         "ext-intl": "*",
-        "phpunit/phpunit": "^7.5 || ^8.5 || ^9.4",
+        "symfony/phpunit-bridge": "^5.0",
```
GrahamCampbell, line 21: Better go with `^5.1`.

GrahamCampbell (changes requested):

composer.json
dunglas: You could even use `dev-master` for this one which is a bit special (the master version support even very old PHPUnit and PHP versions).

chris-doehring: Alright, will replace it with `*`. I don't think that we need to support those very old phpunit versions as this PR would change the php version support to `>=7.1`

.gitignore
chris-doehring: removed

composer.json
dunglas: I mean "even" the master version supports very old PHPUnit versions but that's ok.

composer.json
```diff
@@ -18,7 +18,7 @@
     },
     "require-dev": {
         "ext-intl": "*",
-        "symfony/phpunit-bridge": "^5.0",
+        "symfony/phpunit-bridge": "*",
```
GrahamCampbell, line 21: Do not make this a wildcard! `^5.1` is fine.

GrahamCampbell (changes requested):

composer.json
chris-doehring: To be honest: I'm fine with both. Whats your reason of changing it back to a more fixed version?

GrahamCampbell: Wildcards are almost always a bad idea, and lower versions don't work with PHP 8.

chris-doehring: Ah, I guess what you mean is, that this package (like all symfony packages) has a flexible way of allowing php versions (`>=5.5.9`) and older versions have that too, which could end up in installing an old version of this package with existing or known php 8 issues. I get your worries then. I will change it back to `^5.1`.

pimjansen:
Can we ensure that the existing php versions stay active? The change as is means it is a breaking change for older users on which we should release a major.

We will do this for future release but not now to make it php 8 compatible.

driesvints:
I'd personally drop anything below ^7.2.5. PHP 7.2 is almost EOL: https://www.php.net/supported-versions.php

GrahamCampbell:
Dropping PHP 5 is not a breaking change, and keeping at `^7.1 || ^8.0` is fine for now. If we want to drop 7.1 to use features from 7.2, then we shouldn't hesitate, but no point in dropping for no reason, when everything works. I am happy with this PR exactly in it's current state.

driesvints:
@GrahamCampbell sure

pimjansen:
> Dropping PHP 5 is not a breaking change, and keeping at `^7.1 || ^8.0` is fine for now. If we want to drop 7.1 to use features from 7.2, then we shouldn't hesitate, but no point in dropping for no reason, when everything works. I am happy with this PR exactly in it's current state.

That is how you look at it. If i am an user with an app on php 5 and i target the lib at 1.9 i will receive 1.10 and thus will break my installation.

So therefor it is a breaking change and we should keep supporting it.

GrahamCampbell:
> That is how you look at it. If i am an user with an app on php 5 and i target the lib at 1.9 i will receive 1.10 and thus will break my installation.

No you won't. Composer knows that your old PHP version cannot use a package marked as not compatible with it. Composer will install 1.9.x for you instead.

driesvints:
@pimjansen updating dependencies in stable releases isn't a breaking change: https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

pimjansen:
Ok missed that one though thanks. Let me review it tomorrow to see how we can ensure it is release before the actual PHP release itself as 1.10

fzaninotto:
Hi there! I'm not sure I understand where we are on this PR. 

I understood that phpUnit was the main blocker for supporting php8

I understood that @dunglas' trick allows us to use any phpUnit version with Faker.

So I understand that we can make this lib compatible with php8 without dropping support for php5. 

Or am I missing something?

chris-doehring:
In case you want to support php 5, 7 and 8 at the same time, you have to drop support for php 5.3. `phpunit-bridge` uses traits to fake the `void` return type for those setUp and tearDown methods, which has been introduced in phpunit 8. All that being said without having actually changed _all_ tests. So I would only see at the end if everything really works for `>=5.4`.

What's your point of view regarding dropping the support for php 5, taking the previous comments of different persons into account?

pimjansen (approved):
Looks ok by me for the rest

pimjansen:
@localheinz can you also have a look to push this for 1.10. If not we will not have proper PHP8 support and an huge inflow on bugs.

localheinz:
@pimjansen 

I wouldn't rush support for PHP 8.0, especially taking into account things like named arguments.

.gitignore
```diff
@@ -1,3 +1,4 @@
 .build/
 vendor
 composer.lock
+/.phpunit.result.cache
```
localheinz, line 4: Please point `phpunit/phpunit`s result cache into a location in the already ignored `.build` directory.

If necessary, we can create the directory before running tests.

phpunit.xml.dist
```diff
@@ -1,22 +1,22 @@
...
-    <testsuites>
-        <testsuite name="Faker Test Suite">
-            <directory>./test/Faker/</directory>
-        </testsuite>
-    </testsuites>
+  <testsuites>
+    <testsuite name="Faker Test Suite">
+      <directory>./test/Faker/</directory>
+    </testsuite>
+  </testsuites>
```
localheinz, line 15: All of the changes in this file are unrelated, can we revert them, please?

.github/workflows/continuous-integration.yml
```diff
@@ -135,8 +131,8 @@ jobs:
       - name: "Create build directory"
         run: "mkdir -p .build/logs"
 
-      - name: "Collect code coverage with Xdebug and phpunit/phpunit"
-        run: "vendor/bin/phpunit --coverage-clover=.build/logs/clover.xml"
+      - name: "Collect code coverage with Xdebug and phpunit"
```
localheinz, line 134: Can we revert this unrelated change?

We are still running `phpunit/phpunit`.

.github/workflows/continuous-integration.yml
```diff
@@ -135,8 +131,8 @@ jobs:
...
+        run: "vendor/bin/simple-phpunit --coverage-clover=.build/logs/clover.xml"
```
localheinz, line 135: I am not sure that this is necessary.

pimjansen:
> I wouldn't rush support for PHP 8.0, especially taking into account things like named arguments.

I don't really agree though. The named arguments do not change how the application works so its an easy win. For the next major we can use all the benefits of PHP 7.3 though

localheinz:
@pimjansen 

> I don't really agree though. The named arguments do not change how the application works so its an easy win. For the next major we can use all the benefits of PHP 7.3 though

Named arguments prevent us from renaming arguments without a new major release.

localheinz (changes requested):
Can you point this against the `next` branch, please, and revert all unrelated changes?

fzaninotto:
So it seems there are 2 PRs in this one:

1. makes Faker compatible with php8 by changing how phpUnit works. This requires increasing the minimal PHP version to PHP 5.4. I think we can live with that in `master`
2. Increases the minimal PHP version to 7.X. This will enable new features, but possibly in a non backwards compatible way. Let's do that in `next`

.github/workflows/continuous-integration.yml
chris-doehring: In my opinion this is clear based on the context given, but I will revert it.

driesvints:
I think we can live with just PHP 8 support in the current stable release for Laravel which is the most important right now.  Preferable in Faker 1.10 if that seems reasonable to you all. PHP 8 is fast approaching now and more people are beginning to test on it.

chris-doehring:
I personally disagree with the choice of having to support three php versions at once, but that is up to the maintainer to decide. I will do my best in making it compatible to php `^5.4 || ^7.0 || ^8.0`, so I will change this PR according to the first point @fzaninotto suggested.

I hope that we will get a faker with php 8 support _before_ the php GA release. Faker is a testing tool and there many projects depending on it which already want to upgrade in preparation for the release.

.github/workflows/continuous-integration.yml
chris-doehring: I tried it with regular phpunit and got this:
```
line 1: vendor/bin/phpunit: No such file or directory
```
It is necessary

localheinz:
@chris-doehring 

Can you point this agains the `next` branch, please?

chris-doehring:
@localheinz and what about the php 8 support in master? This branch is currently confliced 40 files in next. In my opnion it makes more sense to continue to work on this branch, make it work with php 5 and then create a new PR with new branch based on next.

phpunit.xml.dist
chris-doehring: reverted

.github/workflows/continuous-integration.yml
chris-doehring: reverted

localheinz:
@chris-doehring 

If you do not want to do it, I will.

chris-doehring:
@localheinz I just don't understand what you want to achieve by doing that. It would be better if you could explain what your plan is. At the end I'm only an author of a PR trying to help this project. Feel free to do it.

localheinz:
@chris-doehring @GrahamCampbell @driesvints @dunglas @fzaninotto @ollieread @pimjansen 

Since I changed the target branch from `master` to `next` and applied a few changes, would you like to take another look?

As far as I am concerned, I wouldn't mind cutting a `1.10.0` release from the `next` branch.

What do you think? Will this work for the Laravel community and everyone else interested in a timely release for PHP 8.0?

composer.json
```diff
@@ -21,12 +21,12 @@
         }
     ],
     "require": {
-        "php": "^7.2"
+        "php": "^7.2 || ^8.0",
+        "symfony/phpunit-bridge": "^5.1.7"
```
chris-doehring, line 25: That looks like a mistake or why did you move that from `require-dev` to `require`?

pimjansen:
@localheinz what will we do with the fixes in master that are done for 1.9.2? Can we try to merge those in vnext as well?

localheinz:
@pimjansen 

> @localheinz what will we do with the fixes in master that are done for 1.9.2? Can we try to merge those in vnext as well?

Definitely out of scope for this pull request!

We can sort this out later. 👍

composer.json
localheinz: 👍

pimjansen:
> @pimjansen 
> 
> > @localheinz what will we do with the fixes in master that are done for 1.9.2? Can we try to merge those in vnext as well?
> 
> Definitely out of scope for this pull request!
> 
> We can sort this out later. 👍 
> 

Totally right, was just figuring out the wat to go with the vnext branch thats why

fzaninotto:
@localheinz if you disagree with the OP, discuss in the PR, but don't push on their PR without their consent, this is rude. 

Besides, you an I posted contradictory messages in this discussion regarding the base branch. Before acting, we must agree on the path to follow.

localheinz:
@fzaninotto 

My intention was to help move forward with allowing the installation of `fzaninotto/faker` on PHP 8.0.

### Option A

As far as I can see, if we want to allow the installation of `fzaninotto/faker` on PHP 8.0 **while running tests on PHP 8.0 as well**, we need to drop support for 

- PHP 5.3
- PHP 5.4
- PHP 5.5
- PHP 5.6
- PHP 7.0
- PHP 7.1

That means that we will end up with the same changes in `master` that we have already made in `next`.

### Option B

Alternatively, we could run

```
$ composer require php:"^5.3 || ^7.0 || ^8.0"
```

on current `master`, and never run a test on PHP 8.0. If that's ok for everyone involved, that certainly is a viable option.

### Opinion

I don't see any value in supporting otherwise unsupported PHP versions. 

If anyone wants to use `fzaninotto/faker` on 

- PHP 5.3
- PHP 5.4
- PHP 5.5
- PHP 5.6
- PHP 7.0
- PHP 7.1

they can do so with any of the versions that are currently available.

Here's what I would have done, if I had full control over the repository:

1. create a `1.9` branch off `1.9.1`and keep it around for bug fixes only (this is what we have done, it is our current `master`)
2. create a `master` branch at the same time, make it the default branch, and start modernizing it aggressively (this is what we have done, it is our current `next`), while keeping backwards-incompatible changes as little as possible

I don't think it makes sense to provide support for PHP 8.0 in `master` (i.e., `1.9`) with **Option A**.

localheinz:
@fzaninotto 

What I mean is this: we can not move forward without making any changes. So either we make the changes and move forward, or we keep everything as it is. Then either someone else will fork the repository and make the changes there, or after enough pressure has built up to hand over the repository, then someone else will make the changes.

Why not make the changes here?

driesvints:
> What do you think? Will this work for the Laravel community and everyone else interested in a timely release for PHP 8.0?

For Laravel we'll need Faker to run on PHP 7.2, 7.3, 7.3 & 8.0 (preferable in Faker v1.10 f that's possible). So I think the current state of the PR reflects that? /cc @GrahamCampbell

GrahamCampbell:
Yeh, exactly. I don't think trying to support PHP 5 a valid position to take in 2020, given that older versions of this package can just be used on PHP 5. It should be easily possible to also support  PHP 7.1, if we support PHP 7.2, but it's not a big deal. I'd be happy with any of the following minimum versions: 7.1.0, 7.2.0, 7.2.5.

GrahamCampbell:
> but don't push on their PR without their consent, this is rude.

I disagree with this. The PR author has the option to specify if they want people to be able to push to their PR branch. They opted-in.

driesvints:
Just want to say thank you to all involved in trying to get this to land. Hope we can get this merged soon! :)

chris-doehring:
I considered it rude to the point that I didn't knew what the plan of @localheinz was. So it felt like I got pushed into a direction I didn't really understood. But never mind now. Let's focus on the topic: I'm happy with the current state of it and I'm also happy about all three options that @localheinz summarized.
And also from my side: Thank you to all people involved! I didn’t expect it to be such a active discussion. :)

localheinz:
Since I unfortunately haven't kept @chris-doehring original commit hashes around to revert all of the changes made in this pull request, how about this:

- We merge this pull request into `next`, and continue modernizing `next`.
- We allow installation of `fzaninotto/faker` on PHP 8.0 by opening a pull request targeting `master` that widens the platform requirements and tag a `1.9.2` release based on `master`.

    ```
    $ composer require php:"^5.3 || ^7.0 || ^8.0"
    ```

This way we have both:

- We can keep the work done by @chris-doehring for `next`
- We can release support for PHP 8.0 for `1.9` without making anymore changes to `master`.

If anyone runs into problems running `1.9.2` on PHP 8.0, we will eventually find out and can still drop support for older PHP versions on `master` and apply the original changes made by @chris-doehring to `master`, and actually run tests for PHP 8.0 on `master`.

What do you think?

chris-doehring:
Sounds good :) I created a new branch in my fork with only my changes, which will stay there in case there is need for it like you described: https://github.com/chris-doehring/Faker/tree/feature/v1-php8-test-compatibility

localheinz:
Thank you, @chris-doehring!

For reference, see #2090.

localheinz (approved):
👍

GrahamCampbell:
Sounds like a bad idea to allow 8.0 without CI testing on 8.0, if that is the suggestion for 1.9.x.

What is stopping dropping old PHP in 1.10.0 and backporting bug fixes to 1.9.x for those who cannot upgrade?

fzaninotto:
Hi all,

And thanks a lot for your contributions in this PR. 

I understand that there are different strategies to move Faker forward, and that I might not be the best person to decide which strategy is the best one. 

Yet I don't want to pass over Faker to another organization because I don't want to lose the benefit of a repo with 20+ stars. It's not just about vanity, it helps me get a better visibility in my job.

So I came to the conclusion that the best solution is to sunset fzaninotto/Faker. As far as I'm concerned, the project is complete. If you want to modify it, you can fork it and republish it under another organization - that's the beauty of open-source.

This decision is effective immediately. I'll publish a blog post to explain it in more details. 

As far as this PR is concerned, I'll close it because 1.9 will never be published.

driesvints:
@fzaninotto understandable. I'll talk this over with the rest of the team to see how we're going to proceed.

Thanks for all your work over the years and good luck with future plans :)

pimjansen:
@driesvints i also sent an email to Graham about future things. Would be good to have a chat about it as well?

freekmurze:
@fzaninotto Thanks for all your work on this repo over the years! Greatly enjoyed working with this package 👍

drupol:
Thanks for your work!

bramceulemans:
Very useful library, thanks for the effort!

mehdirochdi:
thank you so much for this package, it's very useful

GrahamCampbell:
I will be creating fork and inviting collaborators to the organisation.

GrahamCampbell:
I will also setup a slack so we can chat on there.

PovilasKorop:
@fzaninotto I've used faker thousands of times over multiple projects, so can't thank you enough for everything.

reinink:
@fzaninotto Thank you so much for all your work over the years on this project! I've used it a ton, and it's been extremely useful. 🤟

VladimirBerezkin:
@fzaninotto Thank you for your work!

PhillipMwaniki:
Thanks @fzaninotto for everything. Your repo is one of the best things to ever happen in the php eco system. Shukran

fallback:
Thanks, @fzaninotto

atouiahmed:
@fzaninotto 
Thank you for your efforts

svenluijten:
@fzaninotto Thank you for the work you put into this. I don't think I'm overreacting when I say this has had a massive impact on the PHP ecosystem as a whole. Good luck with your next endeavors, whether that's in open source, or outside of it! ❤️

seunoyebode:
@fzaninotto Thank you for all the efforts over the years

carnevalle:
Thank you for your efforts and good luck with future endeavors!

localheinz:
I have just published [`ergebnis/faker:1.9.2`](https://github.com/ergebnis/faker/releases/tag/1.9.2) for people who need support for PHP 8.0 immediately.

elishaukpong:
@fzaninotto  Thank you a lot for the efforts you put into this project, I sincerely enjoyed using it over my development journey.

chris-doehring:
Thank you for your work @fzaninotto. As I already said, I'm using this package plenty of times and it helped me making my tests more reliable and therefor my applications more secure :)

mazraara:
Thanks for your work!

fourstacks:
To chime in with the others in this thread - thanks so much for your work on Faker over the years.  It's been a fantastic tool and I hope it continues in some shape or form in the future.

joecampo:
Thank you for all your hard work throughout the years!

ruslanashaari:
@fzaninotto thanks for your efforts over the years!

negoziator:
I want to thank you very much for all the work you've done. 
I've been using faker heavily through the years. 

Thank you! 😎

GrahamCampbell:
https://github.com/FakerPHP

ssmusoke:
@fzaninotto Faker has been invaluable to the community thank you for all you have done. 

Your approach to sunsetting the project has also opened a path for handling of open source projects too, thank you for pioneering this too

hotmeteor:
Thanks for all your work on this great package, @fzaninotto

Cannonb4ll:
Understandable, thank you for your work on this @fzaninotto  💪

techguydev:
Thanks a lot for your effort and great work!, thanks for sharing this project with all of us. Best wishes in your endeavors.

sjelfull:
Thank you for all the hard work, and good luck!

taytus:
Francois, thank you so much for your work.

lostdesign:
Even if most of the time I've been actually typing or filling fields in the DB by hand, Faker saved me a lot of time when I used it. Been such a lovely package, thanks for all your effort and love that went into this.

amitavroy:
Thank you for your effort... @fzaninotto 
I would like to tell you that you package has helped me a lot.
As a member of the community, I guess I respect your decision and wish you luck.
I hope, someone picks it up and continue this wonderful package.

thijsvdanker:
Thank you @fzaninotto for all your hard work on this great package 👏👏👏

lukio3:
Thank you so much @fzaninotto for your work on getting faker to where it is today.

luigircruz:
Thank you for the years of hard work that you've put on this package @fzaninotto! You've created a tool that is so beneficial to the PHP community.

SeoFood:
WoW thanks man for that amazing package 💕

fgilio:
Thanks for all your work, it's been immensely helpful 🙌

cyrildewit:
@fzaninotto Thank you for your work!

Hasnayeen:
@fzaninotto thank you for all your work!

Ali-Shaikh:
@fzaninotto Thank you for your hard work. Wish you the best in your future endeavours 👍

rogervila:
Thank you @fzaninotto :)

next65:
Thank you @fzaninotto

fabioi:
Thank you a lot, this package helped a lot of people! 🙏

vikas5914:
Thank you @fzaninotto

stelgenhof:
Thank you so much for this great project @fzaninotto !

damiani:
Thanks for all your hard work on this @fzaninotto!

mypromo-ml:
Thank you for all this @fzaninotto ! All the best for you :-)

rnleal:
Thank you @fzaninotto

austintoddj:
Thanks for all the hard work @fzaninotto!

ceejayoz:
Faker has been very useful to me and I just wanted to chime in with thanks as well.

rslhdyt:
Thanks for all the hard work @fzaninotto ❤

PR #2068, opened 2020-10-10
Update optional() in readme to use v2 weight

patrickdaze:
Update documentation for `optional()` to use the new weight system (percent 0 to 100 instead of float 0 to 1).

[`optional()` source code](https://github.com/fzaninotto/Faker/blob/master/src/Faker/Provider/Base.php#L536-L559)

---

🙏 Possible to add `hacktoberfest-accepted` label to this PR? ([Details](https://hacktoberfest.digitalocean.com/details))

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -365,14 +365,15 @@ for ($i = 0; $i < 10; $i++) {
 print_r($values); // [1, 4, null, 9, 5, null, null, 4, 6, null]
 
 // optional() accepts a weight argument to specify the probability of receiving the default value.
-// 0 will always return the default value; 1 will always return the provider. Default weight is 0.5 (50% chance).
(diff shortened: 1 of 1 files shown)


pimjansen:
I dont really see an actual usecase for this where the float is valid as well?

patrickdaze:
I assume it's still there for backwards compatibility. 

This PR doesn't change functionality. It only updates the README based on the new system. Maybe I can open a separate PR for [Faker v2](https://github.com/fzaninotto/Faker/projects/1) because it would be a breaking change? It would basically take care of [the TODO in the `optional()` function](https://github.com/fzaninotto/Faker/blob/b4be9ff96c8581f3452f11a0825022b971c6aac5/src/Faker/Provider/Base.php#L548).

PR #2069, opened 2020-10-13
Add: Swedish (sv_SE) municipality

sebastiansson:


Nyholm (approved):
I can confirm that the list is complete and It looks correct.

PR #2085, opened 2020-10-23
Enhancement: Enable SingleLineArrayWhitespace rule

localheinz:
This PR

* [x] enables the `SlevomatCodingStandard.Arrays.SingleLineArrayWhitespace` rule
* [x] runs `make fix`

Follows #1896.


fzaninotto:
This has zero value. We don't need bike shedding PRs at this time.

localheinz:
@fzaninotto 

Thank you for your feedback, but I disagree. This has nothing to do with bike shedding, this has more to do with cleaning up and improving a code base that, with all due respect, hasn't seen any care in years.

PR #2087, opened 2020-10-23
Fix: Remove psalm/plugin-phpunit

localheinz:
This PR

* [x] removes `psalm/plugin-phpunit`

❗ Blocks #2063.

💁‍♂️ On one hand, `psalm/plugin-phpunit` depends on `phpunit/phpunit`, and on the other hand, `symfony/phpunit-bridge` conflicts with `phpunit/phpunit`. If we want to use `symfony/phpunit-bridge`, we need to remove `psalm/plugin-phpunit`.


pimjansen (approved):

localheinz:
Thank you, @pimjansen!

PR #2088, opened 2020-10-23
Enhancement: Use vimeo/psalm as PHAR

localheinz:
This PR

* [x] uses `vimeo/psalm` as PHAR

❗ Blocks #2063.

💁‍♂️ This allows us to escape dependency hell: `vimeo/psalm:^3.18.2` has at least one dependency, i.e., `webmozart/glob`,  that does not support PHP 8.0 yet.

For reference, see:

- https://github.com/vimeo/psalm/blob/3.18.2/composer.json#L37
- https://github.com/webmozart/glob/blob/4.1.0/composer.json#L12

👉 Note that I am checking in [`.phive/psalm`](https://github.com/fzaninotto/Faker/pull/2088/files#diff-38b8c466eaddc15f8224c6637c02e110c365735374b48a6c6547236ca0b21187), so contributors do not have to install `phive`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,5 @@
 /.github/           export-ignore
+/.phive/            export-ignore
 /test/              export-ignore
 /.editorconfig      export-ignore
 /.gitattributes     export-ignore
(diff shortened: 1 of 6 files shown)


pimjansen (approved):

localheinz:
Thank you, @pimjansen!

